Issue #51, opened 2020-03-01
[new test] Profitablity


josojo:
During order placement, it was a concern whether the orders are really placed profitable for each bracket-safe. 

We should ensure this with our test suite

josojo:
https://github.com/gnosis/dex-liquidity-provision/pull/56

PR #55, opened 2020-03-03
Add price check to order creation

fedgiac:
This PR adds a check that warns the user when the price of dex.ag isn't close to the given price.

Partial remedy to issue #51.

### DIFF
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -60,7 +60,7 @@ const getDexagPrice = async function (tokenBought, tokenSold) {
     const requestResult = await axios.get(url)
     price = requestResult.data.price
   } catch (error) {
-    console.log("Warning: unable to compare the given price with that of dex.ag. The server returns:")
+    console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
     console.log(">", error.response.data.error)
   }
   return price
@@ -72,6 +72,13 @@ const isPriceReasonable = async function (exchange, targetTokenId, stableTokenId
   const targetToken = tokenInfo[targetTokenId]
   const stableToken = tokenInfo[stableTokenId]
   const dexagPrice = await getDexagPrice(targetToken.symbol, stableToken.symbol)
+  if (dexagPrice === undefined) {
+    console.log("Warning: could not perform price check against dex.ag.")
+    const answer = await promptUser("Continue anyway? [yN] ")
+    if (answer != "y" && answer.toLowerCase() != "yes") {
+      return false
+    }
+  }
   if (Math.abs(dexagPrice - price) >= thresholdPercent / 100) {
     console.log("Warning: the chosen price differs by more than", thresholdPercent, "percent from the price found on dex.ag.")
     console.log("         chosen price:", price, targetToken.symbol, "bought for 1", stableToken.symbol)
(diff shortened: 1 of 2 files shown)


scripts/bracket_orders.js
```diff
@@ -45,29 +48,82 @@ const argv = require("yargs")
...
+  const stableToken = tokenInfo[stableTokenId]
+  const dexagPrice = await getDexagPrice(targetToken.symbol, stableToken.symbol)
+  if (dexagPrice === undefined) {
+    console.log("Warning: could not perform price check against dex.ag.")
+    const answer = await promptUser("Continue anyway? [yN] ")
+    if (answer != "y" && answer.toLowerCase() != "yes") {
+      return false
+    }
+  }
+  if (Math.abs(dexagPrice - price) >= thresholdPercent / 100) {
```
josojo: Should this logic branch go into the an else branch, as dexagPrice is undefined and the price check will return an unreasonable result anyways?

scripts/bracket_orders.js
```diff
@@ -45,29 +48,82 @@ const argv = require("yargs")
...
+    const requestResult = await axios.get(url)
+    price = requestResult.data.price
+  } catch (error) {
+    console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
+    console.log(">", error.response.data.error)
+  }
+  return price
+}
+
+const thresholdPercent = 2
```
josojo: Maybe a little bit more descriptive name would be acceptedPriceDeviationInPercentage

scripts/bracket_orders.js
```diff
@@ -45,29 +48,82 @@ const argv = require("yargs")
...
+    if (answer != "y" && answer.toLowerCase() != "yes") {
+      return false
+    }
+  }
+  return true
+}
+
 module.exports = async callback => {
   try {
+    await BatchExchange.setProvider(web3.currentProvider)
```
josojo: We are using these lines of code several times in the project. We should make a function for that...

scripts/bracket_orders.js
```diff
@@ -45,29 +48,82 @@ const argv = require("yargs")
...
+        true,
+        argv.priceRange,
+        argv.validFrom,
+        argv.expiry
+      )
 
-    const answer = await promptUser("Are you sure you want to send this transaction to the EVM? [yN] ")
-    if (answer == "y" || answer.toLowerCase() == "yes") {
-      await signAndSend(masterSafe, transactionData, web3, argv.network)
+      const answer = await promptUser("Are you sure you want to send this transaction to the EVM? [yN] ")
```
josojo: If the price is not okay, do we will send out the transaction? Probably this should go into an else branch

scripts/bracket_orders.js
fedgiac: Absolutely right, I fixed that.

scripts/bracket_orders.js
fedgiac: `priceIsOk` is the final decision of the user, the price might as well be very unreasonable. I let the user send the transaction even if the price check fails because, for example, dex.ag might not have the traded token in their database but the token exists on the exchange.

josojo: Ups, I was confused with the brackets... yeah makes sense

Issue #49, opened 2020-02-28
Reword helper functions to use Transaction instead of directly manipulating data


fedgiac:
Ideally there would be no instance of "MultiSend" outside of the function that bundles transactions. There would also be no need to ever manipulate data to execute transactions with the safes.

This issue subsumes issue #44.

PR #58, opened 2020-03-04
Better transaction bundling

fedgiac:
This PR abstracts away as much as possible regarding MultiSend and execTransaction.
The result should be a better layer of abstraction for bundling and executing transactions.
Some workarounds have been removed, file structure has been reorganized. 

Linked to issue #49.


scripts/utils/internals.js
```diff
@@ -136,6 +147,90 @@ function signTransaction(lw, signers, transactionHash) {
...
+}
+
+/**
+ * Creates a transaction that makes a master Safe execute a transaction on behalf of a (single-owner) owned trader using execTransaction
+ * @param {EthereumAddress} masterAddress Address of a controlled Safe
+ * @param {EthereumAddress} traderAddress Address of a Safe, owned only by master, target of execTransaction
+ * @param {Transaction} transaction The transaction to be executed by execTransaction
+ * @return {Transaction} Transaction calling execTransaction; should be executed by master
+ */
+const getExecTransactionTransaction = async function(masterAddress, traderAddress, transaction, web3, artifacts) {
```
josojo: Here, most likely the replace function did not work as intended

scripts/utils/internals.js
```diff
@@ -136,6 +147,90 @@ function signTransaction(lw, signers, transactionHash) {
...
+  const GnosisSafe = artifacts.require("GnosisSafe")
+  const gnosisSafeMasterCopy = await GnosisSafe.deployed()
+
+  const execData = await execTransactionData(
+    gnosisSafeMasterCopy,
+    masterAddress,
+    transaction
+  )
+  const execTransactionTransaction = {
```
josojo: same here

scripts/utils/internals.js
josojo: Oh it was intended.

I would call the upper function `getExecutionTransactionData' and this one here 'getExecuctionTransaction'

fedgiac: I sticked to the rule that if a function creates a function that runs `operation`, then the function is called `getOperationTransaction`. The problem is that here the operation was called `execTransaction`!
Bit I very much dislike the name (and this naming convention) so I'll definitely change that.

PR #60, opened 2020-03-04
Prices smaller than one

### LINKED ISSUE #46: Orderprices incorrect
The current logic for calculating the order prices works only for prices>1. In the moment the prices gets smaller 1, then 
  ```  const upperBuyAmount = max128.toString()```
can no long hold, as we would need: 
```const upperSellAmount = max128.toString()```

### PR DESCRIPTION
**Pull request by josojo**
closes #46, #23

and tests the price for correctness as mentioned in the todos.

### CODE REVIEW

File: scripts/trading_strategy_helpers.js [hunk 1]
```diff
@@ -100,6 +100,10 @@ const maxUINT = new BN(2).pow(new BN(256)).sub(new BN(1))
  */
 
 const formatAmount = function(amount, token) {
+  if (amount < 1) {
+    amount = amount * 10 ** token.decimals
+    return new BN(amount.toString())
+  }
   return new BN(10).pow(new BN(token.decimals)).muln(amount)
 }
```
fedgiac: In another PR I'm writing a (hopefully) robust function to format amounts, this function could be easily replaced.

scripts/trading_strategy_helpers.js
```diff
@@ -272,29 +276,8 @@ const buildOrderTransactionData = async function(
...
-    // Sell max256 DAI for x ETH
-    // x = max256 / 101
-    const buyPrice = formatAmount(lowerLimit, targetToken)
-    const lowerSellAmount = max128.toString()
-    const lowerBuyAmount = max128
-      .div(buyPrice)
-      .mul(toETH(1))
-      .toString()
+    const [upperSellAmount, upperBuyAmount] = calculateBuyAndSellAmountsFromPrice(upperLimit, targetToken)
+    const [lowerBuyAmount, lowerSellAmount] = calculateBuyAndSellAmountsFromPrice(lowerLimit, targetToken)
```
fedgiac: I see that the output of calculateBuyAndSellAmountsFromPrice is [sellAmount, buyAmount], but in the second line sell and buy are inverted. Is this expected?

test/dfusionMultiSafes.js
```diff
@@ -25,6 +25,27 @@ const {
...
+    // Check buy order prices
+    assert.isBelow(
+      Math.abs(buyOrder.priceDenominator.div(buyOrder.priceNumerator).toNumber() - (minimalPrice + stepSize * index)),
+      1
+    )
+    // Check sell order prices
+    assert.isBelow(
+      Math.abs(sellOrder.priceNumerator.div(sellOrder.priceDenominator).toNumber() - (minimalPrice + stepSize * (index + 1))),
+      1
+    )
```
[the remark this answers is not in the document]
josojo: Sometimes, I am doing braindead things. I did the checks correctly, but then I refactored them just away :D 
https://github.com/gnosis/dex-liquidity-provision/pull/60/commits/7c7ee8a7de28798eaaba43b1cd782d5d9cd5933c

scripts/trading_strategy_helpers.js
josojo: I reused your the function now.

scripts/trading_strategy_helpers.js
josojo: I added a comment for explaintion

test/dfusionMultiSafes.js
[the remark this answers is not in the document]
josojo: I only see two ways to check the correctness of the prices: EIther, we calculate the exact buy and sell amounts again in BN, but then we are basically "reusing" the same logic and the same thinking as in the smart contract or we do this weird float check with the additional multiplicator. While the last one will always be inaccurate, it checks pretty well that the rough numbers are correct with a very different logic than in the smart contract. Hence I think that it more valuable than doing the exact logic with a risk to redo the same thinking error again.

test/dfusionMultiSafes.js
```diff
@@ -23,6 +22,43 @@ const {
...
+  let multiplicator = new BN("10")
+  if (targetPrice < 10) {
+    multiplicator = new BN("10000000")
+  }
+  // Correctness assertions
+  for (const [index, slaveAddress] of slaveSafes.entries()) {
+    const auctionElements = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders(slaveAddress))
+    assert.equal(auctionElements.length, 2)
+    const [buyOrder, sellOrder] = auctionElements
+    // Check buy order prices
```
bh2smith, line 38: Sorry to bring up questions on old changes, but these comments are too vague to know what is being"checked here". Perhaps the all the comments before assertions could be expanded upon and included as failure strings as the optional third parameter of the assertions themselves.

As a new developer coming into this project, this comment contains very little context apart from the fact that the next line is an assertion involving buy order price.

josojo: >an assertion involving buy order price.
It is not just involving buy order price, it is checking the actual buy order price. It checks the order price against the recalculated intended order price. 

How would you have described it in the comments?

bh2smith: That is the problem here. I don't know what this is for, so I can't describe it and the message doesn't help....

bh2smith: What about the buy order price is being checked? is it non-zero?

Issue #24, opened 2020-02-14
Bracket Trading with number of decimals other than 18


bh2smith:
Relates to the TODO list in #21

fedgiac:
Done with PR #89 and #105. :tada:

PR #61, opened 2020-03-04
Add tools for printing representations of ERC20 token amounts

fedgiac:
There are many ad hoc scripts and workaround to represent ERC20 token amounts. This PR creates reliable tools to do exactly that.

First step towards solving issue #24.

### DIFF
--- a/scripts/utils/printing_tools.js
+++ b/scripts/utils/printing_tools.js
@@ -17,7 +17,8 @@ const fromUserToMachineReadable = function (amount, digits) {
   const re = /^(\d+)(\.(\d+))?$/ // a sequence of at least one digit (0-9), followed by optionally a dot and another sequence of at least one digit
   const match = re.exec(amount)
   if (match == null) throw Error("Failed to parse decimal representation of " + amount)
-  const decimalString = (match[3] || "0").padEnd(digits, "0")
+  const decimalString = (match[3] || "").padEnd(digits, "0")
+  if (decimalString.length != digits) throw Error("Too many decimals for the token in input string")
   const integerPart = new BN(match[1])
   const decimalPart = new BN(decimalString)
   const representation = integerPart.mul(bnTen.pow(new BN(digits))).add(decimalPart)
--- a/scripts/utils/printing_tools.js
+++ b/scripts/utils/printing_tools.js
@@ -10,7 +10,7 @@ const bnMaxUint = bnTwo.pow(bn256).sub(bnOne)
  * If the decimal representation has more decimals than the maximum amount possible, then the extra decimals are truncated.
  * @param {string} amount User-friendly representation for the amount of some ERC20 token
  * @param {integer} decimals Maximum number of decimals of the token
- * @return {string} decimal representation of the number of token units corresponding to the input amount
+ * @return {BN} number of token units corresponding to the input amount
  */
 const fromUserToMachineReadable = function (amount, decimals) {
   if (decimals < 0 || decimals >= 256) throw Error("Invalid number of decimals for ERC20 token: " + decimals.toString()) // ERC20 decimals is stored in a uint8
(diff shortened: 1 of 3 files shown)


scripts/utils/printing_tools.js
```diff
@@ -0,0 +1,52 @@
...
+const bnTen = new BN(10)
+const bn256 = new BN(256)
+const bnMaxUint = bnTwo.pow(bn256).sub(bnOne)
+
+/**
+ * A generalized version of "toWei" for tokens with an arbitrary amount of digits.
+ * If the decimal representation has more digits than the maximum amount possible, then the extra digits are truncated.
+ * @param {string} amount User-friendly representation for the amount of some ERC20 token
+ * @param {integer} digits Maximum number of decimal digits of the token
+ * @return {string} decimal representation of the number of token units corresponding to the input amount
```
josojo: I think we should return BN numbers and calculate with BN numbers in all our scripts

scripts/utils/printing_tools.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+const fromUserToMachineReadable = function (amount, digits) {
+  if (digits < 0 || digits > 256) throw Error("Invalid number of digits for ERC20 token: " + digits.toString())
+  const re = /^(\d+)(\.(\d+))?$/ // a sequence of at least one digit (0-9), followed by optionally a dot and another sequence of at least one digit
```
josojo: Many numbers get converted automatically into an e representation 1e-4 for example

scripts/utils/printing_tools.js
```diff
@@ -0,0 +1,52 @@
...
+  const match = re.exec(amount)
+  if (match == null) throw Error("Failed to parse decimal representation of " + amount)
+  const decimalString = (match[3] || "0").padEnd(digits, "0")
+  const integerPart = new BN(match[1])
+  const decimalPart = new BN(decimalString)
+  const representation = integerPart.mul(bnTen.pow(new BN(digits))).add(decimalPart)
```
josojo: Why is only the integer part shifted by a power of ten? Shouldn't it be the whole number?

scripts/utils/printing_tools.js
```diff
@@ -0,0 +1,52 @@
...
+  if (representation.gt(bnMaxUint)) throw Error("Number larger than ERC20 token maximum amount")
+  return representation.toString()
+}
+
+/**
+ * A generalized version of "fromWei" for tokens with an arbitrary amount of digits.
+ * @param {string} amount Decimal representation of the (integer) number of token units 
+ * @param {integer} digits Maximum number of decimal digits of the token
+ * @return {string} Dot-separated decimal representation of the amount of token corresponding to the input unit amount
```
josojo: I think we should just return a  floattype and then we can use the existing print functions to make it readable.

scripts/utils/printing_tools.js
fedgiac: This converts user input to token internal representation. So, if the token has 3 decimals, this parses:
"12.01" -> "12010"
This script works in this way:
1. split number into integer part ("12") and decimal part ("01")
2. pad decimal part with zeroes to the right to get as many digits as there are decimals ("01" -> "010")
3. convert both to BN ("12", "010" -> 12, 10)
4. the integer part is "worth" 10^digits token units, while thanks to the padding the decimal part already represents units (so total amount of unit in the example is 12*10^3 + 10 = 12010)
5. convert to string ("12010")

All the rest of the code is just error checks and the special case of a token with zero decimals.

fedgiac:
The use cases I'm thinking about are:
1. when we get an ERC20 token amount as input from the user and we want an exact representation of it
2. when we want to output an exact amount of tokens in a user-friendly way
I can see this code being reused in other projects, so I'm trying to be careful.

I'm very much against using float because of approximation errors. This way the function can be use when exact amounts matter. The only price we pay because of this choice is that we lose a printing function that specifies the number of decimals to print, which I could implement as an extra function if desired. As an added bonus we'll never encounter the exponential float representation when handling large or small numbers ("1e+20").

fedgiac:
I was unsure whether to use BN to represent amounts in token units. I went for strings since we often convert BNs back to strings when they are use as input to some contract function. However it's not a crucial choice, so I'm very much open to change the output type.

scripts/utils/printing_tools.js
```diff
@@ -0,0 +1,55 @@
...
+const bnMaxUint = bnTwo.pow(bn256).sub(bnOne)
+
+/**
+ * A generalized version of "toWei" for tokens with an arbitrary amount of decimals.
+ * If the decimal representation has more decimals than the maximum amount possible, then the extra decimals are truncated.
+ * @param {string} amount User-friendly representation for the amount of some ERC20 token
+ * @param {integer} decimals Maximum number of decimals of the token
+ * @return {string} decimal representation of the number of token units corresponding to the input amount
+ */
+const fromUserToMachineReadable = function (amount, decimals) {
```
josojo: This function seems to be quite complicated.

Would the following work as well?

```
return BN((amount.toString()+`e`+ decimals).toLocaleString('fullwide', {useGrouping:false}))
```

josojo: Maybe the function names like `convertToNumberWithDecimalsDigits`, 'toGeneralizedWEI' or `convertToWeb3Representation` would be a better fit.

scripts/utils/printing_tools.js
josojo: okay thanks... I missed the smart padding.

josojo:
Okay, now I see the value of the function. 

The conversion logic seems quite complex for rather simple conversions. But if my proposal does not work, then we should go with this more complex code.

josojo:
>. However it's not a crucial choice, so I'm very much open to change the output type.

the original toWEI function also returns a BN. I think it should be a BN, exactly as the original `formatAmount` function in our code is also returning a BN. It is the most convenient if we want to do further calculations.

scripts/utils/printing_tools.js
fedgiac: Mmh, I wasn't able to make it work, but I'd expect it has the same precision problem, since the \`e\` notation hints to some float conversion taking place.
I expect a bad input to be: "222.222222222222222002", 18 decimals.
In fact, running the node console:
```
> parseFloat("222.222222222222222002e18").toString()
  '222222222222222230000'
```

josojo: okay, I was also not able to find a better solution. Let's go with yours

test/printingTools.js
```diff
@@ -0,0 +1,409 @@
...
+            .slice(-18),
+        decimals: 18,
+        error: "tooLargeNumber",
+      },
+    ]
+    testBadEntries(badEntries)
+  })
+})
+
+describe("fromMachineToUserReadable", () => {
```
josojo: Can you update the name here?

test/printingTools.js
```diff
@@ -0,0 +1,409 @@
...
+
+const invalidDecimals = function(decimals) {
+  return "Invalid number of decimals for ERC20 token: " + decimals.toString()
+}
+const tooLargeNumber = function() {
+  return "Number larger than ERC20 token maximum amount (uint256)"
+}
+
+describe("fromUserToMachineReadable", () => {
+  const invalidNumber = function(amount) {
```
josojo: Can we put all the error messages outside of the describe block? Alternatively, you could put them all inside...

test/printingTools.js
```diff
@@ -0,0 +1,409 @@
+// This test can be run directly using mocha:
+// $ npx mocha test/printingTools.js
+// Using truffle test works but it's much slower, since it needs to compile contracts and deploy them to the blockchain.
+
+const assert = require("assert")
+const BN = require("bn.js")
+
+const { toErc20Units, fromErc20Units, bnMaxUint, bnOne } = require("../scripts/utils/printing_tools")
+
+const goodTwoWayPairs = [
```
josojo: This is the only test sample arranged outside of the actual unit test. Can you clean this?

fedgiac: This array is used by both `describe`s, so I'm not sure how to to include it in a unit test without duplicating the code.

PR #64, opened 2020-03-09
Print automatically generated deposit intructions to a file

fedgiac:


scripts/utils/trading_strategy_helpers.js
```diff
@@ -378,35 +380,68 @@ const buildTransferApproveDepositTransaction = async function(
...
+      userAddress: slaves[i]
+    }
+    deposits.push(deposit)
     fundingTransaction = fundingTransaction.concat(
       await calculateTransactionForTransferApproveDeposit(
         masterSafeAddress,
-        stableTokenAddress,
-        slaves[i],
-        investmentStableToken.div(new BN(FleetSizeDiv2)),
+        deposit.tokenAddress,
```
josojo: can we use ...deposit, if the deposit object is arranged differently? I would be required to be arranged like this: tokenAddress, userAddress, amount

PR #66, opened 2020-03-10
Code cleanup

fedgiac:
This PR makes code better to read and reuse. Among other things:
1. Uniform notation for master and bracket safes (no more "fleetOwner", "trader", "slave", "subsafe", ...)
2. Functions that create transactions for something are consistently called buildSomething
3. Many small corrections in the documentation
4. Isolate snippets of repeating code (address shortening, check ownership)
5. From and toErc20Unit functions accept more types as input (step for another PR: use these functions everywhere it's needed in the code)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -47,33 +48,46 @@ const maxUINT = new BN(2).pow(new BN(256)).sub(new BN(1))
...
   return new BN(10).pow(new BN(token.decimals)).muln(amount)
 }
 
+/**
+ * Checks that the first input address is the only owner of the first input address
+ * @param {Address} masterAddress address that should be the only owner
+ * @param {Address} ownedAddress address that is owned
+ * @return {bool} whether ownedAddress is indeed owned only by masterAddress
+ */
+const isOnlyOwner = async function (masterAddress, ownedAddress, artifacts) {
```
josojo: Since this is so specific to the safe, isOnlySaveOwner might be a little bit better name.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -371,21 +375,20 @@ const formatDepositString = function (depositsAsJsonString) {
...
  * @param {number} investmentStableToken Amount of stable tokens to be invested (in total)
+ * @param {Address} targetTokenAddress second token to be traded in bracket strategy
  * @param {number} investmentStableToken Amount of target tokens to be invested (in total)
  * @param {bool} storeDepositsAsFile whether to write the executed deposits to a file (defaults to false)
  * @return {Transaction} all the relevant transaction information to be used when submitting to the Gnosis Safe Multi-Sig
  */
-const buildTransferApproveDepositTransaction = async function(
-  masterSafeAddress,
-  slaves,
+const buildTransferApproveDepositFromOrders = async function(
```
josojo: Maybe the two function `buildTransferApproveDeposit` and `buildTransferApproveDepositFromOrders` could be renamed to `buildTransferApproveDepositWithStandartFunding` and `buildTransferApproveDepositWithCustomFunding`

would you think that this is easier to understand?

scripts/utils/trading_strategy_helpers.js
```diff
@@ -394,44 +397,44 @@ const buildTransferApproveDepositTransaction = async function(
...
 ) {
-  const fleetSize = slaves.length
+  const fleetSize = bracketAddresses.length
   assert(fleetSize % 2 == 0, "Fleet size must be a even number")
   const deposits = []
 
   let fundingTransaction = []
-  const FleetSizeDiv2 = fleetSize / 2
-  for (const i of Array(FleetSizeDiv2).keys()) {
+  const fleetSizeDiv2 = fleetSize / 2
```
josojo: It would be awesome, if we could first generate the deposit file and then call the `buildTransferApproveDepositWithFunding` function with this deposit file. Less code duplication. But let's do it in a separate PR

scripts/utils/trading_strategy_helpers.js
fedgiac: I think `buildTransferApproveDepositFromOrders` is more specific than `buildTransferApproveDepositWithCustomFunding` so I would not change it.
However i also don't like the current combination of the two names. What about renaming `buildTransferApproveDeposit` to a more specific `buildTransferApproveDepositFromList`?

fedgiac:
Sorry about the huge PR, I found new little things to work on every time I checked the code and wasn't able to stop! I'd say we could merge here and move further progress in code polishing to another PR.

Issue #57, opened 2020-03-04
Slow creation of transactions


fedgiac:
Some of the scripts are unnecessarily slow, since they serially query the blockchain to get miscellaneous information about the operations being performed. Many of those queries can be parallelized.

fedgiac:
Most interactions with the blockchain were made quicker in PRs #72, #71, #81, #118, #113, and #116.

PR #72, opened 2020-03-12
Faster withdrawal scripts

fedgiac:
Drastically reduce the running time of the withdrawal scripts by heavily relying on Javascript's asynchronous features.

This PR is a test bench for using the same techniques everywhere else in this repo, so changes and recommendations are very welcome before we start working on the other scripts.

Notable breaking change is the function `fetchTokenInfo`, which is now split in `fetchTokenInfoAtAddresses` and `fetchTokenInfoFromExchange`, both of which return a promise.

Running times of withdraw.js
1. with two safes
  - before: 28±1s
  - after: 12±1s

2. with 20 safes
  - before: 3min 6s (no ±, I ran it only once)
  - after: 14.7±0.1s

Related issue: #57 
Thanks @nlordell for the discussion.

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -572,7 +572,7 @@ const buildTransferFundsToMaster = async function(masterAddress, withdrawals, li
   const ERC20 = artifacts.require("ERC20Mintable")
 
   // TODO: enforce that there are no overlapping withdrawals
-  const masterTransactionsPromises = withdrawals.map((withdrawal) => (async () => {
+  const masterTransactions = await Promise.all(withdrawals.map(async withdrawal => {
     const token = await ERC20.at(withdrawal.tokenAddress)
     let amount
     if (limitToMaxWithdrawableAmount) {
@@ -592,10 +592,7 @@ const buildTransferFundsToMaster = async function(masterAddress, withdrawals, li
     }
     // build transaction to execute previous transaction through master
     return buildExecTransaction(masterAddress, withdrawal.bracketAddress, transactionToExecute, artifacts)
-  }).call())
-
-  const masterTransactions = []
-  for (const transactionPromise of masterTransactionsPromises) masterTransactions.push(await transactionPromise)
+  }))
 
   return buildBundledTransaction(masterTransactions, web3, artifacts)
 }
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -71,6 +71,21 @@ const maxUINT = new BN(2).pow(new BN(256)).sub(new BN(1))
  */
 
 /**
+ * @typedef TokenObject
+ *  * Example:
+ * {
+ *   symbol: "WETH",
+ *   decimals: 18,
+ *   tokenAddress: "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
+ * }
+ * @type {object}
+ * @property {string} symbol symbol representing the token
+ * @property {(number|BN)} decimals number of decimals of the token
+ * @property {address} [tokenAddress] address of the token contract on the EVM
+
+ */
+
+/**
  * Returns an instance of the exchange contract
  */
 const getExchange = async function(web3) {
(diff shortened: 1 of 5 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -85,37 +116,60 @@ const isOnlySafeOwner = async function(masterAddress, ownedAddress, artifacts) {
...
-        globalTokenObjects[id] = tokenInfo
-      }
-      tokenObjects[id] = globalTokenObjects[id]
-    })
-  )
-  return tokenObjects
+        log(`Found token ${tokenInfo.symbol} at address ${tokenInfo.address} with ${tokenInfo.decimals} decimals`)
+        return tokenInfo
+      }).call()
+      tokenPromises[tokenAddress] = globalTokenPromisesFromAddress[tokenAddress]
```
nlordell: should this be outside the `if` statement?

scripts/utils/trading_strategy_helpers.js
```diff
@@ -85,37 +116,60 @@ const isOnlySafeOwner = async function(masterAddress, ownedAddress, artifacts) {
...
+}
+
+/**
+ * Queries EVM for ERC20 token details by token id
+ * and returns a list of detailed token information.
+ * @param {SmartContract} exchange BatchExchange, contract, or any contract implementing `tokenIdToAddressMap`
+ * @param {integer[]} tokenIds list of *unique* token ids whose data is to be fetch from EVM
+ * @return {Promise<TokenObject>[]} list of detailed/relevant token information
+ */
+const globalTokenPromisesFromId = {}
```
nlordell: I think as-is, the doc comment is for this constant instead of the `fetchTokenInfoFromExcahnge` function.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -85,37 +116,60 @@ const isOnlySafeOwner = async function(masterAddress, ownedAddress, artifacts) {
...
- * @param {SmartContract} exchange BatchExchange, contract, or any contract implementing `tokenIdToAddressMap`
- * @param {integer[]} tokenIds list of token ids whose data is to be fetch from EVM
- * @return {TokenObject[]} list of detailed/relevant token information
+ * and returns a list of promises of detailed token information.
+ * @param {Address[]} tokenAddresses list of *unique* token addresses whose data is to be fetch from the EVM
+ * @return {Promise<TokenObject>[]} list of detailed/relevant token information
  */
-const globalTokenObjects = {}
-const fetchTokenInfo = async function(exchange, tokenIds, artifacts, debug = false) {
+const globalTokenPromisesFromAddress = {}
```
nlordell: I think the doc comment is for the constant and not the function how it is currently.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -85,37 +116,60 @@ const isOnlySafeOwner = async function(masterAddress, ownedAddress, artifacts) {
...
-    })
-  )
-  return tokenObjects
+        log(`Found token ${tokenInfo.symbol} at address ${tokenInfo.address} with ${tokenInfo.decimals} decimals`)
+        return tokenInfo
+      }).call()
+      tokenPromises[tokenAddress] = globalTokenPromisesFromAddress[tokenAddress]
+    }
+  }
+  return tokenPromises
```
nlordell: Is there a reason why this method isn't async and doesn't `await` all the promises in here instead of having `await tokenPromises[tokenAddress]` for each token after the function call?

scripts/utils/trading_strategy_helpers.js
```diff
@@ -505,7 +559,7 @@ const buildRequestWithdraw = async function(masterAddress, withdrawals, web3, ar
  * @return {Transaction} Multisend transaction that has to be sent from the master address to withdraw the desired funds
  */
 const buildWithdraw = async function(masterAddress, withdrawals, web3, artifacts) {
```
nlordell: This doesn't have to be `async` as there is no `await`.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -491,7 +545,7 @@ const buildBracketTransactionForTransferApproveDeposit = async (
 withdrawal of the desired funds
 */
 const buildRequestWithdraw = async function(masterAddress, withdrawals, web3, artifacts) {
```
nlordell: I don't think this has to be `async`.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -537,10 +591,13 @@ const buildTransferFundsToMaster = async function(masterAddress, withdrawals, li
...
     }
     // build transaction to execute previous transaction through master
-    const execTransaction = await buildExecTransaction(masterAddress, withdrawal.bracketAddress, transactionToExecute, artifacts)
-    masterTransactions.push(execTransaction)
-  }
-  return await buildBundledTransaction(masterTransactions, web3, artifacts)
+    return buildExecTransaction(masterAddress, withdrawal.bracketAddress, transactionToExecute, artifacts)
+  }).call())
+
+  const masterTransactions = []
```
nlordell: This can be re-written as `await Promise.all(masterTransactionPromises)`.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -275,14 +329,11 @@ const checkSufficiencyOfBalance = async function(token, owner, amount) {
...
-  // it's not necessary to avoid overlapping withdraws, since the full amount is withdrawn for each entry
-  for (const withdrawal of withdrawals) {
+const buildGenericFundMovement = async function(masterAddress, withdrawals, functionName, web3 = web3, artifacts = artifacts) { // TODO: do we need artifacts here?
+  const exchange = await getExchange(web3)
+  
+  // it's not necessary to avoid overlapping withdraws, since the full amount is withdrawn for each entry TODO: is this still true?
+  const masterTransactionsPromises = withdrawals.map((withdrawal) => (async () => {
     // create transaction for the token
     let transactionData
     switch (functionName) {
```
nlordell: Does `encodeABI()` really return a promise?

scripts/utils/trading_strategy_helpers.js
```diff
@@ -275,14 +329,11 @@ const checkSufficiencyOfBalance = async function(token, owner, amount) {
...
-  BatchExchange.setNetwork(web3.network_id)
-  const exchange = await BatchExchange.deployed()
-  const masterTransactions = []
-
-  // it's not necessary to avoid overlapping withdraws, since the full amount is withdrawn for each entry
-  for (const withdrawal of withdrawals) {
+const buildGenericFundMovement = async function(masterAddress, withdrawals, functionName, web3 = web3, artifacts = artifacts) { // TODO: do we need artifacts here?
+  const exchange = await getExchange(web3)
+  
+  // it's not necessary to avoid overlapping withdraws, since the full amount is withdrawn for each entry TODO: is this still true?
```
josojo: Yeah, I believe this is still true, as the smart contract logic has not changed:
https://github.com/gnosis/dex-contracts/blob/cc7e166d1ed3430ffd8c06adf757412699fa2d04/contracts/EpochTokenLocker.sol#L111

It might just be more efficient to check for double/overlapping withdraw requests beforehand.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -550,12 +607,16 @@ const buildTransferFundsToMaster = async function(masterAddress, withdrawals, li
  * @return {Transaction} Multisend transaction that has to be sent from the master address to transfer back the funds stored in the exchange
  */
 const buildWithdrawAndTransferFundsToMaster = async function(masterAddress, withdrawals, web3 = web3, artifacts = artifacts) {
-  const withdrawalTransaction = await buildWithdraw(masterAddress, withdrawals, web3, artifacts)
-  const transferFundsToMasterTransaction = await buildTransferFundsToMaster(masterAddress, withdrawals, false, web3, artifacts)
+  const [withdrawalTransaction, transferFundsToMasterTransaction] = await Promise.all([
```
josojo: Sometimes, I think that the original statement is much easier to read and likely the wins in waiting time might be minimal

scripts/utils/trading_strategy_helpers.js
fedgiac: Yes, thanks for spotting that! I'll write a test to catch this later.

scripts/utils/trading_strategy_helpers.js
fedgiac: Mmh, currently none. The original idea was that I was going to fetch any token information that I might need later on at the start of the code, so I structured the function around this. But this is indeed not the case right now. I just tried to move the token fetching at the start of the code, and the running time of the withdrawal script went from about 11s to 8s! So I'm going to keep this function as it is and to change the script accordingly instead.

scripts/utils/trading_strategy_helpers.js
fedgiac: No, you are right.
Await is used everywhere in this project when `encodeABI` is invoked, I'll fix that in another PR since it requires some other changes (hopefully, no need to retrieve MultiSend and GnosisSafeMasterCopy from the network anymore, I'll think about it.).

scripts/utils/trading_strategy_helpers.js
fedgiac: True. I also agree it makes sense to check, even just to tell the user that the deposits/withdrawals don't look right.

scripts/utils/trading_strategy_helpers.js
fedgiac: I tested it and I didn't see any meaningful impact in the running time (the difference, if it exists, is less than a second). I'll restore this function back to what it was.

fedgiac:
Thanks for the comments! I also noticed that these changes introduce a bug in the logging: the function console.log.apply is apparently not thread safe, and the result is that instead of returning some output it just prints an empty line.
Since logging is a bit messing currently, I'd say we could merge anyway and then take care of the logging in another PR.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -70,6 +70,38 @@ const maxUINT = new BN(2).pow(new BN(256)).sub(new BN(1))
...
+ * @property {string} symbol symbol representing the token
+ * @property {(number|BN)} decimals number of decimals of the token
+ * @property {address} [tokenAddress] address of the token contract on the EVM
+
+ */
+
+/**
+ * Returns an instance of the exchange contract
+ */
+const getExchange = async function(web3) {
```
josojo: does this function needs to be async, if we don't await the promise, just return the promise

fedgiac:
I checked a previous commit: logging was already broken back then in the same way as it is with this PR, this PR (as well as the previous one changing `fetchTokenInfo`) didn't introduce this problem.

fedgiac:
The tiny adaptation requests are very appreciated!

PR #76, opened 2020-03-16
View scripts

### LINKED ISSUE #31: View Scripts
One might like to be able to view the Open Orders, or account balances for an individual bracket strategy. Might be nice to have a script like

```
truffle exec view_fleet_status --masterSafe
```

Which displays all balances, open orders and subsafes... Unfortunately, it might not be so easy to determine all information without providing a few more keyword (possibly the list of subsafes) because we have already deployed two different bracket strategies under the same master safe.

josojo:
closes #31 

This PR adds two scripts:
- find_bracket-traders, which investigates all deployed safes and checks whether the masterSafe is the only owner.
- get_bracket_data that plots the orders and balances of the bracket-trader

Actually, I do not find these scripts too helpful, especially since they are both soo slow. I will investigate whether we can get the current data from the graph quicker.

testplan:
```
npx truffle exec scripts/get_bracket_data.js --brackets 0x1E800A7E4b103cA2C270e162E66d1B099eC8C1A7,0x65539d12c28946831D65Bb32eEc3A33CaBFB9249  --network mainnet
```

### DIFF
--- a/scripts/get_bracket_data.js
+++ b/scripts/get_bracket_data.js
@@ -1,17 +1,6 @@
-const { deployFleetOfSafes } = require("./utils/trading_strategy_helpers")
 const Contract = require("@truffle/contract")
-const {getOrdersPaginated, printOrder} = require("./utils/to_be_imported_from_dex_contracts.js")
-const {
-  buildTransferApproveDepositFromOrders,
-  buildOrders,
-  checkSufficiencyOfBalance,
-  isOnlySafeOwner,
-} = require("./utils/trading_strategy_helpers")
-const BN = require("bn.js")
-
-const { signAndSend, promptUser } = require("./utils/sign_and_send")
-const { toErc20Units } = require("./utils/printing_tools")
-const assert = require("assert")
+const { getOrdersPaginated, printOrder } = require("./utils/to_be_imported_from_dex_contracts.js")
+const { isOnlySafeOwner } = require("./utils/trading_strategy_helpers")
 
 const argv = require("yargs")
   .option("masterSafe", {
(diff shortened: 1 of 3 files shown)


fedgiac:
I think it's very important we find another way to find out all the bracket traders of an address, especially so if we plan to release the scripts to the public.

The best idea I could come up with is to fork the Gnosis Safe proxy to one that behaves in exactly the same way but fires an additional event on Safe creation. This event will contain the owner. (And some other data? Like some designation as "bracket trader"?) Then we can filter all Safe creation events based on the owner, which should reduce a lot the number of Safes available. This is however a breaking change and requires the deployment of a new contract.

For those we have deployed so far, we could hardcode the block number of the first and last Safes we have deployed without the extra event and check all Safes. We can also hardcode the addresses we used to deploy these Safes and filter the event with that.

scripts/get_bracket_data.js
```diff
@@ -0,0 +1,72 @@
...
+    const ProxyFactory = artifacts.require("GnosisSafeProxyFactory.sol")
+
+    const proxyFactory = await ProxyFactory.deployed()
+    const eventData = await proxyFactory.getPastEvents("ProxyCreation", { fromBlock: 0, toBlock: "latest" })
+    const safeDeployments = eventData.map(event => event.args[0]).slice(-100)
+    //For easy rinkeby testing use:
+    //bracketTraderAddresses = ['0xc867f926392a4e55b8800afbbbad8c99efd70e49']
+    console.log("In total we have ", safeDeployments.length, " safe deployments")
+    const isDeployedByMaster = await Promise.all(
+      safeDeployments.map(async bracketTrader => await isOnlySafeOwner(argv.masterSafe, bracketTrader, artifacts))
```
fedgiac: I believe that `async` and `await` are not needed here.
Interestingly, I tested the time this takes to run with both `.slice(-100)` and `.slice(-200)`, and the time goes up by a factor of about two. However, we are sending the calls in parallel: if you replace this instruction with a for loop it takes about 15 times more. I wonder if we are hitting some request limit on Infura?

josojo:
Writing this kind of script is not the right approach. I would prefer to go with dune, as its performance is better than requesting all the data via infura.

bh2smith:
> Writing this kind of script is not the right approach. I would prefer to go with dune, as its performance is better than requesting all the data via infura.

@josojo, does this mean we should also close the related issue?

PR #77, opened 2020-03-16
Fix extension of automatically generated deposit file

fedgiac:


josojo:
Thanks for remembering this issue!

PR #80, opened 2020-03-16
Fix missing await from withdrawal scripts

fedgiac:


scripts/withdraw.js
```diff
@@ -81,7 +81,7 @@ const getAmount = async function(bracketAddress, tokenAddress, exchange) {
 module.exports = async callback => {
   try {
     const masterSafePromise = getSafe(argv.masterSafe, artifacts)
```
josojo: I think it would be cleaner to await the promise here already and store the variable the masterSafe. Or do we have any gains by storing only the promise?

fedgiac: The Safe is needed only at the very end, so loading it in the background from the start is a little efficiency gain. I never measured the time this saves, but it should be a the length of a full network request.

PR #81, opened 2020-03-17
Optimize order creation

fedgiac:
This applies the techniques from PR #72 to order creation.

We could prepare the orders while the user is deciding what to answer in the price check failed prompt. The reason I decided not to do it is logging: if I directly did that with the current code, the user would see the prompts and then information for each order being printed on top of the prompt.
Possible solutions are:
1. Separate order generation and order printing. With this solution we risk that the printed output doesn't match the real order.
2. Isolate the creation of order parameters is a single function and use this function directly to print order information. But this looked messy and redundant.
3. Return log as part of the output. This is very inconsistent with the rest of the code and should be done in case in a different PR.
4. Have a global "log" object that is retrieved when needed. This would create a new global object to keep track of, and would probably become confusing to use very quickly.

How do we want to tackle this?

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -225,13 +225,11 @@ const buildOrders = async function(
 ) {
   const log = debug ? (...a) => console.log(...a) : () => {}
 
-  await BatchExchange.setProvider(web3.currentProvider)
-  await BatchExchange.setNetwork(web3.network_id)
-  const exchange = await BatchExchange.deployed()
+  const exchange = await getExchange(web3)
   log("Batch Exchange", exchange.address)
 
-  const batch_index = (await exchange.getCurrentBatchId.call()).toNumber()
-  const tokenInfoPromises = await fetchTokenInfoFromExchange(exchange, [targetTokenId, stableTokenId], artifacts)
+  const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, [targetTokenId, stableTokenId], artifacts)
+  const batchIndexPromise = exchange.getCurrentBatchId.call()
 
   const targetToken = await tokenInfoPromises[targetTokenId]
   const stableToken = await tokenInfoPromises[stableTokenId]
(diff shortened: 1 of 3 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -244,34 +242,30 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
       // target_token against standard_token. Hence the buyAmounts and sellAmounts are switched in the next line.
       const [lowerBuyAmount, lowerSellAmount] = calculateBuyAndSellAmountsFromPrice(lowerLimit, targetToken)
 
-      log(`Safe ${bracketIndex} - ${bracketAddress}:\n  Buy  ${targetToken.symbol} with ${stableToken.symbol} at ${lowerLimit}`)
-      log(`  Sell ${targetToken.symbol} for  ${stableToken.symbol} at ${upperLimit}`)
+      log(`Safe ${bracketIndex} - ${bracketAddress}:\n  Buy  ${targetToken.symbol} with ${stableToken.symbol} at ${lowerLimit}\n  Sell ${targetToken.symbol} for  ${stableToken.symbol} at ${upperLimit}`)
       const buyTokens = [targetTokenId, stableTokenId]
       const sellTokens = [stableTokenId, targetTokenId]
-      const validFroms = [batch_index + validFrom, batch_index + validFrom]
+      const validFroms = [await batchIndexPromise + validFrom, await batchIndexPromise + validFrom]
```
josojo: It would be good to await this promise just once for performance reasons. Actually, I would prefer to await the batchIndex only once for all orders, so that they all have the same validity.

fedgiac: I believe that awaiting an already resolved promise is very efficient, and an improvement would most likely be non noticeable by the user (my guess is <1ms, but I didn't test it).
Moreover, the output of resolving the same promise should always be the same for all orders, independently of when each of the promises are awaited, so is should never happen that the batchIndices are different.

PR #82, opened 2020-03-17
Clean up artifacts and web3

fedgiac:
Most functions' input is cluttered with `artifacts` and `web3`s. This PR remove most of them.

Instead of directly returning the functions we need, `require` is transformed into a function taking `web3` and `artifacts` as input. This input is then used by the imported functions without any need to inject it into the functions' arguments.

Not much is happening in this PR, most changes are indentation changes.

Still to do:
1. test that the scripts are working correctly.

### DIFF
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -3,7 +3,7 @@ const BatchExchange = Contract(require("@gnosis.pm/dex-contracts/build/contracts
 const { buildOrders } = require("./utils/trading_strategy_helpers")
 const { isPriceReasonable } = require("./utils/price-utils.js")(web3, artifacts)
 const { proceedAnyways } = require("./utils/user-interface-helpers")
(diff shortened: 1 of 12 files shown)


scripts/bracket_orders.js
```diff
@@ -1,9 +1,9 @@
 const Contract = require("@truffle/contract")
 const BatchExchange = Contract(require("@gnosis.pm/dex-contracts/build/contracts/BatchExchange"))
-const { buildOrders } = require("./utils/trading_strategy_helpers")
-const { isPriceReasonable } = require("./utils/price-utils.js")
+const { buildOrders } = require("./utils/trading_strategy_helpers")(web3, artifacts)
+const { isPriceReasonable } = require("./utils/price-utils.js")(web3, artifacts)
```
josojo: Is web3 here needed?

PR #84, opened 2020-03-18
Activating prettier

josojo:
I am highly in favor of prettier enforced by the travis. Let me know whether you are okay with that @fedgiac 

Let's only merge this PR, if all other PRs have been merged, otherwise, it will be a mess :)


fedgiac:
I agree with enforcing prettier, but I think it's better to reject this PR: PR #82 changes a lot of the code because of indentation changes, which means that there will definitely be many conflicts. It might be easier to redo this PR after #82 is merged.

PR #87, opened 2020-03-18
Introducing lowest and highestLimit for bracket strategy

josojo:
- introduces the highest and lower Limit for bracket strategy
- make a sanity check for the highest and lowest bounds ( to be included in PR)
todo:
- adapt the funding of the bracket in compelte_liquidity_provision (separate PR)


scripts/utils/price-utils.js
```diff
@@ -2,6 +2,10 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const { fetchTokenInfoFromExchange } = require("./trading_strategy_helpers")(web3, artifacts)
 
+  const areBoundsReasonable = function(targetPrice, lowestLimit, highestLimit) {
+    return targetPrice * 0.5 < lowestLimit && highestLimit < targetPrice * 1.5
+  }
+
```
fedgiac: Two comments:
1. the use is prompted to confirm because the price check failed but has no feedback on why. We should probably print a line explaining which check failed.
2 The range is not really symmetric, if you accept that the price goes down by a factor of two then arguably you accept that your price goes up by a factor of two.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -237,13 +237,11 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     assert(targetToken.decimals === 18, "Stable tokens must have 18 decimals")
 
     // Number of brackets is determined by bracketAddresses.length
-    const lowestLimit = targetPrice / (1 + priceRangePercentage / 100)
-    const highestLimit = targetPrice * (1 + priceRangePercentage / 100)
     log(`Lowest-Highest Limit ${lowestLimit}-${highestLimit}`)
```
fedgiac: Maybe we don't need this line anymore, since we print the limits in the following log expression?

scripts/complete_liquidity_provision.js
```diff
@@ -81,13 +81,17 @@ module.exports = async callback => {
...
-    if (!(priceCheck && boundCheck)) {
+    if (!priceCheck) {
       if (!(await proceedAnyways("Price check failed!"))) {
         callback("Error: Price checks did not pass")
       }
     }
+    const boundCheck = areBoundsReasonable(argv.targetPrice, argv.lowestLimit, argv.highestLimit)
+    if (!boundCheck) {
+      if (!(await proceedAnyways("Bound check failed!"))) {
+        callback("Error: Price checks did not pass")
```
fedgiac: One last nit: we could use "Error: _Bound_  checks did not pass" instead. Otherwise feel free to merge.

PR #88, opened 2020-03-18
Remove leftover artifacts

fedgiac:
This bug fixes the issue with the extra empty lines when running the tests.


fedgiac:
Because `fetchTokenInfoFromExchange` takes an extra input, `debug`. Before it defaulted to false, so there was no output from the log function, but after I removed `artifacts` from the function definitions but left it in the function call then I implicitly set `debug=artifacts`, which is avaluated as `true` in the if loop. Removing `artifacts` restores the default value of `debug`.

PR #89, opened 2020-03-19
Allow amount of decimals different from 18 in deposits

fedgiac:


scripts/utils/trading_strategy_helpers.js
```diff
@@ -292,7 +289,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     // Sell x ETH for max256 DAI
     // x = max256 / 102
     // priceFormatted = 102000000000000000000
-    price = price.toFixed(18)
+    price = price.toFixed(targetToken.decimals)
```
fedgiac: This changes the logic in an incorrect way. I'm not going to adapt this function to avoid conflicts with PR #105, but there should be a test somewhere telling me that changing this value is wrong.

test/dfusionMultiSafes.js
```diff
@@ -331,6 +341,106 @@ contract("GnosisSafe", function(accounts) {
...
+          highestLimit: 130,
+          currentPrice: 110,
+        }
+        const expectedDistribution = {
+          bracketsWithStableTokenDeposit: 3,
+          bracketsWithTargetTokenDeposit: 5,
+        }
+        for (const tokenSetup of tokenSetups) {
+          const tradeInfo = {}
+          for (const key in tradeInfoWithoutTokens) tradeInfo[key] = tradeInfoWithoutTokens[key]
```
josojo: Doesn't tradeInfo = tradeInfoWithouTokens work as well? There should be some copy command that does not require a for loop :)

fedgiac: I couldn't find a copy command, not even for a shallow copy! If you know of one I'm happy to use it.

josojo: I changed it. But I can't argue that it is strictly better than your solution.
Up to you, whether you want to keep it ... :)

scripts/utils/trading_strategy_helpers.js
```diff
@@ -495,7 +495,6 @@ withdrawal of or to withdraw the desired funds
     const transactions = []
 
     // log(`Deposit Token at ${depositToken.address}: ${tokenSymbol}`)
```
bh2smith: This should be removed.

scripts/withdraw.js
```diff
@@ -53,16 +53,15 @@ const argv = require("yargs")
   })
   .version(false).argv
 
-const getAmount = async function(bracketAddress, tokenAddress, exchange) {
+const getAmount = async function(bracketAddress, tokenInfo, exchange) {
```
bh2smith: I find the name of this function somewhat vague. What "amount" is being gotten here?

scripts/complete_liquidity_provision.js
```diff
@@ -4,6 +4,7 @@ const { sleep } = require("./utils/js_helpers")
 
 const Contract = require("@truffle/contract")
 const {
+  fetchTokenInfoFromExchange,
```
bh2smith: I guess this was defined in a previous PR. Why not fetch token info from EVM?

fedgiac: Maybe a better name would be `fetchTokenInfoByExchangeId`.

bh2smith: What is `ExchangeId`?

fedgiac: There are currently two functions fetching token information from the EVM: `fetchTokenInfoAtAddresses` and `fetchTokenInfoFromExchange`, the latter relying on the former during its execution. So "fromExchange" really just means that you get the addresses of the tokens from the exchange using the input token ids, addresses which are then used to fetch token info from the EVM. The output of the two functions has the same structure. If you have better naming proposals I'm happy to change the names (but this is a change that touches basically every file, so we might want to do that only when there are no other PRs standing).

bh2smith: Nah its all good. Thanks for clearing that up.

PR #90, opened 2020-03-19
new deposit strategy

josojo:
This deposit logic takes into account the current price:

If the sell-order of the targetToken of a bracket is below the current price, the bracket is funded with the stableToken, otherwise, it is funded with the targetToken.


scripts/utils/trading_strategy_helpers.js
```diff
@@ -429,30 +429,35 @@ withdrawal of or to withdraw the desired funds
...
+    targetPrice,
     investmentStableToken,
-    targetTokenAddress,
     investmentTargetToken,
     storeDepositsAsFile = false
   ) {
     const fleetSize = bracketAddresses.length
-    assert(fleetSize % 2 == 0, "Fleet size must be a even number")
+    const stepSizeAsMultiplier = Math.pow(highestLimit / lowestLimit, 1 / bracketAddresses.length)
+    const bracketIndexAtTargetPrice = Math.floor(Math.log(targetPrice / lowestLimit) / Math.log(stepSizeAsMultiplier))
```
fedgiac: `Math.round` is probably safer than `Math.floor`, so if the result of the division is 11.99999999 we return 12 instead of 11.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -429,30 +429,35 @@ withdrawal of or to withdraw the desired funds
...
+    assert(bracketIndexAtTargetPrice <= fleetSize, "TargetPrice is not between the bounds")
```
fedgiac: Shouldn't the assertion use just `<` instead? I'd also assert that `bracketIndexAtTargetPrice>=0`.

fedgiac: It arguably makes sense that the index is outside this range. This is the case if you want to anticipate a movement of the price and want to create your strategy outside of a reasonable range. For example, let's say that the current ETH price is 100 but you believe the price is 200 and want to prepare your brackets between 150 and 250. In this case it's reasonable that `bracketIndexAtTargetPrice` is outside the range, and all brackets are filled with only ETH. That's because `targetPrice` would not really be your "target" but rather the current price. It doesn't make sense to talk about a "target", since everything is defined around min and max of a range. Does this make sense? This would change the PR a bit.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -429,30 +429,35 @@ withdrawal of or to withdraw the desired funds
...
+
```
fedgiac: There are two assertions that we could make:
if `bracketIndexAtTargetPrice==0` then `investmentStableToken` must be zero, since no bracket will receive `stableToken`.
if `bracketIndexAtTargetPrice==fleetSize` then `investmentTargetToken` must be zero, since no bracket will receive `targetToken`.

test/dfusionMultiSafes.js
```diff
@@ -165,52 +167,136 @@ contract("GnosisSafe", function(accounts) {
       }
     })
 
-    it("transfers tokens from fund account through trader accounts and into exchange via automatic deposit logic", async () => {
+    it("transfers tokens from fund account through trader accounts and into exchange via automatic deposit logic, p> 1", async () => {
```
fedgiac: This test creates orders, but order creation is tested only after this test. Maybe we should move this test after the one testing orders and call it "test deposit creation from orders, p>1"? Same reasoning for the next test, p<1.

scripts/utils/price-utils.js
```diff
@@ -1,7 +1,54 @@
...
+    const bracketExchangeBalanceStableToken = (await exchange.getBalance(bracketAddress, stableToken.address)).toString()
+    const bracketExchangeBalanceTargetToken = (await exchange.getBalance(bracketAddress, targetToken.address)).toString()
+    const auctionElements = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders.call(bracketAddress))
+    const bracketOrders = auctionElements.filter(order => order.user.toLowerCase() == bracketAddress.toLowerCase())
+    const stableTokenId = await exchange.tokenAddressToIdMap.call(stableToken.address)
+    const sellStableTokenOrder = bracketOrders.filter(order => order.sellToken == stableTokenId)[0]
+
+    const targetTokenId = await exchange.tokenAddressToIdMap.call(targetToken.address)
+    const sellTargetTokenOrder = bracketOrders.filter(order => order.sellToken == targetTokenId)[0]
+
```
fedgiac: I'd consider three more checks:
1. `auctionElements.length==2`
2. `bracketOrders.filter(order => order.sellToken == stableTokenId) == 1`
2. `bracketOrders.filter(order => order.sellToken == targetTokenId) == 1`

test/dfusionMultiSafes.js
josojo: I think that we are checking in this test the transfer amounts and not the order creation. Hence I would keep it in the "  describe("transfer tests:", async function() { " block.
Or would you create another describe block like this:
 "  describe("transfer tests depended on order information:", async function() { "

scripts/utils/price-utils.js
```diff
@@ -1,7 +1,54 @@
...
+    if (checkThatOrderPriceIsBelowTarget(targetPrice, sellStableTokenOrder)) {
+      // checks whether price is in middle of bracket:
+      if (checkThatOrderPriceIsBelowTarget(1 / targetPrice, sellTargetTokenOrder)) {
+        // Then only the targetToken gets funded
+        assert.equal(bracketExchangeBalanceStableToken, 0)
+      } else {
+        assert.equal(bracketExchangeBalanceStableToken, investmentStableTokenPerBracket.toString())
+      }
+    } else {
+      assert.equal(bracketExchangeBalanceStableToken, 0)
```
fedgiac: `bracketExchangeBalanceStableToken` is string, so `0` should be `"0"`.

scripts/utils/price-utils.js
```diff
@@ -1,7 +1,54 @@
...
+    }
+
+    if (checkThatOrderPriceIsBelowTarget(1 / targetPrice, sellTargetTokenOrder)) {
+      assert.equal(bracketExchangeBalanceTargetToken, investmentTargetTokenPerBracket.toString())
+    } else {
+      assert.equal(bracketExchangeBalanceTargetToken, 0)
```
fedgiac: `bracketExchangeBalanceTargetToken` is string, so `0` should be `"0"`.

scripts/utils/price-utils.js
josojo: I does not really matter. The test is passing even without this, I guess under the hood it is transformed automatically.

scripts/utils/price-utils.js
```diff
@@ -1,13 +1,96 @@
...
+    bracketExchangeBalanceTargetToken,
+    investmentTargetTokenPerBracket,
+    investmentStableTokenPerBracket
+  ) {
+    // For the middle bracket the funding can go in either bracket
+    // it depends on closer distance from the currentPrice to the limit prices fo the bracket-traders
+    return (
+      (bracketExchangeBalanceStableToken == 0 &&
+        bracketExchangeBalanceTargetToken == investmentTargetTokenPerBracket.toString()) ||
+      (bracketExchangeBalanceTargetToken == 0 && bracketExchangeBalanceStableToken == investmentStableTokenPerBracket.toString())
```
fedgiac: `0` should be `"0"`.

scripts/utils/price-utils.js
```diff
@@ -1,13 +1,96 @@
...
+  const checkThatOrderPriceIsBelowTarget = function(currentPrice, order) {
+    const multiplicator = 1000000000
+    return new BN(currentPrice * multiplicator).mul(order.priceNumerator) > order.priceDenominator.mul(new BN(multiplicator))
+  }
+
+  const checkFundingInTheMiddleBracket = function(
+    bracketExchangeBalanceStableToken,
+    bracketExchangeBalanceTargetToken,
+    investmentTargetTokenPerBracket,
+    investmentStableTokenPerBracket
```
fedgiac: The order of the input values is a bit strange, is there a reason why this is not:
```
    bracketExchangeBalanceStableToken,
    bracketExchangeBalanceTargetToken,
    investmentStableTokenPerBracket,
    investmentTargetTokenPerBracket
```

test/dfusionMultiSafes.js
```diff
@@ -165,52 +167,200 @@ contract("GnosisSafe", function(accounts) {
...
-        const bracketPersonalTokenBalance = (await testToken.balanceOf(bracketAddress)).toNumber()
-        // This should always output 0 as the brackets should never directly hold funds
-        assert.equal(bracketPersonalTokenBalance, 0)
+    })
+    it("transfers tokens from fund account through trader accounts and into exchange via automatic deposit logic, p<1 && p>1", async () => {
+      const masterSafe = await deploySafe(gnosisSafeMasterCopy, proxyFactory, [lw.accounts[0], lw.accounts[1]], 2)
+      const fleetSize = 4
+      const lowestLimit = 0.8
+      const highestLimit = 1.2
+      const currentPrice = 1
```
fedgiac: I'd use 0.9: in this way the price is indeed <1 and also you test the case where the price is not in the middle of the range.

test/dfusionMultiSafes.js
fedgiac: Yeah, I was thinking to have a new block since this is not a "pure" deposit. But it's also fine as it is now.

PR #92, opened 2020-03-20
ValidFrom fix

josojo:
Federico found this issue


fedgiac:
Good fix! Do we want to add a comment in the log output telling the user from when the order is valid?

PR #95, opened 2020-03-23
Better use of promises for printing in function buildTransferApproveDepositFromList

josojo:


scripts/utils/trading_strategy_helpers.js
```diff
@@ -372,24 +372,20 @@ withdrawal of or to withdraw the desired funds
...
     for (const deposit of depositList) {
       assert(
         await isOnlySafeOwner(masterAddress, deposit.bracketAddress),
         "All depositors must be owned only by the master Safe"
       )
-      const depositToken = await ERC20.at(deposit.tokenAddress)
-      const tokenSymbol = await depositToken.symbol.call()
-      const tokenDecimals = await depositToken.decimals.call()
-      const unitAmount = fromErc20Units(deposit.amount, tokenDecimals)
+      const tokenInfo = await fetchTokenInfoAtAddresses([deposit.tokenAddress], debug)
```
fedgiac: The syntax is a bit different: `fetchTokenInfoAtAddresses` takes an array of addresses and returns a dictionary of promises, indexed by the address of the token. So this should be something like:
```
const tokenInfoPromises = fetchTokenInfoAtAddresses([deposit.tokenAddress], debug)
const tokenInfo = await tokenInfoPromises[deposit.tokenAddress]
```

fedgiac: I'm very surprised that the checks have passed. How is this possible?

josojo: Your syntax would be the right one for the function `fetchTokenInfoFromExchange`
But I am using `fetchTokenInfoAtAddresses`

fedgiac: Mmmh, they look consistent to me, the output `tokenPromises` is defined as a dictionary and every token is identified by its address.
Maybe the checks passed because here we use token info just for logging, which is not part of any test?

josojo: Ups you are right.
It just works, because the dictionary has only one object and so it is selected automatically, I guess. Let me change it to have it more explicit.

PR #99, opened 2020-03-23
Less error prone safe deployments

### LINKED ISSUE #38: Script fails unexpectedly after deploying some of the safes
When deploying many safes, sometimes the script fails unexpectedly after having deployed some safes. This is likely because the web3 provider doesn’t immediately see the newly created safe when requested and returns an error. Maybe we should add a delay after each wallet is created before accessing it on Infura?
 
This is the error:
 
```
Error: Cannot create instance of GnosisSafe; no code at address 0x3a58a6E39B62Fe2FF81e06B0777BB3a476e90f44
    at Object.checkCode (/path/dex-liquidity-provision/node_modules/truffle/build/webpack:/packages/contract/lib/utils/index.js:287:1)
    at Function.at (/path/dex-liquidity-provision/node_modules/truffle/build/webpack:/packages/contract/lib/contract/constructorMethods.js:74:1)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Truffle v5.1.12 (core: 5.1.12)
Node v10.18.1
```

josojo:
closes #38 

Makes the function deploySafe return only the address of the freshly deployed contract

### DIFF
--- a/scripts/utils/internals.js
+++ b/scripts/utils/internals.js
@@ -85,8 +85,6 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
       .setup(owners, threshold, ADDRESS_0, "0x", ADDRESS_0, ADDRESS_0, 0, ADDRESS_0)
       .encodeABI()
     const transaction = await proxyFactory.createProxy(gnosisSafeMasterCopy.address, initData)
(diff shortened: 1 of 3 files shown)


fedgiac:
Hopefully this fixes #38 for good.

PR #100, opened 2020-03-23
Safer retrieval of tokens

fedgiac:
Change the way tokens are retrieved in deposits and use same strategy to send funds back to master.

Motivation: with the current deposit script we are calling `fetchTokenInfoFromAddresses` once for each deposit. If the token has not been retrieved before, then since all calls are run in parallel we have that each call creates its own promise for the token info at the same time, thus contacting the node multiple times for the same token.


scripts/utils/trading_strategy_helpers.js
```diff
@@ -166,6 +167,25 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return tokenPromises
   }
 
+  /**
+   * Retrieves token information foll all tokens involved in the deposits.
+   * @param {Deposit[]} depositList List of {@link Deposit}
+   * @return {Promise<TokenObject>[]} list of detailed/relevant token information
+   */
+  const fetchTokenInfoForDeposits = function(deposits, debug = false) {
```
josojo: Why don't we call this function fetchTokenInfo and then use it for deposits and withdrawals?

PR #103, opened 2020-03-24
Deployment learnings

josojo:
These are the improvements suggestions from yesterday's deployment session

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -95,14 +95,14 @@ module.exports = async callback => {
       }
     }
     if (argv.fleetSize > 23) {
-      callback("Error: Choose a smaller fleetSize, otherwise your payload will be to big for Metamask")
+      callback("Error: Choose a smaller fleetSize, otherwise your payload will be to big for Infura nodes")
     }
 
     console.log(`2. Deploying ${argv.fleetSize} trading brackets`)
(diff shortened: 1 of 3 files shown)


scripts/complete_liquidity_provision.js
```diff
@@ -92,11 +94,18 @@ module.exports = async callback => {
         callback("Error: Bound checks did not pass")
       }
     }
+    if (argv.fleetSize > 23) {
+      callback("Error: Choose a smaller fleetSize, otherwise your payload will be to big for Metamask")
```
fedgiac: I tested it and this limit is enforced by Infura, if you use a different node the transaction goes through smoothly.

scripts/complete_liquidity_provision.js
```diff
@@ -92,11 +94,18 @@ module.exports = async callback => {
...
+    }
 
     console.log(`2. Deploying ${argv.fleetSize} trading brackets`)
     const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
     console.log("Following bracket-traders have been deployed", bracketAddresses.join())
 
+    // Sleeping for 5 seconds to make sure Metamask has processed all new deployed contracts so that
```
fedgiac: Also this is a workaround because of an issue with Infura, so I'd say that in the comment, Metamask is innocent.

fedgiac:
I don't like that we need this PR, but for now it's definitely necessary.

PR #105, opened 2020-03-24
Allow order placing for tokens with decimal places different than 18

josojo:
I want to refactor the code again, but the logic should already work.

@fedgiac I am very sorry that I needed to change your beautiful and always correct `toErc20Units` function to a function with rounding errors. :(
I don't have a good solution for that right now. I am still thinking about different ways to solve this issue. Suggestions are welcome :)

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -291,10 +291,16 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return buildBundledTransaction(transactions)
   }
 
-  const calculateBuyAndSellAmountsFromPrice = function(price, stableToken, targetToken) {
+  const calculateBuyAndSellAmountsFromPrice = function(price, stableToken, targetToken, maximalRoundingErrorInPercent = 0.1) {
     // decimalsForPrice: Decimals of reference price (OWL Price) + stableToken.decimals - targetToken.decimals
     const decimalsForPrice = 18 + stableToken.decimals - targetToken.decimals
-    price = price.toFixed(decimalsForPrice)
+    if (decimalsForPrice < 0 || decimalsForPrice > 30) {
+      throw "decimalsForPrice is not reasonable. Don't apply the bracket-strategy for this token pair!"
+    }
+    const roundedPrice = price.toFixed(decimalsForPrice)
+    if ((Math.abs(roundedPrice - price) / price) * 100 > maximalRoundingErrorInPercent) {
+      throw "Token decimals will introduce unreasonable rounding error!"
+    }
     const priceFormatted = toErc20Units(price, decimalsForPrice)
     let sellAmount
     let buyAmount
@@ -313,6 +319,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     }
     return [sellAmount, buyAmount]
   }
+
   const checkSufficiencyOfBalance = async function(token, owner, amount) {
     const depositor_balance = await token.balanceOf.call(owner)
     return depositor_balance.gte(amount)
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -294,7 +294,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   const calculateBuyAndSellAmountsFromPrice = function(price, stableToken, targetToken, maximalRoundingErrorInPercent = 0.1) {
     // decimalsForPrice: Decimals of reference price (OWL Price) + stableToken.decimals - targetToken.decimals
     const decimalsForPrice = 18 + stableToken.decimals - targetToken.decimals
-    if (decimalsForPrice < 0 || decimalsForPrice > 30) {
+    if (decimalsForPrice < 0 || decimalsForPrice > 36) {
       throw "decimalsForPrice is not reasonable. Don't apply the bracket-strategy for this token pair!"
     }
     const roundedPrice = price.toFixed(decimalsForPrice)
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -242,9 +242,6 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
 
     const targetToken = await tokenInfoPromises[targetTokenId]
     const stableToken = await tokenInfoPromises[stableTokenId]
-    // TODO - handle other cases later.
-    assert(stableToken.decimals === 18, "Target token must have 18 decimals")
-    assert(targetToken.decimals === 18, "Stable tokens must have 18 decimals")
 
     const stepSizeAsMultiplier = Math.pow(highestLimit / lowestLimit, 1 / bracketAddresses.length)
     log(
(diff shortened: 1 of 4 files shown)


test/dfusionMultiSafes.js
```diff
@@ -29,22 +29,30 @@ const { checkCorrectnessOfDeposits } = require("../scripts/utils/price-utils")(w
...
+    }
     const auctionElements = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders(bracketAddress))
     assert.equal(auctionElements.length, 2)
     const [buyOrder, sellOrder] = auctionElements
+    const decimalsOfSellToken = await (await ERC20.at(await exchange.tokenIdToAddressMap.call(buyOrder.sellToken))).decimals()
+    const decimalsOfBuyToken = await (await ERC20.at(await exchange.tokenIdToAddressMap.call(buyOrder.buyToken))).decimals()
+
     // Check buy order prices
     assert.isBelow(
       Math.abs(
```
josojo: ToDo: write a function measuring the difference between two BN. Then we can assert than the distance is smaller than 1 and we can set a high `multiplicator = new BN("1000000")` for all cases.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -258,10 +255,16 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
         const lowerLimit = lowestLimit * Math.pow(stepSizeAsMultiplier, bracketIndex)
         const upperLimit = lowerLimit * stepSizeAsMultiplier
 
-        const [upperSellAmount, upperBuyAmount] = calculateBuyAndSellAmountsFromPrice(upperLimit, targetToken)
+        const [upperSellAmount, upperBuyAmount] = calculateBuyAndSellAmountsFromPrice(
+          upperLimit,
+          18 + stableToken.decimals - targetToken.decimals
```
fedgiac: This number could be negative, while it's supposed to be the number of decimals of an ERC20 token. In this case the function fails.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -294,15 +297,15 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return buildBundledTransaction(transactions)
   }
 
-  const calculateBuyAndSellAmountsFromPrice = function(price, targetToken) {
+  const calculateBuyAndSellAmountsFromPrice = function(price, decimals) {
```
fedgiac: I'd consider having two input values instead of just one for clarity: `stableTokenDecimals` and `targetTokenDecimals`.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -294,15 +297,15 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
     // Sell targetToken for stableToken at price with unlimited orders
     // Example:
     // Sell 1 ETH at for 102 DAI (unlimited)
     // Sell x ETH for max256 DAI
     // x = max256 / 102
     // priceFormatted = 102000000000000000000
```
fedgiac: This comment is outdated and should be removed.

test/dfusionMultiSafes.js
```diff
@@ -29,22 +29,30 @@ const { checkCorrectnessOfDeposits } = require("../scripts/utils/price-utils")(w
...
-  if ((lowestLimit + highestLimit) / 2 < 10) {
-    multiplicator = new BN("10000000")
-  }
+
   // Correctness assertions
   for (const [index, bracketAddress] of bracketSafes.entries()) {
+    let multiplicator = new BN("100")
+    if (lowestLimit * Math.pow(stepSizeAsMultiplier, index) < 10) {
+      multiplicator = new BN("1000000")
+    }
```
fedgiac: I find these steps difficult to understand. I think this is for implementing floats with a fixed amount of decimals, and adapting the number of decimals depending on some parameter. Does it make sense to implement these floats for clarity?

test/dfusionMultiSafes.js
```diff
@@ -544,6 +554,59 @@ contract("GnosisSafe", function(accounts) {
...
       }
     })
+    it("Places bracket orders on behalf of a fleet of safes and checks prices for p<1, with different decimals than 18", async () => {
+      const masterSafe = await GnosisSafe.at(
+        await deploySafe(gnosisSafeMasterCopy, proxyFactory, [lw.accounts[0], lw.accounts[1]], 2)
+      )
+      const bracketSafes = await deployFleetOfSafes(masterSafe.address, 6)
+      testToken = await TestToken.new("TEST6", 6)
+      const testToken2 = await TestToken.new("TEST4", 4)
+      const targetToken = 0 // "TEST4"
```
fedgiac: `targetToken` is the target token id, right? But then if the id is zero we are building the orders against OWL instead of the new token. To add a new token to the exchange and get back the token and the id you could use the function `addCustomMintableTokenToExchange` from PR #107.

fedgiac:
I think it'd be surprising to have the function `toErc20Units` perform some operations (truncation) on the input while parsing it. It also hides part of the logic of the code, that is "round the amounts down to fit the token.
Also, this truncation is already done in the line:
```
price = price.toFixed(18)
``` 
we could simply replace 18 with `decimals`. (This still has the issue of negative `decimals`, however.)

On the other hand, I think it's not really an issue in this case to truncate the price in this case, since the input has to be truncated to become an order on Dfusion. I'd just move the truncation to its own function, instead of hiding it in `toErc20Units`.

test/dfusionMultiSafes.js
josojo: very good catch! Thanks for reviewing it so carefully.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -294,15 +291,17 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
-    // Sell x ETH for max256 DAI
-    // x = max256 / 102
-    // priceFormatted = 102000000000000000000
-    price = price.toFixed(18)
-    const priceFormatted = toErc20Units(price, targetToken.decimals)
+  const calculateBuyAndSellAmountsFromPrice = function(price, stableToken, targetToken, maximalRoundingErrorInPercent = 0.1) {
+    // decimalsForPrice: Decimals of reference price (OWL Price) + stableToken.decimals - targetToken.decimals
+    const decimalsForPrice = 18 + stableToken.decimals - targetToken.decimals
+    if (decimalsForPrice < 0 || decimalsForPrice > 36) {
+      throw "decimalsForPrice is not reasonable. Don't apply the bracket-strategy for this token pair!"
```
josojo: Test for these errors will come in a separate PR, so that we can continue testing for now.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -294,15 +302,17 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+    }
+    const roundedPrice = price.toFixed(decimalsForPrice)
+    if ((Math.abs(roundedPrice - price) / price) * 100 > maximalRoundingErrorInPercent) {
+      throw "Token decimals will introduce unreasonable rounding error!"
+    }
+    const priceFormatted = toErc20Units(roundedPrice, decimalsForPrice)
```
fedgiac: I don't like very much all these restrictions. The number 18 is still around in the code, and it's not very clear why. It doesn't seem to represent OWL, it kind of represent the amount of precision we use for the price, except that we change our precision depending on the difference of the number of decimals of the two tokens. So, if we trade USDC/DAI we have a precision of 6 decimals, while if we trade DAI/USDC we have a precision of 30.
I tried to come up with a suggestion that sets the precision to 18, but I couldn't make BN play together with floats nicely and consistently. I can't say I'm sure this code works as intended, it's likely fine for USDC DAI since the price is about 1, but if there is any token with two decimals I wouldn't be so confident that everything works as intended. Did I misunderstand something? Maybe we could also ask someone else to review the code to hear another opinion.

josojo: Good concern!

What do you think about my new implementation? 

It feels like we can control the precision pretty easily. I am not too worried about it

fedgiac: Much, much nicer! I love the new changes.

PR #107, opened 2020-03-24
Framework to test deposits

fedgiac:
Creates two functions, `testManualDeposits` and `testAutomaticDeposits`, that allow to add deposit test cases more easily.

### DIFF
--- a/test/dfusionMultiSafes.js
+++ b/test/dfusionMultiSafes.js
@@ -284,70 +284,25 @@ contract("GnosisSafe", function(accounts) {
       await Promise.all(testEntries.map(({ tradeInfo, expectedDistribution }) => testAutomaticDeposits(tradeInfo, expectedDistribution)))
     })
     it("transfers tokens from fund account through trader accounts and into exchange via automatic deposit logic, p<1 && p>1", async () => {
-      const masterSafe = await GnosisSafe.at(
-        await deploySafe(gnosisSafeMasterCopy, proxyFactory, [lw.accounts[0], lw.accounts[1]], 2)
-      )
-      const fleetSize = 4
-      const lowestLimit = 0.8
-      const highestLimit = 1.2
-      const currentPrice = 0.9
-      const bracketAddresses = await deployFleetOfSafes(masterSafe.address, fleetSize)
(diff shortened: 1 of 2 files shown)


test/dfusionMultiSafes.js
```diff
@@ -364,76 +257,102 @@ contract("GnosisSafe", function(accounts) {
...
+            stableTokenInfo: { decimals: 18, symbol: "DAI" },
+            targetTokenInfo: { decimals: 18, symbol: "WETH" },
+          },
+          expectedDistribution: {
+            bracketsWithStableTokenDeposit: 2,
+            bracketsWithTargetTokenDeposit: 2,
+          },
+        },
+      ]
+      await Promise.all(
```
josojo: This seems to be a little bit of overkill. A simple await for the `testAutmaticDeposits` should have been sufficient?

test/dfusionMultiSafes.js
```diff
@@ -364,76 +257,102 @@ contract("GnosisSafe", function(accounts) {
...
           targetToken,
-          depositAmountStableToken.div(new BN(1)),
-          depositAmountTargetToken.div(new BN(3))
+          bracketsWithStableTokenDeposit == 0 ? 0 : depositAmountStableToken.div(new BN(bracketsWithStableTokenDeposit)),
+          bracketsWithTargetTokenDeposit == 0 ? 0 : depositAmountTargetToken.div(new BN(bracketsWithTargetTokenDeposit))
         )
       }
+    }
+    it("transfers tokens from fund account through trader accounts and into exchange via automatic deposit logic, p > 1", async () => {
+      const testEntries = [
```
josojo: I have to admit it is very concise 👍

fedgiac:
I agree 100% with everything. I was quite conflicted about whether to use the printing functions or not. I used the printing functions because I was trying to make the test close to how we use the deposit functions in the scripts.
This has the added bonus that I can test  tokens with an arbitrary amount of decimals without changing any functions. Much of the logic of the deposits when handling the case with different decimals is somehow hidden in the scripts, and I hope it can be tested here.

PR #109, opened 2020-03-25
Compute withdrawal amounts using cached token info

fedgiac:


josojo:
I would have merge the original PR on the branch: manyDecimals into master, and then later one this one directly into master.
Then we have smaller commits on the master branch

PR #113, opened 2020-03-27
Create contract to deploy multiple proxy Safes in the same transaction

fedgiac:
This PR introduces and tests an Ethereum contract (FleetFactory) for deploying multiple proxy Safes with a single function call in the EVM.

Advantages are:
1. reduced gas cost (a bit more than 21k gas for each Safe deployed)
2. shortened bracket deployment time

### DIFF
--- a/contracts/FleetFactory.sol
+++ b/contracts/FleetFactory.sol
@@ -1,0 +1,43 @@
+pragma solidity >= 0.5.0 < 0.7.0;
+
+import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxyFactory.sol";
+import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxy.sol";
+import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+
+contract FleetFactory {
+  GnosisSafeProxyFactory public proxyFactory;
+
+  event FleetDeployed(address indexed owner, address[] fleet);
+
+  constructor(GnosisSafeProxyFactory _proxyFactory) public {
+    proxyFactory = _proxyFactory;
+  }
+
+  function deployFleet(address owner, uint256 size) public returns (address[] memory) {
+    IProxy ownerProxyReader = IProxy(owner);
+    return deployFleetFromTemplate(owner, size, ownerProxyReader.masterCopy());
+  }
+
+  function deployFleetFromTemplate(address owner, uint256 size, address template) public returns (address[] memory) {
+    address[] memory fleet = new address[](size);
+    address[] memory ownerList = new address[](1);
+    ownerList[0] = owner;
+    for (uint i = 0; i < size; i++) {
+      address payable proxy = address(proxyFactory.createProxy(template, ""));
+      fleet[i] = proxy;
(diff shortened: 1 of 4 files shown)


contracts/FleetFactory.sol
```diff
@@ -0,0 +1,44 @@
...
+contract FleetFactory {
+  GnosisSafeProxyFactory public proxyFactory;
+
+  event FleetDeployed(address indexed owner, address[] fleet);
+
+  constructor(GnosisSafeProxyFactory _proxyFactory) public {
+    proxyFactory = _proxyFactory;
+  }
+
+  function deployFleet(address owner, uint256 size) public returns (address[] memory) {
```
josojo: Why don't we hand over the template address? I think handing over the template address would be more general and the owner could also deploy other safes with other modules than just this current own one?

contracts/FleetFactory.sol
```diff
@@ -0,0 +1,44 @@
...
+  function deployFleetFromTemplate(address owner, uint256 size, address template) public returns (address[] memory) {
+    GnosisSafeProxyFactory _proxyFactory = proxyFactory;
+    address[] memory fleet = new address[](size);
+    address[] memory ownerList = new address[](1);
+    ownerList[0] = owner;
+    for (uint i = 0; i < size; i++) {
+      address payable proxy = address(_proxyFactory.createProxy(template, ""));
+      fleet[i] = proxy;
+      GnosisSafe safe = GnosisSafe(proxy);
+      safe.setup(
```
josojo: Here, you could add for readability comments explaining the setup...

contracts/FleetFactory.sol
```diff
@@ -0,0 +1,44 @@
+pragma solidity >= 0.5.0 < 0.7.0;
+
+import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxyFactory.sol";
+import "@gnosis.pm/safe-contracts/contracts/proxies/GnosisSafeProxy.sol";
+import "@gnosis.pm/safe-contracts/contracts/GnosisSafe.sol";
+
+contract FleetFactory {
+  GnosisSafeProxyFactory public proxyFactory;
+
+  event FleetDeployed(address indexed owner, address[] fleet);
```
josojo: Nice that you even thought about indexed events. I agree it is a good decision.

contracts/FleetFactory.sol
fedgiac: Everything was structured around our own use case: since we don't even have a parameter in our scripts to change the template I thought I could get rid of an extra parameter by retrieving the master copy from the owner. Note that you can specify the template using `deployFleetFromTemplate` however.
I absolutely agree that in the general case the template should be among the input values.

contracts/FleetFactory.sol
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(GnosisSafeProxyFactory _proxyFactory) public {
+    proxyFactory = _proxyFactory;
+  }
+
+  function deployFleet(address owner, uint256 size, address template) public returns (address[] memory) {
```
josojo: you could make this function external for minimal gas savings and cleanness

PR #115, opened 2020-03-31
[Easy] Delete simulations folder and update gitignore.

bh2smith:
The simulations folder contains code that is no longer relevant to this project. This PR removes this unused code and related files that no longer need to be ignored.


bh2smith:
> Nice to see you back!

Glad to be back! Sorry we had to delete our old code... but its still there in history

PR #116, opened 2020-03-31
Include multiple Safe deployment in the scripts

fedgiac:
Use the contract created in PR #113 to make bracket deployment in the scripts more efficient.

This PR has two prerequisites before merging:
1. PR #113 is finalized
2. The FleetFactory contract is deployed in at least mainnet and Rinkeby.


scripts/utils/trading_strategy_helpers.js
```diff
@@ -186,16 +186,13 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
     const gnosisSafeMasterCopy = await gnosisSafeMasterCopyPromise
 
-    // TODO - Batch all of this in a single transaction
-    const createdSafes = []
-    for (let i = 0; i < fleetSize; i++) {
-      const newSafeAddress = await deploySafe(gnosisSafeMasterCopy, proxyFactory, [masterAddress], 1)
-      log("New Safe Created", newSafeAddress)
-      createdSafes.push(newSafeAddress)
-    }
+    const transcript = await fleetFactory.deployFleet(masterAddress, fleetSize, gnosisSafeMasterCopy.address)
```
josojo: really cool that it is soo easy

josojo:
I was just quickly merging the conflicts so that I can see the diff correctly

test/test-utils.js
```diff
@@ -22,8 +23,35 @@ const addCustomMintableTokenToExchange = async function(exchange, symbol, decima
     token: token,
   }
 }
+const deploySafe = async function(gnosisSafeMasterCopy, proxyFactory, owners, threshold) {
```
josojo, line 26: Should we use here your new logic as well? 
using it in all our tests would give us more confidence.

Although, I am fine with using the old logic, as other tests are already in place for the new proxy deployments...

fedgiac: The new logic is already used everywhere, this function should be only used to create the master safe now.

PR #119, opened 2020-03-31
Verifies that orders are not profitable for others

josojo:
Adds the check that orders are not profitable for others

---
testplan:
```
npx truffle exec scripts/verify-correctness.js --masterSafe=$MASTER_SAFE --brackets 0xAB795A52D809e4A
a56da23b215Db7663d57CDE3A,0x45d8a3d10cd7Da86Dac21a51B305312Ebf80072d,0x2952bC69739c9534d09525ccCa33950F311CbE88,0x052253771E92a5Ff3c31163492b7e9101e0FC62f,0x90cBbcA16a0d1B802F575c179288B739d545b791,0xFcb1954Aff83F07dB5F3665349797BaD77A75c81,0xfAE2748F91525B7Dd7E9868c16f9Ad5a6EA5DE68,0x410f792da8a0a2e2f43f7b9317dA04E00F05D1AD,0xB498dB083e1E2c25A0d384f87a4D05eB7c05effC,0xB30D7400ab53cB81512629e26080127eFfb32c00,0xC5B24B4Dab763a44AFe57f66eE0074C2598E84ab,0x2cef9A8596aa983aD973bd5B8b5278F3AC83b789,0x1E800A7E4b103cA2C270e162E66d1B099eC8C1A7,0x65539d12c28946831D65Bb32eEc3A33CaBFB9249,0xD34AD431994192618E2E73867FA2B91eC2fAadf8,0x6bA77Ae3defE9101eE99a94aaCeacaDfE0B972E8,0xC77466A2449E519B5bffC3DFAcAAf1Fe1D765667,0x00Adb0E4c11BBe10A694678647A2076f62c12e11,0x069741B2fFC5b1Af74ABbC8f8573bA70e3808409,0x3247CFcA66Df47a641C52f8F8244c02Ea0825208 --network=mainnet
```
or other brackets of your choice.

### DIFF
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -104,19 +104,32 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   }
 
   // returns undefined if the price was not available
-  const getDexagPrice = async function(tokenBought, tokenSold) {
+  const getDexagPrice = async function(tokenBought, tokenSold, globalPriceStorage = null) {
+    if (globalPriceStorage !== null && tokenBought + "-" + tokenSold in globalPriceStorage) {
+      return globalPriceStorage[tokenBought + "-" + tokenSold]
+    }
+    if (globalPriceStorage !== null && tokenSold + "-" + tokenBought in globalPriceStorage) {
+      return 1.0 / globalPriceStorage[tokenSold + "-" + tokenBought]
+    }
     // dex.ag considers WETH to be the same as ETH and fails when using WETH as token
     tokenBought = tokenBought == "WETH" ? "ETH" : tokenBought
     tokenSold = tokenSold == "WETH" ? "ETH" : tokenSold
     // see https://docs.dex.ag/ for API documentation
     const url = "https://api-v2.dex.ag/price?from=" + tokenSold + "&to=" + tokenBought + "&fromAmount=1&dex=ag"
     let price
-    try {
-      const requestResult = await axios.get(url)
-      price = requestResult.data.price
-    } catch (error) {
-      console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
-      console.log(">", error.response.data.error)
+    // try to get price 3 times
+    for (let i = 0; i < 3; i++) {
+      try {
+        const requestResult = await axios.get(url)
+        price = requestResult.data.price
+        break
+      } catch (error) {
+        console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
+        console.log(">", error.response.data.error)
+      }
+    }
+    if (globalPriceStorage !== null) {
+      globalPriceStorage[tokenBought + "-" + tokenSold] = price
     }
     return price
   }
@@ -142,21 +155,19 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return true
   }
 
-  const checkNoProfitableOffer = async (order, exchange, currentMarketPrice) => {
+  const checkNoProfitableOffer = async (order, exchange, globalPriceStorage = null) => {
     const tokenInfo = fetchTokenInfoFromExchange(exchange, [order.buyToken, order.sellToken])
+    const currentMarketPrice = await getDexagPrice(
+      (await tokenInfo[order.buyToken]).symbol,
+      (await tokenInfo[order.sellToken]).symbol,
+      globalPriceStorage
+    )
 
-    if (currentMarketPrice === undefined) {
-      currentMarketPrice = await getDexagPrice(
-        (await tokenInfo[order.buyToken]).symbol,
-        (await tokenInfo[order.sellToken]).symbol
-      )
-    }
     // checks whether the order amount is negligible
-    if ((await valueInUSD(order, tokenInfo)) < 1) {
+    if ((await valueInUSD(order, tokenInfo, globalPriceStorage)) < 1) {
       return true
     }
 
-    // multiplicator is used to counter rounding errors
     const marketPrice = toErc20Units(currentMarketPrice, precisionDecimals)
     const orderPrice = toErc20Units(order.priceNumerator, precisionDecimals)
       .mul(new BN(10).pow(new BN((await tokenInfo[order.sellToken]).decimals)))
@@ -166,10 +177,9 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return marketPrice.lt(orderPrice)
   }
 
-  const valueInUSD = async (order, tokenInfo, currentMarketPrice) => {
-    if (currentMarketPrice === undefined) {
-      currentMarketPrice = await getDexagPrice("USDC", (await tokenInfo[order.sellToken]).symbol)
-    }
+  const valueInUSD = async (order, tokenInfo, globalPriceStorage = null) => {
+    const currentMarketPrice = await getDexagPrice("USDC", (await tokenInfo[order.sellToken]).symbol, globalPriceStorage)
+
     return fromErc20Units(
       toErc20Units(currentMarketPrice, precisionDecimals)
         .mul(order.sellTokenBalance)
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -1,7 +1,7 @@
 module.exports = function(web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const BN = require("bn.js")
-  const precisionDecimals = 50
+  const precisionDecimals = 20
 
   const exchangeUtils = require("@gnosis.pm/dex-contracts")
   const { toErc20Units, fromErc20Units } = require("./printing_tools")
@@ -166,7 +166,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     )
 
     // checks whether the order amount is negligible
-    if ((await valueInUSD(order, tokenInfo, globalPriceStorage)) < 1) {
+    if (Math.floor(await valueInUSD(order, tokenInfo, globalPriceStorage)) < 1) {
       return true
     }
 
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -124,8 +124,10 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
         price = requestResult.data.price
         break
       } catch (error) {
-        console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
-        console.log(">", error.response.data.error)
+        if (i == 2) {
+          console.log("Warning: unable to retrieve price information on dex.ag. The server returns:")
+          console.log(">", error.response.data.error)
+        }
       }
     }
     if (globalPriceStorage !== null) {
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return true
   }
 
+  const checkNoProfitableOffer = async (order, exchange, currentMarketPrice) => {
+    const tokenInfo = fetchTokenInfoFromExchange(exchange, [order.buyToken, order.sellToken])
+
+    if (currentMarketPrice === undefined) {
+      currentMarketPrice = await getDexagPrice(
+        (await tokenInfo[order.buyToken]).symbol,
+        (await tokenInfo[order.sellToken]).symbol
+      )
+    }
+    // checks whether the order amount is negligible
+    if ((await valueInUSD(order, tokenInfo)) < 10) {
+      return true
+    }
+
+    // multiplicator is used to counter rounding errors
+    const multiplicator = 1000000000
+    const marketPrice = new BN(currentMarketPrice * multiplicator)
+    const orderPrice = order.priceNumerator
+      .mul(new BN(10).pow(new BN((await tokenInfo[order.sellToken]).decimals)))
+      .div(new BN(10).pow(new BN((await tokenInfo[order.buyToken]).decimals)))
+      .mul(new BN(multiplicator))
+      .div(order.priceDenominator)
+
+    return marketPrice.lt(orderPrice)
+  }
+
+  const valueInUSD = async (order, tokenInfo, currentMarketPrice) => {
+    if (currentMarketPrice === undefined) {
+      currentMarketPrice = await getDexagPrice("USDC", (await tokenInfo[order.sellToken]).symbol)
+    }
+    const multiplicator = 100000.0
+    return (
+      (currentMarketPrice *
+        order.sellTokenBalance
+          .mul(new BN(multiplicator))
+          .div(new BN(10).pow(new BN((await tokenInfo[order.sellToken]).decimals)))
+          .toNumber()) /
+      multiplicator
+    )
+  }
+
   return {
     isPriceReasonable,
     areBoundsReasonable,
     checkCorrectnessOfDeposits,
+    getDexagPrice,
+    checkNoProfitableOffer,
   }
 }
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -1,7 +1,10 @@
 module.exports = function(web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const BN = require("bn.js")
+  const precisionDecimals = 50
+
   const exchangeUtils = require("@gnosis.pm/dex-contracts")
+  const { toErc20Units, fromErc20Units } = require("./printing_tools")
 
   const { fetchTokenInfoFromExchange } = require("./trading_strategy_helpers")(web3, artifacts)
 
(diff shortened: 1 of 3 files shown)


scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+      )
+    }
+    // checks whether the order amount is negligible
+    if ((await valueInUSD(order, tokenInfo)) < 10) {
+      return true
+    }
+
+    // multiplicator is used to counter rounding errors
+    const multiplicator = 1000000000
+    const marketPrice = new BN(currentMarketPrice * multiplicator)
```
fedgiac, line 158: If `currentMarketPrice * multiplicator` is not an integer, then this line fails. This might be an issue for prices like `0.33333333333333333`. Also, we implicitly assume currentMarketPrice > 10^-9, are we ok with this?

scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+    const orderPrice = order.priceNumerator
+      .mul(new BN(10).pow(new BN((await tokenInfo[order.sellToken]).decimals)))
+      .div(new BN(10).pow(new BN((await tokenInfo[order.buyToken]).decimals)))
+      .mul(new BN(multiplicator))
```
fedgiac, line 162: This shouldn't work for USDC/DAI, since 10^9 * 10*6 / 10**18 < 1, and everything is rounded to zero.

scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+      .div(order.priceDenominator)
+
+    return marketPrice.lt(orderPrice)
```
fedgiac, line 165: Since computing prices is usually a source of mistakes, I wonder if it makes sense to create a standalone function converting prices to some version of "BN float" like you are doing here. Then this function could be tested separately and then we could rely on that everywhere.

scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+    const tokenInfo = fetchTokenInfoFromExchange(exchange, [order.buyToken, order.sellToken])
+
+    if (currentMarketPrice === undefined) {
+      currentMarketPrice = await getDexagPrice(
+        (await tokenInfo[order.buyToken]).symbol,
+        (await tokenInfo[order.sellToken]).symbol
+      )
+    }
+    // checks whether the order amount is negligible
+    if ((await valueInUSD(order, tokenInfo)) < 10) {
```
fedgiac, line 152: Just a nitpick, but I'd lower the negligibility threshold to 1 USD.

scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+  const valueInUSD = async (order, tokenInfo, currentMarketPrice) => {
+    if (currentMarketPrice === undefined) {
+      currentMarketPrice = await getDexagPrice("USDC", (await tokenInfo[order.sellToken]).symbol)
+    }
+    const multiplicator = 100000.0
+    return (
+      (currentMarketPrice *
+        order.sellTokenBalance
+          .mul(new BN(multiplicator))
+          .div(new BN(10).pow(new BN((await tokenInfo[order.sellToken]).decimals)))
```
fedgiac, line 177: This assumes that the amount of tokens is more than 10^-5, otherwise returns zero. This could be problematic with tokens of high values. So for example an amount of bitcoins worth less that 0.1USD is zero.

scripts/utils/price-utils.js
```diff
@@ -139,9 +139,52 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
   }
 
+  const checkNoProfitableOffer = async (order, exchange, currentMarketPrice) => {
+    const tokenInfo = fetchTokenInfoFromExchange(exchange, [order.buyToken, order.sellToken])
+
+    if (currentMarketPrice === undefined) {
+      currentMarketPrice = await getDexagPrice(
+        (await tokenInfo[order.buyToken]).symbol,
+        (await tokenInfo[order.sellToken]).symbol
+      )
```
fedgiac, line 149: I've noticed that very often when I pushed my commits to Github Travis told me that the tests failed just because the one call we make to dex.ag failed to retrieve the price for whatever reason, and I had to rerun Travis to make the tests pass. And this is just a single call to their servers.
As I read this PR, you call dex.ag for every order currently active for a bracket. Considering how many brackets we deployed so far (>100?) I'd say it's basically certain that one of these calls will fail during the execution of the script. Or we'd at least run into some rate limiting. I'd say that at least the price should be recovered only once for each pair, similar to what we do for ERC20 token info. This would at least reduce the number of calls to ~10, which hopefully doesn't make us rerun the verification scripts more than a few times to see if it works. Ideally we would also catch failures and retry after a while for a certain amount of times (let's say 5?).

test/priceUtils.js
```diff
@@ -26,5 +33,105 @@ contract("PriceOracle", function(accounts) {
...
+          sellToken: 1,
+          priceNumerator: new BN("132").mul(new BN(10).pow(new BN(18))),
+          priceDenominator: new BN("1").mul(new BN(10).pow(new BN(18))),
+        },
+      ]
+
+      assert.equal(await checkNoProfitableOffer(orders[0], exchange, 1.0 / 120), true)
+      assert.equal(await checkNoProfitableOffer(orders[1], exchange, 120), true)
+    })
+    it("checks that bracket traders does not sell unprofitable for tokens with the different decimals", async () => {
```
fedgiac, line 69: Some typos, my proposal:
```
detects unprofitable order for tokens with different decimals
```

test/priceUtils.js
```diff
@@ -26,5 +33,105 @@ contract("PriceOracle", function(accounts) {
...
+          sellTokenBalance: new BN("0"),
+          buyToken: 1,
+          sellToken: USDCtokenId,
+          priceNumerator: new BN("101").mul(new BN(10).pow(new BN(18))),
+          priceDenominator: new BN("100").mul(new BN(10).pow(new BN(6))),
+        },
+      ]
+      assert.equal(await checkNoProfitableOffer(orders[0], exchange, 1), false)
+      assert.equal(await checkNoProfitableOffer(orders[1], exchange, 1), true)
+    })
```
fedgiac, line 135: Since the code is almost the same as in the test above (it took me a while to spot the difference!) maybe it would be clearer if we had the same function called on each test instead of repeating the code?

scripts/verify-correctness.js
```diff
@@ -74,6 +75,15 @@ module.exports = async callback => {
         // If the last equation holds, the inverse trade must be profitable as well
       })
     )
+
+    // 5. verify that no bracket-trader offers profitable orders
+    await Promise.all(
+      bracketTraderAddresses.map(async bracketTrader => {
+        const relevantOrders = auctionElementsDecoded.filter(order => order.user.toLowerCase() == bracketTrader)
+        relevantOrders.forEach(async order => assert.equal(await checkNoProfitableOffer(order, BatchExchange), true))
+      })
```
fedgiac, line 84: Nice, promises!

scripts/utils/price-utils.js
josojo: I just tested it again: `const t = new BN(1.34)` gives the same as new BN(1).

To be honest, I don't have a good solution for it right now. because if `currentMarketPrice * multiplicator` gets too big, we get a scientific representation, as eg. 4.546e30 and then this throws as well.

scripts/utils/price-utils.js
josojo: Same as above. Still thinking about a good solution..

scripts/utils/price-utils.js
josojo: Actually, I found a good way :) Thank pushing my focus to the rounding issues again

scripts/utils/price-utils.js
josojo: I think one can just use your existing function: `toErc20Units`. Please double check

test/priceUtils.js
josojo: If we are comparing all 3 added tests, we have: token creations( individual), token additions to the exchange(every time the same code) and order definition(individual) and final check (individual).

Since the token additions are the only code that really is the same, I could factor this into a function. But since it is only 4 lines, I would say that it is fine.
I think the readability improves from this easy 'log book' style.

scripts/utils/price-utils.js
fedgiac: Mmh, I'm not sure how you propose to use `toErc20Units` here. My proposal would be to import the `Fraction` class that Felix created and extend it with a `fromNumber` method that creates a fraction approximation of a float (and for practicality also extend it with `.mulBN`, `.divBN`). Then we can use the code you wrote above, even without the multiplicator.

scripts/utils/price-utils.js
```diff
@@ -101,19 +104,34 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   }
 
   // returns undefined if the price was not available
-  const getDexagPrice = async function(tokenBought, tokenSold) {
+  const getDexagPrice = async function(tokenBought, tokenSold, globalPriceStorage = null) {
+    if (globalPriceStorage !== null && tokenBought + "-" + tokenSold in globalPriceStorage) {
+      return globalPriceStorage[tokenBought + "-" + tokenSold]
+    }
+    if (globalPriceStorage !== null && tokenSold + "-" + tokenBought in globalPriceStorage) {
+      return 1.0 / globalPriceStorage[tokenSold + "-" + tokenBought]
```
fedgiac, line 112: Here we assume that the token symbol does not contain `-`.

scripts/utils/price-utils.js
```diff
@@ -139,9 +157,44 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+  const checkNoProfitableOffer = async (order, exchange, globalPriceStorage = null) => {
+    const tokenInfo = fetchTokenInfoFromExchange(exchange, [order.buyToken, order.sellToken])
+    const currentMarketPrice = await getDexagPrice(
+      (await tokenInfo[order.buyToken]).symbol,
+      (await tokenInfo[order.sellToken]).symbol,
+      globalPriceStorage
+    )
+
+    // checks whether the order amount is negligible
+    if ((await valueInUSD(order, tokenInfo, globalPriceStorage)) < 1) {
```
fedgiac, line 169: Here we are comparing a string with an integer. I'm not sure what the result is in general, but I'd cast the string to integer to be safe.

scripts/verify-correctness.js
```diff
@@ -74,6 +87,17 @@ module.exports = async callback => {
...
+    await Promise.all(
+      bracketTraderAddresses.map(async bracketTrader => {
+        const relevantOrders = auctionElementsDecoded.filter(order => order.user.toLowerCase() == bracketTrader)
+        relevantOrders.forEach(async order =>
+          assert.equal(await checkNoProfitableOffer(order, BatchExchange, globalPriceStorage), true)
+        )
+      })
+    )
+
     callback()
```
fedgiac, line 101: I think this block might not be working as intended: lines 94 to 97 define the mapping function, but the block has no return value. So the result of the mapping is a vector of `undefined` which is instantly resolved to itself by `Promise.all`. Then the function `callback()` is called in line 101, before the checks were completed. At this point I don't know what happens to the code that is being executed in the `forEach`.

fedgiac: Is there a way to make js fail if `undefined` is used at any point in a computation in the code? This would have helped in many other occasions.

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
       assert(await isPriceReasonable(exchange, targetTokenId, stableTokenId, price, acceptedPriceDeviationInPercentage))
     })
+    it("checks that bracket traders does not sell unprofitable for tokens with the same decimals", async () => {
+      const ERC20 = artifacts.require("DetailedMintableToken")
+      const token1 = await ERC20.new("WETH", 18)
+      const token2 = await ERC20.new("DAI", 18)
+      await prepareTokenRegistration(accounts[0], exchange)
+      await exchange.addToken(token1.address, { from: accounts[0] })
+      await prepareTokenRegistration(accounts[0], exchange)
+      await exchange.addToken(token2.address, { from: accounts[0] })
```
fedgiac, line 43: You could use `addCustomMintableTokenToExchange` from the test utils.

test/priceUtils.js
fedgiac: Typos are still there, this is because the same sentence was also used in the test immediately below, which was however changed. So maybe the two tests should have a different name?

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
+      const token1 = await ERC20.new("DAI", 18)
+      const token2 = await ERC20.new("USDC", 6)
+
+      await prepareTokenRegistration(accounts[0], exchange)
+      await exchange.addToken(token1.address, { from: accounts[0] })
+      const DAItokenId = (await exchange.numTokens.call()).toNumber() - 1
+
+      await prepareTokenRegistration(accounts[0], exchange)
+      await exchange.addToken(token2.address, { from: accounts[0] })
+      const USDCtokenId = (await exchange.numTokens.call()).toNumber() - 1
```
fedgiac, line 85: `addCustomMintableTokenToExchange` could be used also here.

fedgiac: Or maybe not? Is `tokenAddressToIdMap` equivalent to `exchange.numTokens - 1` here?

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
+          priceDenominator: new BN("1").mul(new BN(10).pow(new BN(18))),
+        },
+      ]
+
+      const globalPriceStorage = {}
+      globalPriceStorage["DAI-USDC"] = 1.0
+      globalPriceStorage["WETH-DAI"] = 1 / 120.0
+      globalPriceStorage["WETH-USDC"] = 1 / 120.0
+
+      assert.equal(await checkNoProfitableOffer(orders[0], exchange, globalPriceStorage), true)
```
fedgiac, line 71: Here the check succeeds because the amount is negligible, right? I'd add this to a comment somewhere.

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
+          priceNumerator: new BN("101").mul(new BN(10).pow(new BN(6))),
+          priceDenominator: new BN("100").mul(new BN(10).pow(new BN(18))),
+        },
+      ]
+
+      const globalPriceStorage = {}
+      globalPriceStorage["USDC-USDC"] = 1.0
+      globalPriceStorage["DAI-USDC"] = 1.0
+      assert.equal(await checkNoProfitableOffer(orders[0], exchange, globalPriceStorage), true)
+      assert.equal(await checkNoProfitableOffer(orders[1], exchange, globalPriceStorage), true)
```
fedgiac, line 112: Same here, the check succeeds because the amount is negligible. These checks could go to a different test.

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
+          priceDenominator: new BN("100").mul(new BN(10).pow(new BN(6))),
+        },
+      ]
+
+      const globalPriceStorage = {}
+      globalPriceStorage["USDC-USDC"] = 1.0
+      globalPriceStorage["DAI-USDC"] = 1.0
+
+      assert.equal(await checkNoProfitableOffer(orders[0], exchange, globalPriceStorage), false)
+      assert.equal(await checkNoProfitableOffer(orders[1], exchange, globalPriceStorage), true)
```
fedgiac, line 152: Again, negligible amount.

test/priceUtils.js
```diff
@@ -26,5 +33,123 @@ contract("PriceOracle", function(accounts) {
...
+      const DAItokenId = (await exchange.numTokens.call()).toNumber() - 1
+      await prepareTokenRegistration(accounts[0], exchange)
+      await exchange.addToken(token2.address, { from: accounts[0] })
+      const USDCtokenId = (await exchange.numTokens.call()).toNumber() - 1
+
+      const orders = [
+        {
+          // order is profitable for others
+          user: "0x4c7281e2bd549a0aea492b28ef60e3d81fed36e6",
+          sellTokenBalance: new BN("10eed9efc"),
```
fedgiac, line 130: This should definitely fail, since `10eed9efc` is not a number. The standard way to get a value from a hex string is `new BN("10eed9efc", 16)`. This issue was fixed in PR #111, so it's probably better to merge master into this PR to get the last updates.

fedgiac: Just out of curiosity, where is this number from?

josojo: I copied real orders

scripts/verify-correctness.js
```diff
@@ -35,6 +39,15 @@ module.exports = async callback => {
...
     const bracketTraderAddresses = argv.brackets.map(address => address.toLowerCase())
 
+    // fetch all token infos(decimals, symbols etc) and prices upfront for the following verification
+    const globalPriceStorage = {}
+    const relevantOrders = auctionElementsDecoded.filter(order => bracketTraderAddresses.includes(order.user.toLowerCase()))
+    relevantOrders.forEach(async order => {
+      const tokenInfo = await fetchTokenInfoFromExchange([order.sellToken, order.buyToken])
+      await getDexagPrice(tokenInfo[order.sellToken].symbol, tokenInfo[order.sellToken].symbol, globalPriceStorage)
+      await getDexagPrice(tokenInfo[order.sellToken].symbol, "USDC", globalPriceStorage)
+    })
```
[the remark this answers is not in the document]
josojo: Is it? Because we are storing every result in globalPriceStorage and globalTokenPromisesFromId, I thought that it would be efficient. Where do we lose time?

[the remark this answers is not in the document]
fedgiac: Whoops, you are right, I thought that `.forEach` executed each function one after the other but it executes everything in parallel. So there isn't much to gain by changing this code, sorry about that.

scripts/verify-correctness.js
josojo: Unfortunately, I am not aware of such an option. But yeah, this would have helped to spot many things. If we would have used typescript, this also would not have happened.

PR #121, opened 2020-04-01
FleetFactory migration

fedgiac:
Setup tools needed for a clean migration on mainnet, including contract verification on Etherscan.

Migrations are run as follows:
```
export NETWORK_NAME=<insert network name, e.g. rinkeby or mainnet>
export GAS_PRICE_GWEI=<insert gas price in Gwei, e.g. 20>
 export PK=<insert Ethereum private key here>
 export MY_ETHERSCAN_API_KEY=<insert Etherscan API key here>

rm -r build
yarn compile
yarn networks-inject
yarn migrate --network $NETWORK_NAME
yarn networks-extract
yarn verify FleetFactory --network $NETWORK_NAME
```

Contracts deployed on Rinkeby:
Migrations: `0xca39A48b3b0C8284bfF8f2299539e2BbaBE714de`
FleetFactory: `0x1b94BF378B910e5be70b59e774867a1c919821C7`

Contracts deployed on mainnet:
Migrations: `0x094419790e6bfD8071902C7DCa7D59F4c489bF0c`
FleetFactory: `0xc628FB35ca39705CCD0cBDD5080F1b3d267AdC31`

### DIFF
--- a/migrations/1_initial_migration.js
+++ b/migrations/1_initial_migration.js
@@ -1,5 +1,5 @@
 const Migrations = artifacts.require("./Migrations.sol")
 
 module.exports = function(deployer) {
-  deployer.deploy(Migrations)
+  deployer.deploy(Migrations, {gas: 300000})
 }
--- a/migrations/1_initial_migration.js
+++ b/migrations/1_initial_migration.js
@@ -1,5 +1,5 @@
 const Migrations = artifacts.require("./Migrations.sol")
 
 module.exports = function(deployer) {
-  deployer.deploy(Migrations, {gas: 300000})
+  deployer.deploy(Migrations, { gas: 300000 })
 }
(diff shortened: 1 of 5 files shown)


migrations/1_initial_migration.js
```diff
@@ -1,5 +1,5 @@
 const Migrations = artifacts.require("./Migrations.sol")
 
 module.exports = function(deployer) {
-  deployer.deploy(Migrations)
+  deployer.deploy(Migrations, { gas: 300000 })
```
josojo, line 4: why is this needed, any clue?

fedgiac: The reason why I included this is that I hope we can get away with cheaper gas price for deployment. The default gas limit is currently very high, about 6000000, so a miner might ignore our transaction until there is a block that has 6M gas available at a price that we give, in which case we need a higher gas price to deploy at the same speed as with a lower gas limit.
I have to say that I'm not sure it really works like this, so maybe this is overthinking the problem. Considering we are not going to deploy this contract again, we could just remove this hardcoded gas limit.

PR #124, opened 2020-04-02
Better argument hadling

fedgiac:
Reworks how command-line arguments are handled.

Most important change is that the scripts fail if one of the input arguments is not defined. This helps avoiding misspelling of arguments which would lead to some probably unwanted default behavior. Example: running the withdraw script with `--allToken` now fails (the right name is `--allTokens`).

### DIFF
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -26,7 +26,7 @@ const argv = require("./utils/default_yargs")
   })
   .option("brackets", {
     type: "string",
-    describe: "Trader account addresses to place orders on behalf of.",
+    describe: "Trader account addresses to place orders on behalf of",
     demandOption: true,
     coerce: str => {
       return str.split(",")
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -49,8 +49,7 @@ const argv = require("./utils/default_yargs")
     type: "int",
     describe: "Maximum auction batch for which these orders are valid",
     default: 2 ** 32 - 1,
-  })
-  .argv
+  }).argv
 
 module.exports = async callback => {
   try {
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -3,7 +3,7 @@ const { isPriceReasonable, areBoundsReasonable } = require("./utils/price-utils.
 const { proceedAnyways } = require("./utils/user-interface-helpers")
 const { signAndSend, promptUser } = require("./utils/sign_and_send")(web3, artifacts)
 
-const argv = require("yargs")
+const argv = require("./utils/default_yargs")
   .option("targetToken", {
     type: "int",
     describe: "Token whose target price is to be specified (i.e. ETH)",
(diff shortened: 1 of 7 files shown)


scripts/utils/default_yargs.js
```diff
@@ -0,0 +1,12 @@
...
+  .version(false)
+  .strict()
+  .help("help")
+  .epilog(
+    "Make sure that you have an RPC connection to the network in consideration. For network configurations, please see truffle-config.js"
+  )
+  .option("network", {
+    type: "string",
+    describe: "network where the script is executed",
+    choices: ["rinkeby", "mainnet"],
```
josojo, line 11: While we usually don't test against development, we could still offer it here

scripts/utils/default_yargs.js
```diff
@@ -0,0 +1,12 @@
+module.exports = require("yargs")
+  .version(false)
+  .strict()
```
josojo, line 3: oh yeah, strict is good to have!

scripts/utils/default_yargs.js
fedgiac: The issue is that all the scripts rely on the multisig web interface, they will surely not work on development, so I didn't want to offer it explicitly. (It would be great to have a local version of the multisig, however!)
We can still test the scripts on development however: the `network` argument is not required, so by omitting it it automatically runs on development. This is also how I tested this PR.

fedgiac:
Another reason for the change is that `demand` is currently deprecated.

scripts/bracket_orders.js
```diff
@@ -43,12 +49,7 @@ const argv = require("yargs")
     type: "int",
     describe: "Maximum auction batch for which these orders are valid",
     default: 2 ** 32 - 1,
-  })
-  .demand(["targetToken", "stableToken", "currentPrice", "masterSafe", "brackets"])
```
bh2smith, line 47: Wasn't this already doing what this PR claims to do?

scripts/utils/default_yargs.js
bh2smith: What does strict do/mean?

scripts/utils/default_yargs.js
```diff
@@ -0,0 +1,12 @@
...
+  .help("help")
+  .epilog(
+    "Make sure that you have an RPC connection to the network in consideration. For network configurations, please see truffle-config.js"
```
bh2smith, line 6: Cool!

scripts/utils/default_yargs.js
bh2smith: I am also running into this issue with the synthetix integration. Their npm package only exposes deployed versions of their system from which we can query their onchain price feed and token addresses (etc..). However, this will be untestable on the development network without depending on their entire contracts repo and deploying their entire system on every test.

Of course, this comment is not entirely relevant to this PR, I just needed to mentioned it somewhere and have not yet come up with a scheme for testing their stuff locally.

scripts/utils/default_yargs.js
fedgiac: Strict means that the script fails if any of the arguments used during its invocation is not explicitly defined by us in yargs.

scripts/bracket_orders.js
fedgiac: Nope, this line makes the script fail if one of the required arguments is not there. The main goal of the PR is to fail if any unknown argument is used (so, .`strict()`).
Since I was at it I also moved the `demand` flag inside the definition of the required options for readability.

bh2smith: Awesome!

PR #125, opened 2020-04-02
Verify script - allowances

fedgiac:


scripts/verify-correctness.js
```diff
@@ -74,6 +115,15 @@ module.exports = async callback => {
         // If the last equation holds, the inverse trade must be profitable as well
       })
     )
+
+    // TODO: extract addresses of traded tokens from previous checks
+    const tradedTokenAddresses = []
```
fedgiac, line 120: To @josojo: I think it's easier to modify your code to yield the list of traded token than to implement my own fetching here, also because it would probably mean that I have to reread all orders again just to find the token ids of the orders.

scripts/verify-correctness.js
```diff
@@ -74,6 +115,15 @@ module.exports = async callback => {
...
+    const tokenInfo = fetchTokenInfoAtAddresses(tradedTokenAddresses)
+    // 5. verify that no allowances are currently set
+    await assertNoAllowances(argv.masterSafe[0], tokenInfo)
```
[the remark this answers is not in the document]
fedgiac: How do we want to handle this? I could create a list of "exception addresses" that will be ignored by the check.

fedgiac:
PR is ready, I'm only waiting to merge upstream changes to remove `tradedTokenAddresses = []`.

test/verification_script.js
```diff
@@ -0,0 +1,104 @@
+const TestToken = artifacts.require("DetailedMintableToken")
+const { getAllowances, assertNoAllowances } = require("../scripts/utils/trading_strategy_helpers")(web3, artifacts)
+
+contract("verification checks", async accounts => {
+  describe("allowances", async () => {
+    let tokenInfo
+
+    const createToken = async function(symbol, decimals) {
```
josojo, line 8: You could lift this function to some utils file in the testing folder. Then it is easier to reuse

PR #127, opened 2020-04-06
[Synthetix] Dependency and example usage.

bh2smith:
This PR introduces a dependency on `synthetix-js` which exposes access to the entire synthetic platform. For our purposes, we really only require access to the token addresses for `sUSD` and `sETH` (which are dynamic in the sense that they can be redeployed for the purpose of, say, upgrading). Furthermore, this dependency is required for access to their onchain price oracle (provided by Chainlink. That is, 

we will essentially only require access the following parts of their platform:
- `snxjs.sETH.currencyKey()`
- `snxjs.Synthetix.synths(sETHKey)`
- `snxjs.ExchangeRates.rateForCurrency(currencyKey)`

via a JS object `snxjs = new SynthetixJs({ networkId: networkMap[argv.network] })`

**Test Plan**
```
truffle exec scripts/synthetix/facilicate_trade.js --network rinkeby
```

Which will place a spread order of 0.2% between `sUSD` and `sETH` at the rate quoted by their onchain oracle. NOte that this order is valid only for the current auction batch.

This is simply the first of a few following PRs that should introduces the key concepts of this integration.


scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+module.exports = async callback => {
+  try {
+    const snxjs = new SynthetixJs({ networkId: networkMap[argv.network] })
+
+    const exchange = await getExchange(web3)
+    const account = (await web3.eth.getAccounts())[0]
+
+    // Both of these hardcoded tokens have 18 decimal places.
+    // We "trust" that this will always be the case
+    // Although it seems that synthetix has the authority to upgade their token
```
fleupold, line 35: nit: upg**r**ade

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+    // which could mean deploying a new one with a different number of decimal places.
+    const sETHKey = await snxjs.sETH.currencyKey()
+    const sUSDKey = await snxjs.sUSD.currencyKey()
+
+    const sETHAddress = await snxjs.Synthetix.synths(sETHKey)
+    const sUSDAddress = await snxjs.Synthetix.synths(sUSDKey)
```
fleupold, line 41: I'm wondering if we shouldn't use the addresses of the tokens that are listed on dƒusion as if they upgraded the addresses we would have to first list the new tokens anyways in order to place orders.

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+    await sendTxAndGetReturnValue(
+      exchange.placeValidFromOrders,
+      buyTokens,
+      sellTokens,
+      validFroms,
+      validTos,
+      buyAmounts,
+      sellAmounts,
+      {
+        from: account,
```
fleupold, line 89: nit: isn't account[0] used by default anywaus?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+// truffle uses network by name and SynthetixJS uses network by ID
+const networkMap = {
+  "mainnet": 1,
+  "rinkeby": 4,
+}
+
+const argv = require("yargs")
+  .option("spread", {
+    type: "float",
+    describe: "Percentage increase required for trade (fees not accounted)",
```
fleupold, line 21: What exactly do you mean by fees not accounted? Don't our limit orders include the Gnosis protocol trading fee?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
+const { getExchange } = require("../utils/trading_strategy_helpers")(web3, artifacts)
+const { SynthetixJs } = require("synthetix-js")
+const { calculateBuyAndSellAmountsFromPrice, fetchTokenInfoFromExchange } = require("../utils/trading_strategy_helpers")(web3, artifacts)
+
+// TODO - put this somewhere more generic. Copied from dex-contracts
+const sendTxAndGetReturnValue = async function (method, ...args) {
```
fleupold, line 6: For now it doesn't seem as if you are using the return value, are you?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+    const formatedRate = snxjs.utils.formatEther(exchangeRate)
+    console.log("sETH Price", snxjs.utils.formatEther(exchangeRate))
+
+    const batch_index = (await exchange.getCurrentBatchId.call()).toNumber()
+
+    const buyTokens = [etherTokenId, referenceTokenId]
+    const sellTokens = [referenceTokenId, etherTokenId]
+
+    const [upperSellAmount, upperBuyAmount] = calculateBuyAndSellAmountsFromPrice(
+      formatedRate * (1 + argv.spread / 100),
```
fleupold, line 63: Maybe we should use their contracts to determine the fee? https://github.com/Synthetixio/synthetix/blob/master/contracts/Exchanger.sol#L75

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+    const exchangeRate = await snxjs.ExchangeRates.rateForCurrency(sETHKey)
+    const formatedRate = snxjs.utils.formatEther(exchangeRate)
+    console.log("sETH Price", snxjs.utils.formatEther(exchangeRate))
+
+    const batch_index = (await exchange.getCurrentBatchId.call()).toNumber()
+
+    const buyTokens = [etherTokenId, referenceTokenId]
+    const sellTokens = [referenceTokenId, etherTokenId]
+
+    const [upperSellAmount, upperBuyAmount] = calculateBuyAndSellAmountsFromPrice(
```
fleupold, line 62: Are we using an unlimited amount for now?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+  .option("spread", {
+    type: "float",
+    describe: "Percentage increase required for trade (fees not accounted)",
+    default: 0.2,
+  })
+  .version(false).argv
+
+module.exports = async callback => {
+  try {
+    const snxjs = new SynthetixJs({ networkId: networkMap[argv.network] })
```
fleupold, line 28: nit: `await web3.eth.net.getId()`

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,97 @@
...
+      formatedRate * (1 - argv.spread / 100),
+      stableToken,
+      targetToken
+    )
+
+    const buyAmounts = [lowerBuyAmount, upperBuyAmount]
+    const sellAmounts = [lowerSellAmount, upperSellAmount]
+
+    const validFroms = Array(2).fill(batch_index)
+    const validTos = Array(2).fill(batch_index + 1)
```
fleupold, line 77: Is this correct? I thought an order where validFrom == validTo is valid for exactly one batch.

scripts/synthetix/facilicate_trade.js
bh2smith: Ya, I'd be happy to do that. Makes for less contract calls!

scripts/synthetix/facilicate_trade.js
bh2smith: Yes. Forgot about that! Love this suggestion. Although in the essence of explicitness do we really want to expect the reader to just know this?

scripts/synthetix/facilicate_trade.js
bh2smith: This text was copy-pasted from the `place_spread_orders` script in `dex-contracts` and written before I had realized that fees were embedded directly into the orders themselves. Will change this text or possibly just make all the configuration parameters part of the containerized environment. WDYT?

scripts/synthetix/facilicate_trade.js
bh2smith: Great point. I wasn't happy about copying this function over from another repo so, may as well eliminate it!

scripts/synthetix/facilicate_trade.js
bh2smith: Nice find! Thanks a bunch!

scripts/synthetix/facilicate_trade.js
bh2smith: Yes we are. Although this function `calculateBuyAndSellAmountsFromPrice` is not well-documented. That it was it does.

scripts/synthetix/facilicate_trade.js
bh2smith: Does this really work? I will give it a try. Would be much cleaner.

scripts/synthetix/facilicate_trade.js
bh2smith: This may be true, but if it is not, then I will have to submit an issue to `dex-telegram-bot` stating that its validity report is incorrect.

Observe the following two orders (one placed at 28 minutes with one minute expiry and the other at the beginning of a batch with 4 minute expiry).
 
Screenshot 2020-04-07 at 9 48 36 AM


Will take a look at the contract.

scripts/synthetix/facilicate_trade.js
bh2smith: Only problem I have with this then would be that I need to keep a mapping of the token addresses for at least both `rinkeby` and `mainnet` networks. I'll check it out.

scripts/synthetix/facilicate_trade.js
bh2smith: Actually, looking at a comment from below, I will attempt to fetch the spread form the synthetix platform via smart contract call. Zwei Vogel mit Ein Stein.

scripts/synthetix/facilicate_trade.js
bh2smith: All fun and games until you find this error:
```
Error: The send transactions "from" field must be defined!
```

scripts/synthetix/facilicate_trade.js
bh2smith: Yes. Indeed, setting validity period to have equal validFrom and validTo mean that an order is valid for one batch (since the intervals are _invlusive_) as we can see here:

https://github.com/gnosis/dex-contracts/blob/b075120c985d4ad725ab8aaf6a325500907868ca/contracts/BatchExchange.sol#L802-L804

I think @anxolin and @alfetopito will be interested to know this and I will now report an issue in the telegram bot.

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+    )
+    const buyAmounts = [lowerBuyAmount, upperBuyAmount]
+    const sellAmounts = [lowerSellAmount, upperSellAmount]
+
+    // Fetch auction index and declare validity interval for orders.
+    const batch_index = (await exchange.getCurrentBatchId.call()).toNumber()
+    const validFroms = Array(2).fill(batch_index)
+    const validTos = Array(2).fill(batch_index + 1)
+
+    // TODO - use replaceOrder if possible.
```
bh2smith, line 80: Also @fleupold, as you rightfully pointed out, `replaceOrders` is nothing but cancellation followed by new order placement.

https://github.com/gnosis/dex-contracts/blob/b075120c985d4ad725ab8aaf6a325500907868ca/contracts/BatchExchange.sol#L273-L284

This will also be of interest to an issue in `dex-web`(?) regarding a feature request I once made about toping up an existing order that has already been filled. Ahh yes, it was about "Renewing Expired Orders" https://github.com/gnosis/dex-react/issues/389.

Of course, this is still of interest to a front end, but technically does not correspond to updating the `validTo` field of an existing order.

scripts/synthetix/facilicate_trade.js
bh2smith: Done anyway.

scripts/synthetix/facilicate_trade.js
bh2smith: Made the change here, but its a bit uglier that I would have hoped.

```js
    const sUSDTosETHFee = parseFloat(
      snxjs.utils.formatEther(
        await snxjs.Exchanger.feeRateForExchange(sUSDKey, sETHKey)
      )
    )
```

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+    const snxjs = new SynthetixJs({ networkId: await web3.eth.net.getId() })
+    const exchange = await getExchange(web3)
+    const defaultAccount = (await web3.eth.getAccounts())[0]
+
+    // Both of these hardcoded tokens are assumed to have 18 decimal places.
+    // We "trust" that this will always be the case although it seems
+    // that synthetix reserves the authority to upgrade their token
+    // This could mean issuing a new one with a different number of decimals.
+    const sETHKey = await snxjs.sETH.currencyKey()
+    const sUSDKey = await snxjs.sUSD.currencyKey()
```
fleupold, line 29: Can we read the currency key directly from the addresses we load from our trusted list just to make sure they are in line with each other (cf. https://etherscan.io/address/0xAe38b81459d74A8C16eAa968c792207603D84480#readContract)

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // Fetch token IDs and other relevant token information for order placement.
+    const referenceTokenId = (await exchange.tokenAddressToIdMap.call(sUSDAddress[process.env.NETWORK_NAME])).toNumber()
```
fleupold, line 32: nit: I'd reuse the value we get from web3.eth.net.getId() here to make sure we only rely on one way of getting the network ID.

Also for the variable name, would it make sense to call this susdTokenId or the like?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+      )
+    )
+    const sUSDTosETHFee = parseFloat(
+      snxjs.utils.formatEther(
+        await snxjs.Exchanger.feeRateForExchange(sUSDKey, sETHKey)
+      )
+    )
+
+    // Compute buy-sell amounts based on unlimited orders with rates from above.
+    const [lowerBuyAmount, lowerSellAmount] = calculateBuyAndSellAmountsFromPrice(
```
fleupold, line 61: Can we call this etherBuyAmount/susdSellAmount and below susdBuyAmount/etherSellAmount to make it clear which currency is which?

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+    const sETHKey = await snxjs.sETH.currencyKey()
+    const sUSDKey = await snxjs.sUSD.currencyKey()
+
+    // Fetch token IDs and other relevant token information for order placement.
+    const referenceTokenId = (await exchange.tokenAddressToIdMap.call(sUSDAddress[process.env.NETWORK_NAME])).toNumber()
+    const etherTokenId = (await exchange.tokenAddressToIdMap.call(sETHAddress[process.env.NETWORK_NAME])).toNumber()
+
+    const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, [referenceTokenId, etherTokenId])
+    const stableToken = await tokenInfoPromises[referenceTokenId]
+    const targetToken = await tokenInfoPromises[etherTokenId]
```
fleupold, line 37: This is only needed to get the decimal. I'd say we can save these two calls and just hardcode this to 18 decimals (the way they compute their exchange rates seem to only work for tokens with 18 decimals).

This would become even more easy if we read token information from the tokenList.json

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+
+    const buyTokens = [etherTokenId, referenceTokenId]
+    const sellTokens = [referenceTokenId, etherTokenId]
+
+    // Compute Rates and Fees based on price of sETH.
+    // Note that sUSD always has a price of 1 within synthetix protocol.
+    const exchangeRate = await snxjs.ExchangeRates.rateForCurrency(sETHKey)
```
fleupold, line 44: is this denominated in ETH or USD?

scripts/synthetix/facilicate_trade.js
bh2smith: I have to option to simply construct the key directly from the token address if that is what you are referring to? It would, however, require another dependency for just one thing.

```js
const ethers = require("ethers")
ethers.utils.formatBytes32String(‘sETH’)
```

fleupold: Actually that might be a good idea. No need to do a RPC call every 5 minutes for something that is fixed.

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,88 @@
...
+
+const sUSDAddress = {
+  mainnet: "0xAe38b81459d74A8C16eAa968c792207603D84480",
+  rinkeby: "0x1b642a124CDFa1E5835276A6ddAA6CFC4B35d52c",
+}
+
+const sETHAddress = {
+  mainnet: "0xD0DC005d31C2979CC0d38718e23c82D1A50004C0",
+  rinkeby: "0x0647b2C7a2a818276154b0fC79557F512B165bc1",
+}
```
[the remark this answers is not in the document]
bh2smith: Yes. I suspected you were going to suggest this. This bring up yet another error with the telegram bot.

@anxo and/or @alfetopito 

Addresses for both `sUSD` and `sETH` are included in the `tokenList.json` file on rinkeby and mainnet, but the telegram-bot still says "maybe sETH" and "maybe sUSD". NOt sure if there is some other logic governing whether or not telegram-bot knows the tokens or not.

scripts/synthetix/facilicate_trade.js
bh2smith: Yes, this is a goo idea.. the change here will also go along with using the curated `tokenList.json`

scripts/synthetix/facilicate_trade.js
bh2smith: YES! I would be very happy to do so.

scripts/synthetix/facilicate_trade.js
bh2smith: price of `sETH` in `sUSD` from my understanding.

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,84 @@
...
+    // Both of these hardcoded tokens are assumed to have 18 decimal places.
+    // We "trust" that this will always be the case although it seems
+    // that synthetix reserves the authority to upgrade their token
+    // This could mean issuing a new one with a different number of decimals.
+    const sETHKey = ethers.utils.formatBytes32String("sETH")
+    const sUSDKey = ethers.utils.formatBytes32String("sUSD")
+
+    // Avoid querying exchange by tokenAddress for fixed tokenId
+    const buyTokens = [sETH, sUSD].map(token => token.exchangeId)
+    const sellTokens = [sUSD, sETH].map(token => token.exchangeId)
```
fleupold, line 52: Ubernit (feel free to ignore): Can we move these down to where they are used.

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,84 @@
...
+const { SynthetixJs } = require("synthetix-js")
+const ethers = require("ethers")
+const { calculateBuyAndSellAmountsFromPrice } = require("../utils/trading_strategy_helpers")(web3, artifacts)
+
+// These are fixed constants for the current version of the dex-contracts
+const sETHByNetwork = {
+  1: {
+    address: "0x5e74C9036fb86BD7eCdcb084a0673EFc32eA31cb",
+    exchangeId: 8,
+    decmials: 18,
```
fedgiac, line 11: Small typo: should be decimals. (Same error in other three blocks below.)

scripts/synthetix/facilicate_trade.js
anxolin: Yes, that's an issue in the telegram bot. I think TheGraph already takes this into account. Good call to do the issue and PR for that 👍

scripts/synthetix/facilicate_trade.js
```diff
@@ -0,0 +1,84 @@
...
+
+    // Fetch auction index and declare validity interval for orders.
+    // Note that order validity interval is inclusive on both sides.
+    const batchId = (await exchange.getCurrentBatchId.call()).toNumber()
+    const validFroms = Array(2).fill(batchId)
+    const validTos = Array(2).fill(batchId)
+
+    await exchange.placeValidFromOrders(buyTokens, sellTokens, validFroms, validTos, buyAmounts, sellAmounts, {
+      from: defaultAccount,
+    })
```
fedgiac, line 79: The current batch id is used to create these orders. This means that these orders will appear at some point during the current batch. Another possibility is to create the order with batch id `batchId+1`, in this way the orders will be visible for the entire duration of the batch. Do we have some particular agreement with Synthetix about the timing of the price? Who provides the funds, and how?

bh2smith: This won't be a concern for the use case here. Essentially people should assume that they can get the "going rate" in any batch so, in fact we might even place this order in the last minute of a batch so that nobody has a chance to have access to an old price estimate. 

In brief, it is intended that the orders are placed in the current batch.

scripts/synthetix/facilicate_trade.js
bh2smith: Have decided against this because `tokenList` is awkward to access for our purposes here.

scripts/synthetix/facilicate_trade.js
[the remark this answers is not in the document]
alfetopito: https://github.com/gnosis/dex-telegram/issues/185#issuecomment-610586420

scripts/synthetix/facilicate_trade.js
bh2smith: Included in next PR.

PR #129, opened 2020-04-07
[Synthetix] Docker image for order placement script.

bh2smith:
This PR introduces a Dockerfile with a tiny linux image and the project dependencies required to run `facilitate_trade`. Observe that the network must be declared at the time of build, but the private key can be passed in dynamically.

**Testplan**

```sh

docker build --rm -t synthetix-orders .
docker run -e PK=$YOUR_PRIVATE_KEY -t synthetix-orders truffle exec scripts/synthetix/facilicate_trade.js --network rinkeby
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,10 @@ FROM node:10.15-alpine
 
 RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
 COPY package.json .
-RUN yarn
+RUN yarn && yarn global add truffle && yarn cache clean
 # Quick-fix to strange local truffle error.
-RUN yarn global add truffle
 
 COPY . .
 
-ARG NETWORK
-ENV evm_network=$NETWORK
 # Run the command on container startup
-# TODO - replace rinkeby with dynamic env variable
-CMD [ "sh", "-c", "truffle exec scripts/synthetix/facilicate_trade.js --network $evm_network" ]
+CMD ["sh", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:10.15-alpine
+
+RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
+COPY package.json .
+RUN yarn
+# Quick-fix to strange local truffle error.
+RUN yarn global add truffle
+
+COPY . .
+
+ARG NETWORK
+ENV evm_network=$NETWORK
+# Run the command on container startup
+# TODO - replace rinkeby with dynamic env variable
+CMD [ "sh", "-c", "truffle exec scripts/synthetix/facilicate_trade.js --network $evm_network" ]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
+
+FROM node:8.16 as builder
```
fleupold, line 2: Let's use a more up to date and slim version such as `node:10.15-alpine` here.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Create build directory
+WORKDIR /tmp/build
+
+# Install first dependencies, so docker can cache this layer
+COPY package*.json truffle-config.js ./
```
fleupold, line 8: Unless we call install after this line there is no point in having multiple copies.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY contracts contracts
+
+# Then the rest of files
+COPY . .
+
+RUN npm install
```
fleupold, line 14: Let's use yarn

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+RUN npm run networks-inject
+
+
+# Apply cron job
+COPY cron/synthetix-task /etc/crontabs/root
+RUN crontab /etc/crontabs/root
+
+# Run the command on container startup
+CMD [ "sh", "-c", "echo 'Starting container service' && /usr/sbin/crond -f" ]
```
fleupold, line 34: As mentioned over slack, we will handle to "cron" part inside kubernetes. All this container needs to be able to do is run a command when you do something like

`docker run -ti dex-liquidity-provision "npx truffle exec scripts/snx.js --network rinkeby"`

bh2smith: Right, cool, but what about the environment variables?

Dockerfile
bh2smith: Yes. for sure, but don't I have to `npm install yarn` first?

README.md
```diff
@@ -25,4 +25,14 @@ Run test:
...
\ No newline at end of file
+Use scripts as described in `scripts/README.md`.
+
+## Synthetix Spread Orders
+
+This service requires a service that runs a script every 5 minutes. We have configured this to run inside a Docker container that can be interacted with as follows:
+
+```sh
+docker build -t gnosispm/dex-liquidity-provision .
+ export PK=<YOUR_PRIVATE_KEY>
```
fleupold, line 36: When running locally PK needs to be passed into the run command via the `-e` argument right?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10.15-alpine
+
+RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
+
+COPY package.json .
+RUN yarn && yarn global add truffle && yarn cache clean
+
+COPY . .
+
+CMD ["sh", "-c"]
```
fleupold, line 10: I think this should be your ENTRYPOINT (https://goinbigdata.com/docker-run-vs-cmd-vs-entrypoint/). CMD will be overridden by whatever you run the image with. Also maybe use bash since it's more powerful than sh and we already installed it.

bh2smith: Sorry I missed this. New PR with this change #132

PR #131, opened 2020-04-08
[Easy] Bitcore Lib Fix

bh2smith:
Finally a working fix for this dreadful `bitcore-lib` dependency by introducing a "resolution" version.

Test Plan:

Unfortunately I can only describe the test at this point. Worked for me when building a docker image of this project in #129. I recommend checking out that branch and observing that the build works with this, but not without!


fedgiac:
I love this, such a simple solution for this recurring inconvenience!

PR #134, opened 2020-04-09
Compute prices without precision issues

fedgiac:
I introduce two new functions:
1. `getOutputAmountFromPrice` computes how much stable token you get from the given price and target token amount.
2. `getUnlimitedOrderAmounts` computes the largest possible order you can make at a given price.

The goal of this PR is to reduce rounding error, especially when the number of decimals of the two traded tokens are significantly different from each other.

The PR is not yet complete: I'm still waiting for version `0.2.9` of `dex-contracts` to be published on npmjs.com. All tests pass on a local install.

### DIFF
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -180,5 +180,6 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     checkCorrectnessOfDeposits,
     getOutputAmountFromPrice,
     getUnlimitedOrderAmounts,
+    max128,
   }
 }
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -159,11 +159,11 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   /**
    * Computes the stable and target token amounts needed to set up an unlimited order in the exchange
    * @param {number} price amount of stable tokens in exchange for one target token
-   * @param {integer} stableTokenDecimals number of decimals of the stable token
    * @param {integer} targetTokenDecimals number of decimals of the target token
+   * @param {integer} stableTokenDecimals number of decimals of the stable token
    * @return {BN[2]} amounts of stable token and target token for an unlimited order at the input price
    */
-  const getUnlimitedOrderAmounts = function(price, stableTokenDecimals, targetTokenDecimals) {
+  const getUnlimitedOrderAmounts = function(price, targetTokenDecimals, stableTokenDecimals) {
     let targetTokenAmount = max128.clone()
     let stableTokenAmount = getOutputAmountFromPrice(price, targetTokenAmount, targetTokenDecimals, stableTokenDecimals)
     if (stableTokenAmount.gt(targetTokenAmount)) {
@@ -171,7 +171,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
       targetTokenAmount = getOutputAmountFromPrice(1/price, stableTokenAmount, stableTokenDecimals, targetTokenDecimals)
       assert(stableTokenAmount.gte(targetTokenAmount), "Error: unable to create unlimited order")
     }
-    return [stableTokenAmount, targetTokenAmount]
+    return [targetTokenAmount, stableTokenAmount]
   }
 
   return {
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -2,6 +2,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const BN = require("bn.js")
   const exchangeUtils = require("@gnosis.pm/dex-contracts")
+  const { Fraction } = require("@gnosis.pm/dex-contracts/src")
 
   const { fetchTokenInfoFromExchange } = require("./trading_strategy_helpers")(web3, artifacts)
 
@@ -139,9 +140,28 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return true
   }
 
+  /**
+   * Computes the amount of output token units from their price and the amount of input token units
+   * Note that the price is expressed in terms of tokens, while the amounts are in terms of token units
+   * @param {number} price amount of output token in exchange for one input token
+   * @param {BN} inputTokenAmount amount of token units that are exchanged at price
+   * @param {integer} inputDecimals number of decimals of the input token
+   * @param {integer} outputDecimals number of decimals of the output token
+   * @return {BN} amount of output token units obtained
+   */
+  const getOutputAmountFromPrice = function(price, inputAmount, inputDecimals, outputDecimals) {
+    const priceFraction = Fraction.fromNumber(price)
+    const unitPriceFraction = priceFraction.mul(
+      new Fraction(new BN(10).pow(new BN(outputDecimals)), new BN(10).pow(new BN(inputDecimals)))
+    )
+    const outputAmountFraction = unitPriceFraction.mul(new Fraction(inputAmount, 1))
+    return outputAmountFraction.toBN()
+  }
+
   return {
     isPriceReasonable,
     areBoundsReasonable,
     checkCorrectnessOfDeposits,
+    getOutputAmountFromPrice,
   }
 }
--- a/scripts/utils/price-utils.js
+++ b/scripts/utils/price-utils.js
@@ -6,6 +6,8 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
 
   const { fetchTokenInfoFromExchange } = require("./trading_strategy_helpers")(web3, artifacts)
 
+  const max128 = new BN(2).pow(new BN(128)).subn(1)
+
   const checkCorrectnessOfDeposits = async (
     currentPrice,
     bracketAddress,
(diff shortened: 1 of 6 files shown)


scripts/utils/price-utils.js
```diff
@@ -134,9 +137,48 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+   * @param {integer} stableTokenDecimals number of decimals of the stable token
+   * @return {BN[2]} amounts of stable token and target token for an unlimited order at the input price
+   */
+  const getUnlimitedOrderAmounts = function(price, targetTokenDecimals, stableTokenDecimals) {
+    let targetTokenAmount = max128.clone()
+    let stableTokenAmount = getOutputAmountFromPrice(price, targetTokenAmount, targetTokenDecimals, stableTokenDecimals)
+    if (stableTokenAmount.gt(targetTokenAmount)) {
+      stableTokenAmount = max128.clone()
+      targetTokenAmount = getOutputAmountFromPrice(1/price, stableTokenAmount, stableTokenDecimals, targetTokenDecimals)
+      assert(stableTokenAmount.gte(targetTokenAmount), "Error: unable to create unlimited order")
```
josojo, line 172: I can not imagine any case where this would throw an error!

scripts/utils/price-utils.js
```diff
@@ -134,9 +137,48 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+
+
+  /**
+   * Computes the stable and target token amounts needed to set up an unlimited order in the exchange
+   * @param {number} price amount of stable tokens in exchange for one target token
+   * @param {integer} targetTokenDecimals number of decimals of the target token
+   * @param {integer} stableTokenDecimals number of decimals of the stable token
+   * @return {BN[2]} amounts of stable token and target token for an unlimited order at the input price
+   */
+  const getUnlimitedOrderAmounts = function(price, targetTokenDecimals, stableTokenDecimals) {
```
josojo, line 166: Much simpler logic and easier to understand. Nice!

test/priceUtils.js
```diff
@@ -1,5 +1,124 @@
...
+        inputDecimals: 0,
+        outputDecimals: 70,
+        expectedOutputAmountString: "10.1",
+      },
+    ]
+    for (const { price, inputAmountString, inputDecimals, outputDecimals, expectedOutputAmountString } of testCases) {
+      const inputAmount = toErc20Units(inputAmountString, inputDecimals)
+      const expectedOutputAmount = toErc20Units(expectedOutputAmountString, outputDecimals)
+      const outputAmount = getOutputAmountFromPrice(price, inputAmount, inputDecimals, outputDecimals)
+      assertEqualUpToFloatPrecision(outputAmount, expectedOutputAmount)
```
josojo, line 57: It's nice that you first multiply and then later divide by the inputDecimials  and outputDecimals, as this allows to easy test verification

scripts/utils/price-utils.js
fedgiac: I also think this shouldn't ever throw, it's there just in case there was some very unexpected rounding issue since I use `1/price` instead of `price`. This is also why I added `1+Number.EPSILON` and `1-Number.EPSILON` to the test cases, to be sure that nothing like that was going on.

scripts/utils/price-utils.js
```diff
@@ -2,6 +2,9 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const BN = require("bn.js")
   const exchangeUtils = require("@gnosis.pm/dex-contracts")
+  const { Fraction } = require("@gnosis.pm/dex-contracts/src")
+
+  const max128 = new BN(2).pow(new BN(128)).subn(1)
```
bh2smith, line 7: I thought this value was already declared in another file somewhere. Couldn't we just import it?

scripts/utils/price-utils.js
```diff
@@ -134,9 +137,48 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+  /**
+   * Computes the amount of output token units from their price and the amount of input token units
+   * Note that the price is expressed in terms of tokens, while the amounts are in terms of token units
+   * @param {number} price amount of output token in exchange for one input token
+   * @param {BN} targetTokenAmount amount of token units that are exchanged at price
+   * @param {integer} targetTokenDecimals number of decimals of the input token
+   * @param {integer} stableTokenDecimals number of decimals of the output token
+   * @return {BN} amount of output token units obtained
+   */
+  const getOutputAmountFromPrice = function(price, targetTokenAmount, targetTokenDecimals, stableTokenDecimals) {
```
bh2smith, line 149: Love the documentation! This function looks a lot like an already existing (undocumented function) `computeBuyAndSellAmountsFromPrice`. Do you think this would be an appropriate PR to include some doc strings on that as well?

scripts/utils/price-utils.js
bh2smith: I am wondering if we sometimes switch the nothign of "stableToken" and "targetToken" like, for example, when computing the bracket orders. If this is the case, then I am beginning to think that sometimes the "stableToken" is being passed into this function as the `targetToken`and the names have become misleading. If this is the case, would we agree that this naming convention could be improved? I wouldn't know what to suggest, but I suppose if there are always only two tokens and the calculations are all made relative to each other then the `referenceToken` is always stable relative to the `targetToken`. Could these be a more name pair?

test/dfusionMultiSafes.js
```diff
@@ -21,11 +21,10 @@ const {
   buildTransferFundsToMaster,
   buildWithdrawAndTransferFundsToMaster,
   isOnlySafeOwner,
-  max128,
   maxU32,
 } = require("../scripts/utils/trading_strategy_helpers")(web3, artifacts)
 const { waitForNSeconds, execTransaction } = require("../scripts/utils/internals")(web3, artifacts)
-const { checkCorrectnessOfDeposits } = require("../scripts/utils/price-utils")(web3, artifacts)
```
bh2smith, line 28: I see now, that it was moved from `helpers` to `utils`. If we are going to move this then we should also be moving `maxU32` along with it. Otherwise I would say that the change is not justified.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -609,15 +588,13 @@ withdrawal of the desired funds
     buildWithdrawAndTransferFundsToMaster,
     buildBracketTransactionForTransferApproveDeposit,
     buildTransferApproveDepositFromOrders,
-    calculateBuyAndSellAmountsFromPrice,
```
bh2smith, line 612: Does this no longer exist? I am currently using this function in the `synthetix-liqudity` script.  I suppose it is my own fault for not having a test for this, but I believe there are at least two places in the code base depending on this function so we had better make sure to accommodate for this change. Alternatively, and probably more appropriately considering the size of this change already, I might propose that this be removed along with the necessary adjustments in a separate PR.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -282,35 +290,6 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     return buildBundledTransaction(transactions)
   }
 
-  // TODO - pass stableTokenDecmials and targetTokenDecimals since these are the only attributes used here.
-  const calculateBuyAndSellAmountsFromPrice = function(price, stableToken, targetToken) {
```
bh2smith, line 286: Careful here, this function is used by other scripts that haven't been touched here. I have expanded on this below.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -246,10 +246,18 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+          stableToken.decimals
+        )
         // While the first bracket-order trades standard_token against target_token, the second bracket-order trades
         // target_token against standard_token. Hence the buyAmounts and sellAmounts are switched in the next line.
-        const [lowerBuyAmount, lowerSellAmount] = calculateBuyAndSellAmountsFromPrice(lowerLimit, stableToken, targetToken)
+        const [lowerBuyAmount, lowerSellAmount] = getUnlimitedOrderAmounts(
+          lowerLimit,
+          targetToken.decimals,
+          stableToken.decimals
+        )
```
bh2smith, line 260: Not sure this this was pointed out, but you will see that the `targetToken` and `stableToken` were switched here when the new function was introduced. Was this intended or is this a fatal mistake?

scripts/utils/trading_strategy_helpers.js
bh2smith: After digging deeper, I see that there is only one file remaining that still needs to be adjusted for this change. That is here. 

https://github.com/gnosis/dex-liquidity-provision/blob/2112744c69e02754d53fdc02ea4b7dce06ac4cbf/scripts/synthetix/facilitate_trade.js#L62-L64

I will make sure to include a unit test for this script ASAP, but we need to be careful of things like this.

bh2smith: This was taken care of now in #136

scripts/utils/price-utils.js
fedgiac: I agree that many functions in this repo are prone to having stable and target token mixed up. As I see it, the main discriminant between "stable" and "target" is which price you are using: the price `p` always means `1` target token in exchange for `p` stable tokens. I'm not sure how to improve on the current naming system, though. Never give the price alone, but have a "price triplet" with price, "stable" and "target"? But what when there is no price?

scripts/utils/trading_strategy_helpers.js
fedgiac: It's intended: the function `getUnlimitedOrderAmounts` puts target token first and then stable token, while `calculateBuyAndSellAmountsFromPrice` puts stable first and then target. I did this because we use target first then stable in most (but not all!) functions in this repo.

fedgiac:
Sorry about the issue with your work on the Synthetix part of this repo, I'm very much (too much?) relying on tests to catch these kind of issues. Next time I'll do a repo-wide search for all functions I remove.

bh2smith:
> I'm very much (too much?) relying on tests to catch these kind of issues.

No worries, I am going to make tests ASAP (i.e. my top priority). I believe I remember a time when I had to convince you that unit testing was important... or did I just dream that!?

PR #136, opened 2020-04-09
[HotFix] Use Safe Gas Station's Fast Gas Price.

bh2smith:
We are experiencing trouble with our synthetix-liquidity service. Somehow the gas estimates on transaction submission were incredibly low (1 Gwei) and now there is a long list of pending transactions that are completely stuck (https://etherscan.io/address/0x5ce64e024d54eee4e6409829cf0ff8529e5da170). This means that none of the orders will be placed untill this account “catches up”. I have tried cancelling these transactions by overwriting them in geth with a higher gas price with no success.

For now this order placement service has been stopped and I am waiting for the transactions to catch up.

Ideally what I would like to do is use

```
gasPrice = max(ethGasStation["average"], 5)
```
So that we always set the price to at least 5 Gwei.


fedgiac:
What about having a contract that fetches the price directly from Synthetix and then places the orders? In this way a delayed transaction would not fail, but it would just work in the next batch. You'd only have an active transaction at a time, and if it has not been accepted yet you increase its gas price. This might also save you some gas, assuming that the reading of the price from Synthetix is not too much higher than a storage read (200 gas), since you currently spend 4300 gas just to send the transaction data to the network.

scripts/synthetix/facilitate_trade.js
```diff
@@ -30,9 +33,15 @@ const sUSDByNetwork = {
...
 
+const gasStationURL = {
+  1: "https://safe-relay.gnosis.io/api/v1/gas-station/",
+  4: "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/"
+}
+
 module.exports = async callback => {
   try {
     const networkId = await web3.eth.net.getId()
+    const gasPrices = await (await fetch(gasStationURL[networkId])).json()
```
fleupold, line 44: Is the outer await necessary? I'd think that .json() is a synchronous method.

scripts/synthetix/facilitate_trade.js
```diff
@@ -60,8 +67,16 @@ module.exports = async callback => {
...
+    const [buyETHAmount, sellSUSDAmount] = getUnlimitedOrderAmounts(
+      formatedRate * (1 - sUSDTosETHFee),
+      sUSD.decimals,
+      sETH.decimals
+    )
+    const [sellETHAmount, buySUSDAmount] = getUnlimitedOrderAmounts(
+      formatedRate * (1 + sETHTosUSDFee),
+      sETH.decimals,
+      sUSD.decimals
+    )
```
fedgiac, line 79: `getUnlimitedOrderAmounts` has a different order input than `calculateBuyAndSellAmountsFromPrice`: the order of stable token and target token is swapped (see discussion in PR #134). Both tokens have 18 decimals, so it's not an urgent issue, but the decimals should be swapped  in both calls to avoid possible future issues.

bh2smith: Thanks, this is what I wanted to know. Although I don't see any good reason for changing the order.

PR #137, opened 2020-04-10
Consistent .js file names

### LINKED ISSUE #122: [Easy] Inconsistent file name conventions
In the scripts directory alone we are using all of `snake_case`, `camelCase` and `kebab-case`. According to this JS style guide it appears that snake_case or kebab-case are the standard. 

I would propose sticking with one or the other, but not both. 

Note that kebab case is used for github repo names and node package naming so maybe we should stick with snake case for filenames to avoid conflict. 

Suggestions?

fedgiac:
Rename all Javascript files sticking to snake case convention. Fixes #122.


.travis.yml
```diff
@@ -14,4 +14,3 @@ script:
   - yarn lint
   - yarn pretty-check
   - yarn test
-  - npx mocha test/printingTools.js
```
josojo, line 17: How is this test called, once this line is removed?

bh2smith: I don't understand the question here. Is it called or not? Why are we removing it?

fedgiac:
Sorry I didn't motivate the test change properly, probably it should have been its own micro-PR.
This line just uses Mocha directly for running the tests in the file instead of going through Truffle because it's much quicker. It comes from the time when we were including every single new test file manually into Travis.
The same test is now run while running `yarn test`. You can see [the build an old PR](https://travis-ci.com/github/gnosis/dex-liquidity-provision/builds/159668708) to see that a test is run twice (Ctrl+f `toErc20Units`). In this PR [it's still running, but only once](https://travis-ci.com/github/gnosis/dex-liquidity-provision/builds/159762873#L445).
I changed it because the Travis build of this PR was failing, since I renamed the file it was testing.

PR #140, opened 2020-04-13
Fix check deposits

fedgiac:
This PR unearthed two issues in the function `checkCorrectnessOfDeposits`:
1. the function made no sense for tokens with digits!=18, yet no test failed because of that.
2. the function compared two BN using literal `<` and `>` instead of `.lt`, `.gt`.

Point 2. greatly aggravated my debugging of point 1.: when using `<` or `>` for comparison, the BNs are converted to strings and then lexicographically compared. This means that as long as the strings have the same length then the comparison works as expected, but if the length is different then it might yield unwelcomed surprises: for example, `"101"<"11" === true`.

I added a test case which would have failed with the code as it was before this PR and I rewrote the function to be, I hope, more straightforward. I would gladly accept comments to improve readability.

Finally, I created yet another test case to check whether the function works with extreme values for prices and number of decimals. ~~The test fails, even after this PR, so I left it commented out. Rewriting the function helped me to identify that the failure point likely lies somewere else, namely in the order creation: the price at which the brackets buy appears to be higher than the price at which the brackets sell, sometimes. I'll look into this and fix it in another PR, in the best case this issue is limited to the very unrealistic ranges and prices used.~~ i accidentally swapped highest and lowest limit in the test. Issue will be permanently fixed with #141.

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -17,6 +17,7 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     investmentStableTokenPerBracket,
     investmentTargetTokenPerBracket
   ) => {
+    const currentUnitPrice = getUnitPrice(currentPrice, await targetToken.decimals(), await stableToken.decimals())
     const bracketExchangeBalanceStableToken = (await exchange.getBalance(bracketAddress, stableToken.address)).toString()
     const bracketExchangeBalanceTargetToken = (await exchange.getBalance(bracketAddress, targetToken.address)).toString()
     const auctionElements = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders.call(bracketAddress))
@@ -34,9 +35,9 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
     const sellTargetTokenOrder = sellTargetTokenOrders[0]
 
     // Check that tokens with a lower price than the target price are funded with stableTokens
-    if (checkThatOrderPriceIsBelowTarget(currentPrice, sellStableTokenOrder)) {
+    if (checkThatOrderPriceIsBelowTarget(currentUnitPrice, sellStableTokenOrder)) {
       // checks whether price is in middle of bracket:
-      if (checkThatOrderPriceIsBelowTarget(1 / currentPrice, sellTargetTokenOrder)) {
+      if (checkThatOrderPriceIsBelowTarget(currentUnitPrice.inverted(), sellTargetTokenOrder)) {
         assert(
           checkFundingInTheMiddleBracket(
             bracketExchangeBalanceStableToken,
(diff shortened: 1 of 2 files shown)


scripts/utils/price_utils.js
```diff
@@ -17,64 +17,51 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+          investmentStableTokenPerBracket,
+          investmentTargetTokenPerBracket
+        )
+      )
     }
   }
 
-  const checkThatOrderPriceIsBelowTarget = function(currentPrice, order) {
-    const multiplicator = 1000000000
-    return new BN(currentPrice * multiplicator).mul(order.priceNumerator) > order.priceDenominator.mul(new BN(multiplicator))
```
josojo, line 75: good catch that this does not work

test/dfusion_multi_safes.js
```diff
@@ -437,6 +470,25 @@ contract("GnosisSafe", function(accounts) {
           await testAutomaticDeposits(tradeInfo, expectedDistribution)
         }
       })
+      /*it("with extreme prices and decimals", async () => {
```
josojo: Is this one already working?

scripts/utils/price_utils.js
```diff
@@ -17,64 +17,51 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
-            bracketExchangeBalanceStableToken,
-            bracketExchangeBalanceTargetToken,
-            investmentStableTokenPerBracket,
-            investmentTargetTokenPerBracket
-          )
-        )
-      } else {
-        assert.equal(bracketExchangeBalanceTargetToken, investmentTargetTokenPerBracket.toString())
-      }
+    if (priceSellingTargetToken.lt(currentUnitPrice)) {
```
josojo, line 50: Much more concise logic. Great!

test/dfusion_multi_safes.js
```diff
@@ -437,6 +470,25 @@ contract("GnosisSafe", function(accounts) {
           await testAutomaticDeposits(tradeInfo, expectedDistribution)
         }
       })
+      it("with extreme prices and decimals", async () => {
+        const tradeInfo = {
+          fleetSize: 4,
+          lowestLimit: 5e194,
+          highestLimit: 20e194,
+          currentPrice: 10e194,
+          bracketsWithStableTokenDeposit: 2,
```
josojo, line 479: Will this variable be used?

test/dfusion_multi_safes.js
fedgiac: Yep!

test/dfusion_multi_safes.js
fedgiac: Nope, I removed it, thanks.

bh2smith:
I suppose you stumbled upon this error when trying to make the tests work with tokens having decimals different from 18. Very nice find and very nice fix. Yet another argument for using a more sophisticated (type-safe) language.

PR #141, opened 2020-04-14
Fail order construction with bad limits

fedgiac:
Currently, we can create orders for which the lowest limit is higher than the highest limit. Everything goes through smoothly and all brackets end up unprofitable, losing money from the point they receive funds.
This PR makes `buildOrder` fail in this scenario.


bh2smith:
Yes! Very cool addition!

PR #145, opened 2020-04-15
Update README.md

c3rnst:
name change.


fedgiac:
You probably also want to change line 17 of the same file: you can modify `(this is needed to deploy Dfusion)` to `(this is needed to deploy the exchange smart contract)`.

PR #152, opened 2020-04-20
Verify script - check master owners

fedgiac:


scripts/utils/verify_scripts.js
```diff
@@ -11,20 +11,40 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
+    else {
+      log("- Verify owners of masterSafe")
+      const threshold = master.getThreshold()
+      const owners = master.getOwners()
+      assert.equal(
+        await threshold,
+        masterThreshold,
+        "Master threshold is " + (await threshold) + " while it is supposed to be " + masterThreshold
+      )
+      assert(equalUpToOrder(await owners, masterOwners), "Master owners are different than expected")
```
josojo, line 45: In order to save some code, would here also work something like this:
```
assert.isDeepEqual((await owners).sorted, masterOwners.sorted, "")
```
?

scripts/utils/verify_scripts.js
```diff
@@ -11,20 +11,40 @@ module.exports = function(web3 = web3, artifacts = artifacts) {
...
   const GnosisSafeProxy = artifacts.require("GnosisSafeProxy.sol")
   const gnosisSafeMasterCopy = GnosisSafe.deployed()
 
   const pageSize = 50
 
-  const verifyBracketsWellFormed = async function(masterAddress, bracketAddresses, logActivated = false) {
+  const verifyBracketsWellFormed = async function(
+    masterAddress,
+    bracketAddresses,
+    masterThreshold,
```
josojo, line 25: Maybe you wanna indicate that these values do not have to be set by providing default values:
```
masterThreshold = null
```

scripts/utils/verify_scripts.js
fedgiac: Very nice, this is much cleaner than creating my own comparison function.

Issue #78, opened 2020-03-16
Deposit file path relative to script path instead of user path 


fedgiac:
In some scripts the user must give the path of a file as an argument (e.g., when withdrawing and depositing). If the path specification is relative, then the script uses the path of the script.  It would be more natural for a user to give a path relative to the console working directory.

fedgiac:
Solved with #155 and #166.

PR #155, opened 2020-04-22
Reorganize withdraw script

fedgiac:
Move the logic of the withdraw script to its own testable function. Tests are added.

This PR inadvertently fixed issue #78 for the withdraw script. Another PR should fix the issue for all scripts, otherwise we would have an incostistent behavior between scripts when managing filesystem paths.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "ganache-cli": "6.9.0",
     "prettier": "^1.19.1",
     "run-with-testrpc": "^0.3.0",
+    "tmp-promise": "^2.0.2",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
   },
   "scripts": {
@@ -44,7 +45,6 @@
     "openzeppelin-solidity": "^2.4.0",
     "sol-merger": "^2.0.1",
     "synthetix-js": "^2.21.6",
-    "tmp-promise": "^2.0.2",
     "truffle": "^5.1.18",
     "truffle-plugin-verify": "^0.3.10"
   },
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "openzeppelin-solidity": "^2.4.0",
     "sol-merger": "^2.0.1",
     "synthetix-js": "^2.21.6",
+    "tmp-promise": "^2.0.2",
     "truffle": "^5.1.18",
     "truffle-plugin-verify": "^0.3.10"
   },
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -19,7 +19,7 @@ module.exports = function(web3, artifacts) {
     }
   }
 
-  const getAmount = async function(argv, bracketAddress, tokenInfo, exchange, printOutput = false) {
+  const getMaxWithdrawableAmount = async function(argv, bracketAddress, tokenInfo, exchange, printOutput = false) {
     const log = printOutput ? (...a) => console.log(...a) : () => {}
     let amount
     const token = tokenInfo.instance
@@ -59,7 +59,7 @@ module.exports = function(web3, artifacts) {
         withdrawals.map(async withdrawal => ({
           bracketAddress: withdrawal.bracketAddress,
           tokenAddress: withdrawal.tokenAddress,
-          amount: await getAmount(
+          amount: await getMaxWithdrawableAmount(
             argv,
             withdrawal.bracketAddress,
             await tokenInfoPromises[withdrawal.tokenAddress],
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -1,0 +1,100 @@
+module.exports = function(web3, artifacts) {
+  const { fromErc20Units, shortenedAddress } = require("./utils/printing_tools")
+  const {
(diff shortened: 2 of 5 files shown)


scripts/wrapper/withdraw.js
```diff
@@ -0,0 +1,100 @@
...
+  const {
+    getExchange,
+    fetchTokenInfoForFlux,
+    buildRequestWithdraw,
+    buildWithdraw,
+    buildTransferFundsToMaster,
+    buildWithdrawAndTransferFundsToMaster,
+  } = require("./utils/trading_strategy_helpers")(web3, artifacts)
+
+  const getAmount = async function(argv, bracketAddress, tokenInfo, exchange, printOutput = false) {
```
bh2smith, line 12: What is "Amount" here in "getAmount"

fedgiac:
Yep! Testing already caught that the import paths I use are missing an extra `../`.

scripts/wrapper/withdraw.js
```diff
@@ -0,0 +1,116 @@
...
+      const pendingWithdrawal = await exchange.getPendingWithdraw(bracketAddress, tokenInfo.address)
+      if (pendingWithdrawal[1].toNumber() == 0) {
+        log("Warning: no withdrawal was requested for address", bracketAddress, "and token", tokenInfo.symbol)
+        amount = "0"
+      }
+      if (amount != "0" && pendingWithdrawal[1].toNumber() >= currentBatchId) {
+        log("Warning: amount cannot be withdrawn from the exchange right now, withdrawing zero")
+        amount = "0"
+      }
+      amount = pendingWithdrawal[0].toString()
```
josojo, line 38: It sucks that we don't have a proper way to determine the withdrawal balance in the smart contract. But it should be the max amount of 
```
 max((await token.balanceOf(bracketAddress)).toString(), pendingWithdrawal[0].toString())
```

(just noting this here, I guess its anyways just copy paste code)

test/scripts/withdraw.js
```diff
@@ -0,0 +1,254 @@
...
+        )
+      }
+    }
+  }
+  it("requests withdrawals", async () => {
+    const amounts = [{ token: { decimals: 18, symbol: "DAI" }, amount: "1000" }]
+    const [masterSafe, bracketAddresses, tokenInfo] = await setup(2, amounts)
+    const deposits = evenDeposits(bracketAddresses, tokenInfo[0], "1000")
+    await deposit(masterSafe, deposits)
+    const depositFile = await tmp.file()
```
josojo, line 109: do we really need the async part of the tmp-promise package? Wouldn't the sync tmp package suffice?

bh2smith: What is this new dependency doing for us? Weren't we able to read the deposit file before without the additional dependency?

package.json
```diff
@@ -44,6 +44,7 @@
     "openzeppelin-solidity": "^2.4.0",
     "sol-merger": "^2.0.1",
     "synthetix-js": "^2.21.6",
+    "tmp-promise": "^2.0.2",
```
bh2smith, line 47: This is only used in unit tests and should thus be in `dev-dependencies`. In fact, there are probably many items here that are only `dev-dependencies` which should be moved (e.g. `eth-gas-reporter`, `eslint`, ...)

scripts/wrapper/withdraw.js
fedgiac: You are saying that the user might get less than the amount in the pending withdrawal, and thus the multisend withdraw+transferFundsToMaster would fail, right? This is a very good point, the fix is going to be ugly however. The only way I see is to use `web3.getStorageAt` and manually compute the storage position of `balanceState[braketAddress]`. Alternatively, we give up on the bundled transaction and we make the user do withdraw and transferFundsToMaster separately, so a grand total of three transactions to withdraw.

test/scripts/withdraw.js
fedgiac: As it is right now, we don't need the async version because I prioritized readability. On the other hand, there is also no reason to avoid the async version, since it doesn't really add any complexity and actually has a more intuitive naming convention IMHO. (For example, in the sync version `.path` is `.name` and `.cleanup()` becomes a separate function `cleanupCallback()`.)

fedgiac:
The new dependency (`tmp-promise`) is used to create a temporary file in which to store the deposits. The file is used to test the creation of deposits from file.
I use a package because of:
- compatibility, this should work on different OSes
- [security](https://en.wikipedia.org/wiki/Symlink_race)

A reasonable alternative is to create a temporary file in the folder where the script is executed. I believe this to be a less clean solution. Also, the file is currently not deleted if a test throws, which would pollute the execution path with an extra file. I see the file pollution on test failure as it is now as less of an issue since a temporary folder would be cleared after a reboot and we don't need to run tests in a production environment, where a constant increase of storage space would be problematic.

scripts/wrapper/withdraw.js
josojo: Exactly,

lets go for the `web3.getStorageAt` route, as longterm we want to allow more and more users to use these liquidity provision scripts. (maybe even with front-end support)

let me create an issue for that.

PR #157, opened 2020-04-23
lodash vulnerability

bh2smith:
Fixes the vulnerability depicted in the screenshot below

Screenshot 2020-04-23 at 8 42 34 AM

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4418,15 +4418,10 @@ lodash.flatmap@^4.5.0:
   resolved "https://registry.yarnpkg.com/lodash.flatmap/-/lodash.flatmap-4.5.0.tgz#ef8cbf408f6e48268663345305c6acc0b778702e"
   integrity sha1-74y/QI9uSCaGYzRTBcaswLd4cC4=
 
-lodash@4.17.15, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4:
+lodash@4.17.15, lodash@=4.17.11, lodash@^4.17.12, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4:
   version "4.17.15"
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.15.tgz#b447f6670a0455bbfeedd11392eff330ea097548"
   integrity sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A==
-
-lodash@=4.17.11:
-  version "4.17.11"
-  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.11.tgz#b39ea6229ef607ecd89e2c8df12536891cac9b8d"
-  integrity sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==
 
 loose-envify@^1.0.0, loose-envify@^1.4.0:
   version "1.4.0"
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "truffle-plugin-verify": "^0.3.10"
   },
   "resolutions": {
-    "bitcore-lib": "8.1.1"
+    "bitcore-lib": "8.1.1",
+    "lodash": "^4.17.12"
   }
 }


yarn.lock
```diff
@@ -4418,16 +4418,11 @@ lodash.flatmap@^4.5.0:
   resolved "https://registry.yarnpkg.com/lodash.flatmap/-/lodash.flatmap-4.5.0.tgz#ef8cbf408f6e48268663345305c6acc0b778702e"
   integrity sha1-74y/QI9uSCaGYzRTBcaswLd4cC4=
 
-lodash@4.17.15, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4:
+lodash@4.17.15, lodash@=4.17.11, lodash@^4.17.12, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4:
```
fedgiac, line 4421: I'm not familiar with resolutions or dependencies in general, but it looks like `lodash@=4.17.11` is still there?

bh2smith: Yes. I don't understand what happened here either. I think these are all the versions that were requested from sub dependencies, and immediately below shows the resolution value that was chosen. I was just trying to fix the issue without manually editing the lock file. I wonder if @nlordell has a suggestion on how to do this better, or what happened here.

bh2smith:
> Thanks for tanking care of this. Just updating all our packages doesn't do the trick?

That is a good question. I suppose I could figure out which of our dependencies has this silly thing as its sub dependency.

josojo:
I would have simply waited until the packages of depending on loaddash update themselves. For an easy notification, I activated the dependabot

yarn.lock
nlordell:
```
lodash@4.17.15, lodash@=4.17.11, lodash@^4.17.12, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.4:
  version "4.17.15"
  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.15.tgz#b447f6670a0455bbfeedd11392eff330ea097548"
```

From what I understand of the lock file syntax, it is basically saying that all `package.json` dependencies that request the versions `4.17.15, =4.17.11, ^4.17.12, ^4.17.13, ^4.17.14, ^4.17.15, ^4.17.4` resolve to the actual version `4.17.15`.

When I read this, to me it suggests that only `4.17.15` is actually being used, and that there are 7 different requested versions that resolve to this.

fedgiac: Does it mean that when you set a "resolution" in your package.json it might make yarn ignore strict requirements like `lodash@=4.17.11`?
So, if somebody used resolutions there'd be the chance of using broken packages because they'd use wrong dependencies?
Or maybe `@=` doesn't actually mean equal and there is some leeway?

nlordell: > Does it mean that when you set a "resolution" in your package.json it might make yarn ignore strict requirements like lodash@=4.17.11?

From what I understand that is precisely [a use case of `resolution`s](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this).

> So, if somebody used resolutions there'd be the chance of using broken packages because they'd use wrong dependencies?

Yes, I think that is why `resolution`s should be used sparingly. I think patching security vulnerabilities in transient dependencies in an appropriate use for them (obviously, some sort of testing is required to ensure that this does not break anything).

> Or maybe @= doesn't actually mean equal and there is some leeway?

As far as I know `=VERSION` specs mean exact matches.

nlordell:
> I would have simply waited until the packages of depending on loaddash update themselves. For an easy notification, I activated the dependabot

Just so you know I've had issues recently with dependabot not updating transient dependencies.

yarn.lock
fedgiac: Scary, I thought they'd try to enforce a single version of the package as long as the prerequisites were met. Very good to know that this is not the case!

nlordell: Im actually not sure what it does in that situation. I’d imagine it would try to find resolve multiple version requests with as few resolved (as in in the lock file, not in the `resolve` section) versions as possible.

Note that here it’s not possible because `=4.17.11` and `^4.17.12` have no overlap

bh2smith:
Does anyone think this should be reopened and merged?

fedgiac:
Now that I understand better what it does, I think we shouldn't merge this PR.
I checked what's the source of this issue, and apparently it comes from our old friend `bitcore-lib`, which requires the exact version `lodash@4.17.11`, one less than the first that it's not affected by the bug. If we could unresolve `bitcore-lib` from 8.1.1 then `bitcore-lib` would be updated and this issue would be fixed. We can't remove this resolution because `eth-lightwallet` requires this old version of `bitcore-lib`, and `eth-lightwallet` is not being maintained so tough luck getting an update fixing this issue.
We could solve this by removing our dependency on `eth-lightwallet`. I believe we used it only for testing, it's one of the packages included by Richard in his toy example on how to work with multiple Safes and has never really been touched ever since. I think there are a couple of todos in the code suggesting to remove lightwallets already, I'd be in favor of this.

bh2smith:
@fedgiac YES! This is a great idea! Lets go for it!

PR #166, opened 2020-04-23
Fix relative paths in deposit script

fedgiac:
Fixes issue #78 and makes the behavior of the deposit script consistent with that of the withdraw script after PR #155. With this change, the command-line examples in the readme become correct.

### DIFF
--- a/scripts/transfer_approve_deposit.js
+++ b/scripts/transfer_approve_deposit.js
@@ -19,7 +19,7 @@ module.exports = async (callback) => {
     const masterSafe = await GnosisSafe.at(argv.masterSafe)
 
     const deposits = require(argv.depositFile)
-    // TODO - make a simpler to construct deposit file style
+    // TODO - make it simpler to construct deposit file style
     console.log("Preparing transaction data...")
     const transaction = await buildTransferApproveDepositFromList(masterSafe.address, deposits, true)
 
--- a/scripts/transfer_approve_deposit.js
+++ b/scripts/transfer_approve_deposit.js
@@ -21,7 +21,7 @@ module.exports = async (callback) => {
     const masterSafe = await GnosisSafe.at(argv.masterSafe)
 
     const deposits = JSON.parse(await fs.readFile(argv.depositFile, "utf8"))
-    // TODO - make it simpler to construct deposit file style
+
     console.log("Preparing transaction data...")
     const transaction = await buildTransferApproveDepositFromList(masterSafe.address, deposits, true)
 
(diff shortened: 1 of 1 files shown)


scripts/transfer_approve_deposit.js
```diff
@@ -18,8 +20,8 @@ module.exports = async (callback) => {
     const GnosisSafe = artifacts.require("GnosisSafe")
     const masterSafe = await GnosisSafe.at(argv.masterSafe)
 
-    const deposits = require(argv.depositFile)
-    // TODO - make a simpler to construct deposit file style
+    const deposits = JSON.parse(await fs.readFile(argv.depositFile, "utf8"))
+    // TODO - make it simpler to construct deposit file style
```
bh2smith, line 24: I think this TODO should be removed if we aren't going to TODO it.

scripts/transfer_approve_deposit.js
```diff
@@ -18,8 +20,8 @@ module.exports = async (callback) => {
     const GnosisSafe = artifacts.require("GnosisSafe")
     const masterSafe = await GnosisSafe.at(argv.masterSafe)
 
-    const deposits = require(argv.depositFile)
-    // TODO - make a simpler to construct deposit file style
+    const deposits = JSON.parse(await fs.readFile(argv.depositFile, "utf8"))
```
bh2smith, line 23: I am struggling to understand why this change is necessary.

fedgiac: The difference is that `require` uses the path of the script as the base path, while `fs.readFile` uses the current working directory.
So, if your current working directory is `/path/to/dex-liquidity-provision/` and your deposit file is in `./examples/deposits.json`, then before this PR you'd have to run:
```
npx truffle exec scripts/transfer_approve_deposits.js [...] --depositFile="../examples/deposits.json"
```
(Note the two dots in the relative path.)
Now you can just use:

``` 
npx truffle exec scripts/transfer_approve_deposits.js [...] --depositFile="./examples/deposits.json"
```
(Note the single dots in the relative path.)
I noticed this bug when I tried to rely on the autocomplete to fill the path, which doesn't work because it autocompleted based on the current working directory.

scripts/transfer_approve_deposit.js
fedgiac: Fair enough, I don't see this happening any time soon.

scripts/transfer_approve_deposit.js
bh2smith: Thanks for clearing that up! I now understand.

PR #167, opened 2020-04-23
use placeOrder instead of placeOrdersValidFrom

### LINKED ISSUE #148: Make orders valid immediately, and not only in the future batch request
In order to debug and validate the deployment of the liquidity, one wants to see the repercussion on the order book immediately and not only in a future batch.

Hence, orders should be valid immediately

(E.g. due to this delay, we were not recognizing the error in the front-end between WBTC -USDC after the deployment immediately)

josojo:
closes #148 

testplan:
unit tests

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -221,7 +221,6 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     lowestLimit,
     highestLimit,
     debug = false,
-    validFrom = 3,
     expiry = maxU32 - 1 // We did not choose maxU32, in order to distinguish orders from "non-expiring" orders of the web interface
   ) {
     const log = debug ? (...a) => console.log(...a) : () => {}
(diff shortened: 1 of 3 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -265,30 +263,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
         log(
```
josojo, line 263: Initially, I thought it would be smarter to split the buyOrderTransactionCreation and the sellOrderTransactionCreation into two streams of Promises. But then we would no longer be able to write the buy and sell price for one bracket in one line.

Hence I first await Promises, which return [buyOrderPromise, sellOrderPromise]. Then I flat the array and await the inner promises again.

fedgiac: I like the approach, it looks good. I wonder what's the difference in gas price at the end. We might even end up saving some gas because we don't need to send the extra data to set the valid from. and I don't see a difference in the number of calls.

josojo:
@bh2smith 
placing orders were only sometimes rejected if the execution was taking too long since the configured validFrom < currentBatch at the time of mining. 

But with this PR, we don't specify validFrom at all, and then it can not fail.

The main drawback of the 3 batches was that sometimes signing takes longer and sometimes it is much quicker and then we want to check/see the liquidity immediately in the orderbooks.

scripts/utils/trading_strategy_helpers.js
josojo: Yeah, it should be more expensive. But most likely, it is negligible.

fedgiac: Maybe more problematic, it should also increase the size of the payload, not by much but we are already close to the limit. I wonder if we should add tests to check that gas cost and payload size stay below a certain limit?

PR #168, opened 2020-04-23
Withdraw without withdrawal file

fedgiac:
The command-line options of the withdraw script are changed:
- `--allTokens` is removed;
- `--from` is introduced as an alternative to `--withdrawalFile` and accepts a comma-separated list of bracket addresses. Tokens to be withdrawn must be specified manually with one of two new command-line options: `--tokens` or `--tokenIds`, which take a list of addresses or ids respectively for the tokens to withdraw. Amounts are computed automatically.

Note that using a deposit file is more efficient, especially if only one token is deposited at each address. If there are n brackets and m tokens, then using a deposit file would cost `~n` web3 requests, while using `--from` takes `~nm` requests since it needs to retrieve the balances of each pair (token, bracket). Anyway, if the amount to withdraw is zero the corresponding withdraw transaction is not created.

### DIFF
--- a/scripts/withdraw.js
+++ b/scripts/withdraw.js
@@ -22,14 +22,14 @@ const argv = require("./utils/default_yargs")
   })
   .option("tokens", {
     type: "string",
-    describe: "comma separated address list of tokens to withdraw, to use in combination with --from",
+    describe: "comma separated address list of tokens to withdraw, to use in combination with --brackets",
     coerce: (str) => {
       return str.split(",")
     },
   })
   .option("tokenIds", {
     type: "string",
-    describe: "comma separated list of exchange ids for the tokens to withdraw, to use in combination with --from",
+    describe: "comma separated list of exchange ids for the tokens to withdraw, to use in combination with --brackets",
     coerce: (str) => {
       return str.split(",")
     },
--- a/scripts/withdraw.js
+++ b/scripts/withdraw.js
@@ -5,18 +5,34 @@ const prepareWithdraw = require("./wrapper/withdraw")(web3, artifacts)
 const argv = require("./utils/default_yargs")
   .option("masterSafe", {
     type: "string",
-    describe: "Address of Gnosis Safe owning bracketSafes",
+    describe: "address of Gnosis Safe owning bracketSafes",
     demandOption: true,
   })
   .option("withdrawalFile", {
     type: "string",
     describe: "file name (and path) to the list of withdrawals",
-    demandOption: true,
   })
-  .option("allTokens", {
-    type: "boolean",
-    default: false,
-    describe: "ignore amounts from withdrawalFile and try to withdraw the maximum amount available for each bracket",
+  .option("from", {
+    type: "string",
+    describe:
+      "comma-separated list of brackets from which to withdraw the entire balance. Compatible with all valid combinations of --requestWithdraw, --withdraw, --transferFundsToMaster",
+    coerce: (str) => {
+      return str.split(",")
+    },
+  })
+  .option("tokens", {
+    type: "string",
+    describe: "comma separated address list of tokens to withdraw, to use in combination with --from",
+    coerce: (str) => {
+      return str.split(",")
+    },
+  })
+  .option("tokenIds", {
+    type: "string",
+    describe: "comma separated list of exchange ids for the tokens to withdraw, to use in combination with --from",
+    coerce: (str) => {
+      return str.split(",")
+    },
   })
   .option("requestWithdraw", {
     type: "boolean",
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -49,19 +49,27 @@ module.exports = function (web3, artifacts) {
     else if (argv.withdraw) {
       const currentBatchId = Math.floor(Date.now() / (5 * 60 * 1000)) // definition of BatchID, it avoids making a web3 request for each withdrawal to get BatchID
       const pendingWithdrawal = await exchange.getPendingWithdraw(bracketAddress, tokenData.address)
-      if (pendingWithdrawal[1].toNumber() == 0) {
-        log("Warning: no withdrawal was requested for address", bracketAddress, "and token", tokenData.symbol)
-        amount = "0"
-      }
-      if (amount != "0" && pendingWithdrawal[1].toNumber() >= currentBatchId) {
-        log("Warning: amount cannot be withdrawn from the exchange right now, withdrawing zero")
-        amount = "0"
-      }
       amount = pendingWithdrawal[0].toString()
+      if (pendingWithdrawal[1].toNumber() >= currentBatchId) {
+        const batchIdsLeft = pendingWithdrawal[1].toNumber() - currentBatchId + 1
+        log(
+          "Warning: requested withdrawal of " +
+            amount +
+            " " +
+            tokenData.symbol +
+            " for bracket " +
+            bracketAddress +
+            " cannot be executed until " +
+            batchIdsLeft +
+            " " +
+            (batchIdsLeft == 1 ? "batch" : "batches") +
+            " from now, skipping"
+        )
+        amount = "0"
+      }
     } else {
       amount = (await token.balanceOf(bracketAddress)).toString()
     }
-    if (amount == "0") log("Warning: address", bracketAddress, "has no balance to withdraw for token", tokenData.symbol)
     return amount
   }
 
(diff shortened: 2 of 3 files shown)


scripts/withdraw.js
```diff
@@ -5,18 +5,34 @@ const prepareWithdraw = require("./wrapper/withdraw")(web3, artifacts)
...
-  .option("allTokens", {
-    type: "boolean",
-    default: false,
-    describe: "ignore amounts from withdrawalFile and try to withdraw the maximum amount available for each bracket",
+  .option("from", {
+    type: "string",
+    describe:
+      "comma-separated list of brackets from which to withdraw the entire balance. Compatible with all valid combinations of --requestWithdraw, --withdraw, --transferFundsToMaster",
+    coerce: (str) => {
+      return str.split(",")
```
bh2smith, line 20: Glad to see some pythonic looking syntax here!

scripts/wrapper/withdraw.js
```diff
@@ -12,34 +13,68 @@ module.exports = function (web3, artifacts) {
...
+            " " +
+            tokenData.symbol +
+            " for bracket " +
+            bracketAddress +
+            " cannot be executed until " +
+            batchIdsLeft +
+            " " +
+            (batchIdsLeft == 1 ? "batch" : "batches") +
+            " from now, skipping"
+        )
```
bh2smith, line 72: I have to say that this looks pretty horrendous. I think you might be able to get away with some cleaner string formatting here. Are you familiar with the backtick string formatting?

```
`Warning: requested withdrawal of ${amount} ${tokenData.symbol} ...etc
```

Should work here...

fedgiac: Nice idea, it's definitely more readable.

bh2smith: Thanks so very much! I really like the way it looks now.

scripts/withdraw.js
```diff
@@ -5,18 +5,34 @@ const prepareWithdraw = require("./wrapper/withdraw")(web3, artifacts)
...
   .option("withdrawalFile", {
     type: "string",
     describe: "file name (and path) to the list of withdrawals",
-    demandOption: true,
   })
-  .option("allTokens", {
-    type: "boolean",
-    default: false,
-    describe: "ignore amounts from withdrawalFile and try to withdraw the maximum amount available for each bracket",
+  .option("from", {
```
josojo, line 15: I would consider calling this option "brackets", as we used this syntax for the deposits and the complete-liquidity-provision script

scripts/wrapper/withdraw.js
```diff
@@ -48,25 +77,50 @@ module.exports = function (web3, artifacts) {
...
-    const tokenInfoPromises = fetchTokenInfoForFlux(withdrawals)
-    const exchange = await getExchange(web3)
+    let withdrawals
+    let tokenInfoPromises
+    if (argv.withdrawalFile) {
+      withdrawals = JSON.parse(await fs.readFile(argv.withdrawalFile, "utf8"))
+      tokenInfoPromises = fetchTokenInfoForFlux(withdrawals)
+    } else {
+      const exchangePromise = getExchange(web3)
+      if (argv.tokens) tokenInfoPromises = fetchTokenInfoAtAddresses(argv.tokens)
```
josojo, line 87: Would the following be more succinct?
```
const tokenAddresses = argv.tokens ||  await Promise.all(
          argv.tokenIds.map(async (id) => (await exchangePromise).tokenIdToAddressMap(id))
        )
tokenInfoPromises = fetchTokenInfoAtAddresses(tokenAddresses)
```

scripts/wrapper/withdraw.js
```diff
@@ -48,25 +77,50 @@ module.exports = function (web3, artifacts) {
...
-            argv,
-            withdrawal.bracketAddress,
-            await tokenInfoPromises[withdrawal.tokenAddress],
-            await exchange
-          ),
-        }))
-      )
+      withdrawals = []
+      const candidateWithdrawalPromises = []
+      for (const [, tokenDataPromise] of Object.entries(tokenInfoPromises))
```
josojo, line 101: I think the script would be easier to understand if you would await all tokenInfoPromises before going over them in this loop. Then you don't get promises of promises, and the performance lost would be negligible, right?

If I am not mistaken, then you would not even need
```
 for (const candidateWithdrawalPromise of candidateWithdrawalPromises)
```
then? But I am not sure right now.

test/scripts/withdraw.js
```diff
@@ -99,115 +99,353 @@ contract("Withdraw script", function (accounts) {
...
+      const depositFile = await tmp.file()
+      await fs.writeFile(depositFile.path, JSON.stringify(deposits))
 
-    for (const { amount, tokenAddress, bracketAddress } of deposits) {
-      const requestedWithdrawal = (await exchange.getPendingWithdraw(bracketAddress, tokenAddress))[0].toString()
-      const bracketBalance = (await token.balanceOf(bracketAddress)).toString()
-      assert.equal(requestedWithdrawal, "0", "A withdrawal request is still pending")
-      assert.equal(bracketBalance, amount, "Bad amount requested to withdraw")
-    }
+      const argv1 = {
```
josojo, line 136: like this clustering...

scripts/wrapper/withdraw.js
```diff
@@ -12,34 +13,62 @@ module.exports = function (web3, artifacts) {
...
+    printOutput = false
+  ) {
     const log = printOutput ? (...a) => console.log(...a) : () => {}
     let amount
-    const token = tokenInfo.instance
-    if (argv.requestWithdraw) amount = (await exchange.getBalance(bracketAddress, tokenInfo.address)).toString()
+    const token = tokenData.instance
+    if (argv.requestWithdraw) amount = (await exchange.getBalance(bracketAddress, tokenData.address)).toString()
     else if (argv.withdraw) {
-      const currentBatchId = Math.floor(Date.now() / (5 * 60 * 1000)) // definition of BatchID, it avoids making a web3 request for each withdrawal to get BatchID
```
bh2smith, line 28: Although this is now gone, I wanted to point out that this appears to be the wrong calculation for batch ID. Shouldn't it be ` Math.floor(Date.now() / 300)` (i.e. without multiplication by 1000)

scripts/wrapper/withdraw.js
```diff
@@ -48,25 +77,50 @@ module.exports = function (web3, artifacts) {
...
+    } else {
+      const exchangePromise = getExchange(web3)
+      if (argv.tokens) tokenInfoPromises = fetchTokenInfoAtAddresses(argv.tokens)
+      else {
+        const tokenAddresses = await Promise.all(
+          argv.tokenIds.map(async (id) => (await exchangePromise).tokenIdToAddressMap(id))
+        )
+        tokenInfoPromises = fetchTokenInfoAtAddresses(tokenAddresses)
+      }
+      const exchange = await exchangePromise
```
bh2smith, line 94: I don't understand the purpose of creating the promise if it is only going to be awaited a couple of lines below. This seems like an unnecessary line of code when we can simple write `const exchange = await getExchange(web3)` back at line 86.

test/scripts/withdraw.js
```diff
@@ -99,115 +99,353 @@ contract("Withdraw script", function (accounts) {
...
+      }
+      const transaction1 = await prepareWithdraw(argv1)
+      await execTransaction(masterSafe, lw, transaction1)
+      await waitForNSeconds(301)
+
+      const argv2 = {
+        masterSafe: masterSafe.address,
+        withdrawalFile: depositFile.path,
+        withdraw: true,
+      }
```
bh2smith, line 149: It appears to be that `argv1 == argv2`. Was this intended?

test/scripts/withdraw.js
```diff
@@ -99,115 +99,353 @@ contract("Withdraw script", function (accounts) {
...
```
bh2smith, line 184: Here they are again (equal).

scripts/wrapper/withdraw.js
fedgiac: Nice, will change!

scripts/wrapper/withdraw.js
fedgiac: I tried to make the code more linear with the new changes. I still couldn't avoid creating some intermediate promises to populate the withdrawal list, but there are less nested calls so it should be more readable.

scripts/wrapper/withdraw.js
fedgiac: I'd argue in favor of the current code: it reduces the running time by one network request assuming that `argv.tokens` is given by the user.
This is because (if `argv.token` is set) the fetching of the exchange is executed in parallel to the fetching of token information, while if you await the exchange in line 86 then the fetching of token information only starts immediately after the exchange has been retrieved.

bh2smith: Ok then. If this is the case, then I am in favour of this. Was not aware this how it worked! Cool stuff.

PR #175, opened 2020-04-24
Rework package.json

fedgiac:
Clean up of `package.json`.
Dependencies have been correctly sorted into dev and non-dev.
Unused package `run-with-testrpc` has been removed.

A nice addition that could also be included in dex-contracts is `engines`: if you try to run `yarn install` with for example Node version 11 the command fails telling the user the Node version is incompatible. I remember my first experience with `dex-contracts` was to debug a mysteriously failing installation due to the fact that the Node version I used was 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,11 @@
 {
   "name": "liquidity-provision-scripts",
   "description": "Scripts to deploy and manage automated traders on the Gnosis Protocol exchange",
+  "keywords": [
+    "Gnosis Protocol",
+    "liquidity",
+    "trading"
+  ],
   "devDependencies": {
     "@gnosis.pm/mock-contract": "^3.0.0",
     "eslint": "^6.8.0",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
     "tmp-promise": "^2.0.2",
     "truffle-plugin-verify": "^0.3.10"
   },
+  "engines": {
+    "node": "^10.0.0"
+  },
   "scripts": {
     "lint": "eslint .",
     "prettier": "prettier --write './**/*.js'",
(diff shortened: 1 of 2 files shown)


bh2smith:
> if you try to run yarn install with for example Node version 11

I really like the sound of this!

package.json
```diff
@@ -1,19 +1,32 @@
 {
   "name": "liquidity-provision-scripts",
-  "version": "0.1.0",
-  "description": "",
+  "description": "Scripts to deploy and manage automated traders on the Gnosis Protocol exchange",
+  "private": true,
```
bh2smith, line 4: What is this "private" for?

package.json
```diff
@@ -1,19 +1,32 @@
...
     "@gnosis.pm/mock-contract": "^3.0.0",
     "eslint": "^6.8.0",
     "eslint-plugin-no-only-tests": "^2.4.0",
     "eslint-plugin-react": "^7.19.0",
-    "eth-lightwallet": "^4.0.0",
-    "ethereumjs-abi": "^0.6.5",
     "ganache-cli": "6.9.1",
     "prettier": "^2.0.5",
-    "run-with-testrpc": "^0.3.0",
+    "sol-merger": "^2.0.1",
```
bh2smith, line 24: when was this introduced and what for?

package.json
fedgiac: It says that the package is not published on NPM. When we finally publish the package we'll remove this line.

package.json
fedgiac: It's used to produce an "Etherscan-friendly" version of a contract's source code by collecting all contracts involved in the compilation of a contract into a single file. It's only required for running `yarn flatten`.

Issue #120, opened 2020-04-01
[Easy] Duplicate log from function composition


bh2smith:
the log statement `"Fetching token data from EVM"` occurs in both `fetchTokenInfoFromExchange` and `fetchTokenInfoAtAddresses` while the former calls the later. This means (when debug == true) that we will observe identical log statements twice in a row.

Furthermore, although I am not entirely sure, but it appears this line

```js
const log = debug ? () => console.log.apply(arguments) : () => {}
```

could be moved out of individual functions as a global statement for the definition of `log`

fedgiac:
Currently logging has many problems, for example the line you mention doesn't appear to be working, see issue #74.This issue has been neglected in the hope of eventually reworking how logging works in the code. This is especially needed since we started to rely on asynchronous calls in the code: the order in which the logging information is basically random, so you would see e.g. the orders of Safe 12 before that of Safe 1.

Issue #74, opened 2020-03-13
Logging not working in fetchTokenInfo


fedgiac:
The output is not logged correctly to the console. In general, we could define the function `log` globally instead of once for each function.

bh2smith:
closed by #177

bh2smith:
@fedgiac - just so you know, these issues would be closed automatically when your PRs are merged if you explicitly write "closes #XXX" in the PR description.

PR #177, opened 2020-04-28
Fix logging when fetching tokens

fedgiac:
Partially solves issues #74 and #120, further improvements require a different logging strategy. The logs are now printed if the flag `debug` is set to true, there are no repeating log lines, and the fetching message only appears if token info is actually being fetched from the EVM instead of retrieved locally.

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -120,10 +120,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     const log = debug ? (...a) => console.log(...a) : () => {}
     const ERC20 = artifacts.require("ERC20Detailed")
 
-    log("Fetching token data from EVM")
+    let requiresFetching = false
     const tokenPromises = {}
     for (const tokenAddress of tokenAddresses) {
       if (!(tokenAddress in globalTokenPromisesFromAddress)) {
+        requiresFetching = true
         globalTokenPromisesFromAddress[tokenAddress] = (async () => {
           const tokenInstance = await ERC20.at(tokenAddress)
           const [tokenSymbol, tokenDecimals] = await Promise.all([tokenInstance.symbol.call(), tokenInstance.decimals.call()])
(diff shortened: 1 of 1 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -151,20 +153,25 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
     for (const id of tokenIds) {
       if (!(id in globalTokenPromisesFromId)) {
+        requiresFetching = true
         globalTokenPromisesFromId[id] = (async () => {
           const tokenAddress = await exchange.tokenIdToAddressMap(id)
-          log(`Token id ${id} corresponds to token at address ${tokenAddress}`)
-          return fetchTokenInfoAtAddresses([tokenAddress], debug)[tokenAddress]
+          const tokenInfo = await fetchTokenInfoAtAddresses([tokenAddress], false)[tokenAddress]
+          log(
+            `Found token ${tokenInfo.symbol} for exchange id ${id} at address ${tokenInfo.address} with ${tokenInfo.decimals} decimals`
```
bh2smith, line 167: I see you are really enjoying this type of string formatting!

fedgiac: Now that I'm aware it exists, I'll never use anything else if I have to print a variable!

fedgiac:
I also don't really like the repetition, but I don't see how to fix it without adding complexity.
A question is also: if I call `exchange.tokenIdToAddressMap(id)` but the token data for that address is already cached, do I want to log "Fetching token data from EVM"?
If yes, the only solution I see is an extra variable `printFetchingLog` in `fetchTokenInfoAtAddresses`, which seems artificial to me.
If no, I still have to change the logic of `fetchTokenInfoFromExchange` to make it so that it calls `fetchTokenInfoAtAddresses` only once, otherwise `fetchTokenInfoFromExchange` could print "Fetching token data from EVM" multiple times for a single execution.
Considering that this is a stopgap PR pending a better logging infrastructure, I'd leave it as it is.

PR #179, opened 2020-04-29
Fixed gas limit

josojo:
Closes #154 

This PR fixes in practice the gas limit to a value around 6m.
But we can not set the value straight forward, as Metamask still adds a gas-buffer.

testplan:

```
export PK...
source config/rinkeby.env

npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=2 --lowestLimit=190 --highestLimit=215 --currentPrice=207 --masterSafe=$MASTER_SAFE --investmentTargetToken=0 --investmentStableToken=0 --fleetSize=<Random even. number < 20 > --network=rinkeby
```
and check in meta mask the proposed gas, which should be around 6 million.

PR #180, opened 2020-04-29
Calculating the nonce

josojo:
This Pr calculates the nonce for the script: complete_liquidity_provision.

This allows us to smoothen the user workflow, as both tx are sent to the gnosis-interface simultaneously.

testplan:
try out the new feature like that:
```
npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=2 --lowestLimit=190 --highestLimit=215 --currentPrice=187 --masterSafe=$MASTER_SAFE --investmentTargetToken=0 --investmentStableToken=0 --fleetSize=20 --network=rinkeby
```
and make sure that you can sign both transactions in our rinkeby safe

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -9,7 +9,7 @@ const {
   checkSufficiencyOfBalance,
 } = require("./utils/trading_strategy_helpers")(web3, artifacts)
 const { isPriceReasonable, areBoundsReasonable } = require("./utils/price_utils")(web3, artifacts)
-const { signAndSend, promptUser } = require("./utils/sign_and_send")(web3, artifacts)
+const { signAndSend } = require("./utils/sign_and_send")(web3, artifacts)
 const { proceedAnyways } = require("./utils/user_interface_helpers")(web3, artifacts)
 const { toErc20Units } = require("./utils/printing_tools")
 const { sleep } = require("./utils/js_helpers")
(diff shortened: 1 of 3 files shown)


scripts/utils/sign_and_send.js
```diff
@@ -64,7 +66,6 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
 
     const interfaceLink = `https://${linkPrefix[network]}gnosis-safe.io/app/#/safes/${masterSafe.address}/transactions`
     console.log("Transaction awaiting execution in the interface", interfaceLink)
-    console.log("Remember to increase the gas limit!")
```
fedgiac, line 67: This change should not appear here but only in PR #179, right?

scripts/complete_liquidity_provision.js
```diff
@@ -136,16 +136,12 @@ module.exports = async (callback) => {
       true
     )
 
-    console.log("4. Sending out orders")
-    await signAndSend(masterSafe, orderTransaction, argv.network)
+    console.log("4. Sending the order placing transaction to gnosis-safe interface, please sign this transaction first")
```
fedgiac, line 139: I propose to change "sign" to "execute" here, since I think that it's actually fine to sign them in any order.

scripts/complete_liquidity_provision.js
```diff
@@ -136,16 +136,12 @@ module.exports = async (callback) => {
...
+    await signAndSend(masterSafe, orderTransaction, argv.network, nonce)
 
-    console.log("5. Sending out funds")
-    const answer = await promptUser(
-      "Are you sure you that the order placement was correct, did you check the telegram bot? [yN] "
-    )
-    if (answer == "y" || answer.toLowerCase() == "yes") {
-      await signAndSend(masterSafe, bundledFundingTransaction, argv.network)
-    }
+    console.log("5. Sending the funds transferring transaction, please sign this transaction second")
```
fedgiac, line 143: Same comment as before, propose change "sign" to "execute".

scripts/utils/sign_and_send.js
```diff
@@ -24,8 +24,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @param {Address} masterAddress Address of the master safe owning the brackets
    * @param {Transaction} transaction The transaction to be signed and sent
    */
-  const signAndSend = async function (masterSafe, transaction, network) {
-    const nonce = await masterSafe.nonce()
+  const signAndSend = async function (masterSafe, transaction, network, nonce = null) {
+    if (nonce === null) {
+      nonce = (await masterSafe.nonce()).toNumber()
+    }
```
bh2smith, line 30: If we are already doing this here, why is it being given as a parameter in the function call? is there any reason why we might want to send our own nonce?

josojo: >we might want to send our own nonce
Yes, in this PR we want to send the second transaction to the interface before the first one was executed. Hence,
`(await masterSafe.nonce()).toNumber()` would give us the wrong nonce for the second transaction. That is why I am calculating the value.

josojo:
> It looks like the arguments being passed into the script are becoming quite a few. Do we think maybe its a good time to move all this configuration material into a YAML file and somehow make the scripts parse these instead?

Federico and I had the idea to refactor the code and make the functions operate on classes. This would avoid many of the parameters and is probably cleaner than yaml.
But... right now, we don't have time for that...

PR #187, opened 2020-04-30
Fix - longer waiting time, before creating the orders

josojo:
Part of the learning of the incident.
Because we were not waiting long enough, the scripts were failing and I was hardcoding the brackets manually.

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -109,9 +109,9 @@ module.exports = async (callback) => {
     const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
     console.log("List of bracket traders in one line:", bracketAddresses.join())
 
-    // Sleeping for 1 seconds to make sure Infura nodes have processed all newly deployed contracts so that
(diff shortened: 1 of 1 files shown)


scripts/complete_liquidity_provision.js
```diff
@@ -109,9 +109,9 @@ module.exports = async (callback) => {
     const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
     console.log("List of bracket traders in one line:", bracketAddresses.join())
 
-    // Sleeping for 1 seconds to make sure Infura nodes have processed all newly deployed contracts so that
+    // Sleeping for 3 seconds to make sure Infura nodes have processed all newly deployed contracts so that
     // they can be awaited.
-    await sleep(1000)
+    await sleep(3000)
```
bh2smith, line 114: Does this work for sure? How did we deduce this number?

josojo: Empirically,

No, there is no guarantee. A better fix would be to talk to only one node, but then the script does no longer become executable for everyone, as not everyone has access to a single node. Infura is just the industry standard.

Issue #191, opened 2020-04-30
Make complete liquidity provision compatible with hardcoded addresses


fedgiac:
It's natural to hardcode addresses in the complete liquidity provision scripts, but it creates extra risks that were not considered when creating the script. Some ideas:
- add `--brackets` to the command-line arguments, which makes the script skip deployment
- add extra checks, ideally only when `--brackets` is given

fleupold:
I was going to look into this task but have a question:

It seems like we already have a `bracket_orders.js` script that allows to pass existing `--bracket` via the command line.

Isn't this sufficient? Do we also want to have this functionality for `buildTransferApproveDepositFromOrders`? 

Or is this script no longer used and we want to change `complete_liquidity_provision.js` to skip the first step when provided with existing brackets?

josojo:
Yes, you are right Felix, one could always run 
```
placeOrders
--> create a custom made deposit file
buildTransferApproveDepositFromOrders
```
once the brackets are available, instead of the 'complete_liquidity_provision' with the brackets. But in practice, one avoids these 3 time-consuming steps, if one could do it with one step, just to save some time. Its really just a convenience feature, which would have prevented the error from last thursday

fleupold:
So is there a value in having the `bracket_orders.js` and `transfer_approce_deposit.js` scripts by themselves or can they be removed?

Also, how does one create the deposit file you are referring to?

In case we want to use the complete script for this, I assume this tasks captures adding another optional `--bracket` command line option to the complete script and skip fleet deployment if it is specified.
The task description mentions to do some "checks" in that case. Are these the priceCheck and boundCheck that are currently performed in `bracket_orders` script?

fedgiac:
> So is there a value in having the bracket_orders.js and transfer_approce_deposit.js scripts by themselves or can they be removed?

Currently there is none, it has been a long time since they have been used. In the future they could have their own purpose however: at the moment `complete_liquidity_provision.js` is intended to be used for the first deployment of the safes, so `bracket_orders.js` and `transfer_approve_deposit.js` could be used afterwards. `transfer_approve_deposit.js` could also be extended to have a more fine-grained control on how deposits work, for example setting an unlimited allowance, depositing to the brackets but not to the exchange, and so on.

> Also, how does one create the deposit file you are referring to?

[...]

PR #196, opened 2020-04-30
Verifying that no orders have been placed on the brackets beforehand

josojo:
This code would have caught my error and would have prevented false order placement.

testplan: unit tests only

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -24,7 +24,7 @@ The following scripts will use this account to propose transactions to the inter
 Setup the following env variables for the deployment process:
 
 ```
-export PK=<your private key for the proposer account>
+ export PK=<your private key for the proposer account>
 export GAS_PRICE_GWEI=<look up the suggestion from ethgasstation.info>
 export NETWORK_NAME=<network>
 export MASTER_SAFE=<master safe>
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -41,7 +41,7 @@ It will send one ethereum transaction and send two transactions request to the g
 The ethereum transasction will create the bracket-traders. For this transaction the provided private key will be used to pay for the gas.
 The first request of the script will generate orders on behalf of the bracket-traders.
 Please sign this transaction in the gnosis-safe interface and double check that the prices of the orders are as intended - for example in [telegram-mainnet](https://t.me/gnosis_protocol) or [telegram-rinkeby](https://t.me/gnosis_protocol_dev) channels.
-Then continue the script and send out the second request, which will fund the bracket-traders' accounts on the exchange. Notice: One has to wait until the transaction is mined, as otherwise the nonce for the next request will be incorrect.
+The second request generates a transaction funding the bracket-traders' accounts on the exchange.
 Making the requests to the gnosis-interfaces does not cost any gas. However, signing and executing the transactions in the gnosis-safe interface will incur gas costs.
 
 Here is an example script invocation:
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -132,7 +132,7 @@ module.exports = async (callback) => {
       // Detect if provided brackets have existing orders.
       const existingOrders = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
-          return await hasExistingOrders(safeAddr, exchange)
+          return hasExistingOrders(safeAddr, exchange)
         })
       )
       const dirtyBrackets = bracketAddresses.filter((_, i) => existingOrders[i] == true)
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -8,6 +8,7 @@ const {
   buildOrders,
   checkSufficiencyOfBalance,
   isOnlySafeOwner,
+  hasExistingOrders,
 } = require("./utils/trading_strategy_helpers")(web3, artifacts)
 const { isPriceReasonable, areBoundsReasonable } = require("./utils/price_utils")(web3, artifacts)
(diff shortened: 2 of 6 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -109,6 +109,18 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+   * Checks that a bracket has not yet made any orders
+   */
+  const checkNoOrdersPlaced = async function (bracket, exchange) {
+    const orders = await exchange.getEncodedUserOrders.call(bracket)
+    if (orders === null) {
+      return true
+    } else {
+      return false
+    }
+  }
```
bh2smith, line 122: Wouldn't this be easier if we just check the `orderIndex` of this user?

josojo: What kind of orderIndex? Are you talking about the length of the orders vector? In this case, I can not get a length of an null object.

fedgiac:
This PR would make it impossible to create anything but the initial orders of a bracket. I think we should verify the presence of orders outside this helper function but rather in some other scripts, like when implementing a solution to issue #191.

scripts/utils/trading_strategy_helpers.js
bh2smith: I see. I guess the contract does not expose a public vie like `getOrderIndex(address)`. I wonder if this might be a good thing to include in v2.

josojo:
> This PR would make it impossible to create anything but the initial orders of a bracket. I think we should verify the presence of orders outside this helper function but rather in some other scripts, like when implementing a solution to issue #191.

yeah, I am okay with that. Then I will change this PR to tackle #191

scripts/utils/trading_strategy_helpers.js
bh2smith: I think maybe we could/should make this function more sophisticated by checking that there are no "valid" orders. At least this way, we could reuse brackets instead of abandoning them.

bh2smith:
Is this expected to be merged soon? I feel like I have been requested to use this verification in #210

bh2smith:
I have commandeered this PR and made the discussed changes related to #210 which has already been merged. This now incorporates the `hasExistingOrders` check in the usage of complete liquidity provision. Please be aware that some additional changes were also made here that are directly related to usability issues encountered while testing this.

In brief:

- include assertion that env var `PK` must be explicitly available to run the `sign_and_send`. I failed to execute my test run on rinkeby where I would usually assume truffle to use a default account (but this is no longer possible "for safety")

- allow script user to "proceed anyways" when brackets admit existing orders.
- include a space in the private key export on the readme file. This is to help prevent any unknowing users from exposing their private key in their bash history.
- Rewrote some warning messages regarding order of transaction execution because the statements as they were were not true. This is because the nonce is explicitly stated in transaction submission so there is only one order in which they can be executed.


This PR is now ready for re-review @josojo, @fedgiac and possibly @fleupold.

scripts/README.md
```diff
@@ -24,7 +24,7 @@ The following scripts will use this account to propose transactions to the inter
 Setup the following env variables for the deployment process:
 
 ```
-export PK=<your private key for the proposer account>
+ export PK=<your private key for the proposer account>
```
bh2smith, line 27:
```suggestion
 export PK=<private key of proposer account>  \\ Note that the additional space at the beginning excludes this command from your bash history.
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -249,8 +261,6 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
 
     const buyAndSellOrderPromises = await Promise.all(
       bracketAddresses.map(async (bracketAddress, bracketIndex) => {
-        assert(await isOnlySafeOwner(masterAddress, bracketAddress), "each bracket should be owned only by the master Safe")
```
bh2smith, line 252: This check is done by the script itself and shouldn't be performed twice.

scripts/complete_liquidity_provision.js
```diff
@@ -128,6 +129,19 @@ module.exports = async (callback) => {
           }
         })
       )
+      // Detect if provided brackets have existing orders.
+      const existingOrders = await Promise.all(
+        bracketAddresses.map(async (safeAddr) => {
+          return await hasExistingOrders(safeAddr, exchange)
```
fedgiac, line 135: Nitpick: no await needed here.
```suggestion
          return hasExistingOrders(safeAddr, exchange)
```

scripts/complete_liquidity_provision.js
```diff
@@ -128,6 +129,19 @@ module.exports = async (callback) => {
...
+        })
+      )
+      const dirtyBrackets = bracketAddresses.filter((_, i) => existingOrders[i] == true)
+      if (
+        existingOrders.some((t) => t == true) &&
```
fedgiac, line 140: Nice syntax! You could even go further and use the identity function:
```suggestion
        existingOrders.some((t) => t) &&
```

bh2smith: I don't understand this suggestion... I am assuming that the result of the mapping needs to be a boolean statement? Also, what is the identity function if it isn't `{t => t}`?

fedgiac: My understanding is that ` existingOrders.some(f)` returns `true` if `f(x)` is true for at least one `x` in `existingOrders`. In this case `f(x) = (x == true)`, but then you could just write `f(x) = x` to get back `true`.

bh2smith:
>  I don't know if it makes sense to decide who'll merge first.

No need to worry about that...

PR #200, opened 2020-05-01
Individuell proposer account

### LINKED ISSUE #158: different proposer accounts
Introduce customized proposer accounts. In this way there would be less risk of our users shooting themselves in the foot.

---
not sure whether it is worth the effort. Investigate effort

josojo:
closes #158 

I eliminated the standardized proposer account. 

I think in this PR, we could eliminate the lightwallet integration completely, but then I think in the future we want to have keystore file system for sure. 
I am not yet sure, which keystore filesystem we should use, since the one from the gnosis-safe team is no longer maintained:
https://github.com/ConsenSys/eth-lightwallet/blob/master/lib/signing.js

Hence, I have decided to keep the lightwallet in there, in case we switch back in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,17 +14,6 @@
   "bugs": {
     "url": "https://github.com/gnosis/dex-liquidity-provision/issues"
   },
-  "devDependencies": {
-    "@gnosis.pm/mock-contract": "^3.0.0",
-    "eslint": "^6.8.0",
-    "eslint-plugin-no-only-tests": "^2.4.0",
-    "eslint-plugin-react": "^7.19.0",
-    "ganache-cli": "6.9.1",
-    "prettier": "^2.0.5",
-    "sol-merger": "^2.0.1",
-    "tmp-promise": "^2.0.2",
-    "truffle-plugin-verify": "^0.3.10"
-  },
   "engines": {
     "node": "^10.0.0"
   },
@@ -40,13 +29,25 @@
     "networks-inject": "CONF_FILE=$(pwd)'/migration_conf.js'  node node_modules/@gnosis.pm/util-contracts/src/inject_network_info.js",
     "networks-extract": "CONF_FILE=$(pwd)'/migration_conf.js'  node node_modules/@gnosis.pm/util-contracts/src/extract_network_info.js"
   },
+  "devDependencies": {
+    "@gnosis.pm/mock-contract": "^3.0.0",
+    "eslint": "^6.8.0",
+    "eslint-plugin-no-only-tests": "^2.4.0",
+    "eslint-plugin-react": "^7.19.0",
+    "ganache-cli": "6.9.1",
+    "prettier": "^2.0.5",
+    "sol-merger": "^2.0.1",
+    "tmp-promise": "^2.1.1",
+    "truffle-plugin-verify": "^0.3.10"
+  },
   "dependencies": {
     "@gnosis.pm/dex-contracts": "^0.2.13",
     "@gnosis.pm/owl-token": "^3.1.0",
     "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.1",
-    "@gnosis.pm/solidity-data-structures": "^1.2.2",
+    "@gnosis.pm/solidity-data-structures": "=1.2.4",
     "@gnosis.pm/util-contracts": "^2.0.6",
-    "@truffle/contract": "^4.2.2",
+    "@truffle/contract": "4.1.15",
+    "@openzeppelin/contracts": "=2.5.0",
     "axios": "^0.19.2",
     "bn.js": "^5.1.1",
     "canonical-weth": "^1.4.0",
@@ -55,8 +56,7 @@
     "eth-lightwallet": "^4.0.0",
     "ethereumjs-util": "^6.2.0",
     "node-fetch": "^2.6.0",
-    "openzeppelin-solidity": "^2.5.1",
-    "synthetix-js": "^2.21.12",
+    "synthetix-js": "=2.21.12",
     "truffle": "^5.1.24",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
   },
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "node-fetch": "^2.6.0",
     "openzeppelin-solidity": "^2.5.1",
     "synthetix-js": "^2.21.12",
-    "truffle": "^5.1.23",
+    "truffle": "^5.1.24",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
   },
   "resolutions": {
--- a/package.json
+++ b/package.json
@@ -53,14 +53,10 @@
     "canonical-weth": "^1.4.0",
     "dotenv": "^8.0.0",
     "eth-gas-reporter": "^0.2.17",
-    "eth-lightwallet": "^4.0.0",
     "ethereumjs-util": "^6.2.0",
     "node-fetch": "^2.6.0",
     "synthetix-js": "=2.21.12",
     "truffle": "^5.1.24",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
-  },
-  "resolutions": {
-    "bitcore-lib": "8.1.1"
   }
 }
--- a/scripts/utils/internals.js
+++ b/scripts/utils/internals.js
@@ -81,10 +81,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   }
 
   function signHashWithPrivateKey(hash, privateKey) {
-    const Util = require("ethereumjs-util")
-
-    const msgBuff = new Buffer(Util.stripHexPrefix(hash), "hex")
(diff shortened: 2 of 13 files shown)


scripts/utils/internals.js
```diff
@@ -100,6 +100,14 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     }
   }
 
+  function signHashWithPrivateKey(hash, privateKey) {
+    const Util = require("ethereumjs-util")
+
+    const msgBuff = new Buffer(Util.stripHexPrefix(hash), "hex")
+    const sig = Util.ecsign(msgBuff, new Buffer(privateKey, "hex"))
+    return "0x" + sig.r.toString("hex") + sig.s.toString("hex") + sig.v.toString(16)
```
fedgiac, line 108: Is the padding right here? Unless `sig.r`, `sig.s`, `sig.v`, have an ad hoc implementation of `toString` that takes into account padding, I believe the returned hash would be incorrect if any of these parameters started with a zero byte.

scripts/utils/sign_and_send.js
```diff
@@ -41,10 +41,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       ADDRESS_0,
       nonce
     )
-    const lightWallet = await createLightwallet()
-    const account = lightWallet.accounts[0]
-    console.log(`Signing and posting multi-send transaction request from proposer account ${account}`)
-    const sigs = signTransaction(lightWallet, [account], transactionHash)
+    const privateKey = process.env.PK
+    const account = web3.eth.accounts.privateKeyToAccount("0x" + privateKey)
```
fedgiac, line 45: Implementation question: does this line change the output of `web3.eth.getAccounts()`?

bh2smith: Also seems hackey to be doing string manipulation in the vicinity of a private key...

bh2smith:
> we could eliminate the lightwallet integration completely

I would be very much in favour of this. It is a non-standard library to my understanding anyway and is probably also the reason for that silly `bitcore-lib` issue we have been having.

josojo:
> I'd rather pick a different package to compute transaction hashes instead of manipulating private keys directly

I don't manipulate private keys. I can see that one would like to argue for a vault, but I feel like I don't make anything worse, as we have the PK anyways already exported as env. I think the vault integration that does not depend on unmaintained repos should happen in another PR.

>In any case I thought the main motivation of this PR was to remove eth-lightwallet because of this comment.

No. I think that it is not acceptable that anyone can propose transactions to our safe. Once the transaction is in the interface, there is a much higher risk that it gets signed accidentally.
With this PR, external parties can no longer propose solutions to our Gnosis safe. 

> Also, would the complete liquidity provision script work for us with the current setup? 

Yes, we need to modify the setup for this of the current Master_safe, but it's worth it, for sure.

scripts/utils/internals.js
josojo: We are code already in the function:
```
  function signTransaction(lw, signers, transactionHash) 
```

scripts/utils/sign_and_send.js
josojo: According to the documentation, is only creating an account object, which I am storing in the account variable:
https://web3js.readthedocs.io/en/v1.2.0/web3-eth-accounts.html#privatekeytoaccount

josojo: @bh2smith  How would you propose to do it better?

scripts/utils/internals.js
josojo:
```
const buf1 = Buffer.allocUnsafe(26);

buf1[0] = 0 ;
for (let i = 20; i < 26; i++) {
  // 97 is the decimal ASCII value for 'a'.
  buf1[i] = i ;
}
console.log(buf1);
console.log(buf1.toString('hex'));
```
use this small script to convince yourself that buffers do not cut of paddings

fedgiac: I didn't notice we already use this code in another function in our code. This is indeed copied from the Gnosis Safe code, so even if the padding was incorrect this would be a "feature" rather than a bug. It would be nicer if we could import all of this from their repo instead, but after a brief look it doesn't look like this is possible.
(By the way, an implicit assumption of your example is that the two buffers always have the same length for different signatures, which might not be the case. Also, the last output is a JS number, so this is intentionally not padded, which to me is a bit surprising.)

scripts/utils/internals.js
```diff
@@ -100,6 +100,14 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     }
   }
 
+  function signHashWithPrivateKey(hash, privateKey) {
+    const Util = require("ethereumjs-util")
```
fedgiac, line 104: `ethereumjs-util` is already imported as `ethUtil` in the same file.

bh2smith: Great catch! ⚾ 🧤

scripts/utils/sign_and_send.js
bh2smith: I guess this is standard, 

https://github.com/ethereum/web3.js/issues/1072#issuecomment-365184511

However, I wonder what would happen if I provided the "0x" along with the private when exporting it. For example, if I had copied my PK from ganache?

Screenshot 2020-05-05 at 4 29 10 PM

which includes the 0x.

fedgiac, line 45: Since we only ever use `account.address` we could directly append `address` to the definition of `account`.

fedgiac:
I think the changes in this PR are good and needed. I also think the changes in the implementation do not introduce any issue that was not already present before this PR, so overall the PR is approved.

I still think we should remove `eth-lightwallet` and clean up all the remaining unused code. The code of this PR is compatible with our current liquidity setup, it's sufficient to extract the default Ganache private key from the hardcoded mnemonic. Only exception is the complete liquidity provision script, since the account that creates the brackets is also the proposer account with the current changes, and must thus hold funds. Since we'll have new brackets only in the new multisig, I don't think we need this kind of compatibility.

josojo:
Okay, then I will remove the light wallet before we merge this one.

package.json
```diff
@@ -53,14 +53,10 @@
...
     "eth-gas-reporter": "^0.2.17",
-    "eth-lightwallet": "^4.0.0",
     "ethereumjs-util": "^6.2.0",
     "node-fetch": "^2.6.0",
     "synthetix-js": "=2.21.12",
     "truffle": "^5.1.24",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
-  },
-  "resolutions": {
-    "bitcore-lib": "8.1.1"
```
bh2smith, line 64: This makes me smile from ear to ear!

test/dfusion_multi_safes.js
```diff
@@ -731,7 +710,12 @@ contract("GnosisSafe", function (accounts) {
       }
 
       const requestWithdrawalTransaction = await buildRequestWithdraw(masterSafe.address, withdrawals)
-      await execTransaction(masterSafe, lw, requestWithdrawalTransaction, "request withdrawal for all brackets")
+      await execTransaction(
+        masterSafe,
+        safeOwner.privateKey,
+        requestWithdrawalTransaction,
+        "request withdrawal for all brackets"
```
fedgiac, line 717: I really wonder where this string came from. Maybe you can remove it, since you removed a similar one somewhere else in the code?

test/dfusion_multi_safes.js
```diff
@@ -820,7 +802,12 @@ contract("GnosisSafe", function (accounts) {
       // tries to transfer more funds to master than available, script should be aware of it
       const transferFundsToMasterTransaction = await buildTransferFundsToMaster(masterSafe.address, withdrawalsModified, true)
 
-      await execTransaction(masterSafe, lw, transferFundsToMasterTransaction, "transfer funds to master for all brackets")
+      await execTransaction(
+        masterSafe,
+        safeOwner.privateKey,
+        transferFundsToMasterTransaction,
+        "transfer funds to master for all brackets"
```
fedgiac, line 809: Candidate string to remove, as before.

scripts/utils/sign_and_send.js
```diff
@@ -45,10 +45,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       ADDRESS_0,
       nonce
     )
-    const lightWallet = await createLightwallet()
-    const account = lightWallet.accounts[0]
-    console.log(`Signing and posting multi-send transaction request from proposer account ${account}`)
-    const sigs = signTransaction(lightWallet, [account], transactionHash)
+    const privateKey = process.env.PK
+    const account = web3.eth.accounts.privateKeyToAccount(privateKey)
```
fleupold, line 49: This method is expecting a 0x prefixed string but the PK is exported without the 0x in truffle (https://web3js.readthedocs.io/en/v1.2.0/web3-eth-accounts.html#privatekeytoaccount)

fleupold:
I see now that if you tested this change by exporting a PK with the 0x prefix it works.

I should have checked the discussion on this PR before commenting above as it was originally implemented in the way that I would have expected it (allowing to paste a PK from MM).

I might have further been mislead by the absence of a test plan.

PR #209, opened 2020-05-06
[fix] hardcoded gas value

josojo:
In the previous PR hardcoding the gasLimit, I only checked that MetaMask shows sufficient gas for the transaction. 
But actually, the gasAmount is then restricted by the Gnosis safe to 4 million, which might not always be sufficient. Hence, I increased the hardcoded value.

Screen Shot 2020-05-06 at 13 14 54

testplan:
run the following command:

```
export NETWORK_NAME=rinkeby
export GAS_PRICE_GWEI=9
export MASTER_SAFE=your master safe
export PK=...

npx truffle exec scripts/complete_liquidity_provision.js --targetToken=0 --stableToken=7 --lowestLimit=0.10 --highestLimit=1 --currentPrice=0.70 --masterSafe=$MASTER_SAFE --investmentTargetToken=0.05 --investmentStableToken=0.05 --fleetSize=20 --network=$NETWORK_NAME
```
and see the transaction goes through.

The downside of this PR is that we are signing tx of 9M gas in Metamask.

### DIFF
--- a/scripts/utils/sign_and_send.js
+++ b/scripts/utils/sign_and_send.js
@@ -28,9 +28,9 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     if (nonce === null) {
       nonce = (await masterSafe.nonce()).toNumber()
     }
-    const safeTxGas = 4000000 // Since the gas can not yet be estimated reliably, we are hard coding it.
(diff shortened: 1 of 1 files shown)


fleupold:
Might become superseded by #211

josojo:
Felix PR is better :)

Issue #193, opened 2020-04-30
Verification scripts for signer


fedgiac:
A signer must currently trust the executor of the script and sign almost blindly. A verification script could repeat the steps executed by the deployment script and confirm that the (Safe) hash of the transaction is correct.

fleupold:
It seems relatively straightforward to introduce a dry-run parameter and print the transaction hash instead of sending it inside signAndSend.

I will also add an optional nonce argument so that the same nonce as in the original tx can be used.

It's a little bit trickier for deploying the actual fleet of safes. There we rely on event emission to get the contract addresses. In order to dry-run this it would be beneficial to also return them from that function so that we could `.call` it without actually executing it.

Can we easily change the `FleetFactory.sol` code or would this have consequences on the previously created contracts?

Alternatively we could also require to pass in hardcoded brackets (which will eventually skip this step), but it would be nice if the two wouldn't have to be tightly coupled.

@fedgiac @josojo what do you guys think?

fedgiac:
I'm not sure we can easily merge the dry run and the script for the verifier.

I see two implicit assumptions in this approach:
1. the verifier script is run before the script actually deploying the bracket (otherwise `.call` would return new addresses)
2. no other brackets have been created _by the FleetFactory contract_ between when the verifier script runs and the deployer script runs (otherwise the two hashes wouldn't match)

Point 2. is because (to my current understanding) an address is computed deterministically from the address of the creator (in our case `FleetFactory`) and the nonce counting the number of transactions sent from this address.

I think 2 is particulary worrisome as soon as we are not the only ones using the scripts anymore.
Having the nonce as an extra parameter helps, but I'm not sure how the nonce of a contract is exactly computed and if and how a user could retrieve it from some nice interface like Etherscan. Also, it's yet another command-line option to understand.

So I still think the verifier script should read from the event. In this case you can filter based on the fleet owner, and the assumptions become:
1. the verifier script is run _after_ the script actually deploying the bracket
2. no other brackets have been created _with the same owner by the FleetFactory contract_  between when the deployer script runs and the verifier script runs

The second set of assumptions is not exempt from issue: anyone could create in between the deployer and verifier transactions a new fleet with the same owner, thus making the hash check fail. This is still annoying but assumes an active adversary.

In general, it's definitely feasible to change the logic of `FleetFactory.sol`, very little depends on it.

[...]

PR #210, opened 2020-05-06
Use existing brackets instead of new fleet (if desired)

### LINKED ISSUE #191: Make complete liquidity provision compatible with hardcoded addresses

bh2smith:
This change allows users to specify the existing sub safes for use in the bracket strategy. Note that step numbers were removed because the logical ordering no longer makes sense.

Observe that the safes are required to be solely owned by master safe and the script will error out if this is not the case.

I made a few grammatical changes to the README while I was on this, will happily move it into a different PR.

closes #191

**Test Plan:**

This one will work, 

```
npx truffle exec scripts/complete_liquidity_provision.js --targetToken 1 --stableToken 4 --lowestLimit 200 --fleetSize 2 --highestLimit 216 --currentPrice 208 --masterSafe 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --investmentTargetToken 0.1 --investmentStableToken 1 --network rinkeby --brackets=0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636,0xfA4a18c2218945bC018BF94D093BCa66c88D3c40
```

and return the following output:

```
==> Performing safety checks
==> Skipping safe deployment and using brackets safeOwners
==> Building orders and deposits
Batch Exchange 0xC576eA7bd102F7E476368a5E98FA455d1Ea34dE2
Constructing bracket trading strategy order data between the limits 200-216 USDC per WETH
Safe 0 - 0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636:
  Buy  WETH with USDC at 200
  Sell WETH for  USDC at 207.8460969082653
Safe 1 - 0xfA4a18c2218945bC018BF94D093BCa66c88D3c40:
  Buy  WETH with USDC at 207.8460969082653
  Sell WETH for  USDC at 216.00000000000006
Transaction bundle size 4
==> Sending the order placing transaction to gnosis-safe interface, please execute this transaction first
Aquiring Transaction Hash
Signing and posting multi-send transaction request from proposer account 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1
Transaction awaiting execution in the interface https://rinkeby.gnosis-safe.io/app/#/safes/0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c/transactions
==> Sending the funds transferring transaction, please execute this transaction second
Aquiring Transaction Hash
Signing and posting multi-send transaction request from proposer account 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1
Transaction awaiting execution in the interface https://rinkeby.gnosis-safe.io/app/#/safes/0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c/transactions
```

For a that failing example replace either one of those brackets addresses with some other safe address that is not solely owner by master safe

Example:

- `0xe6858FF4C3Dbecb91E7A91aEC9913DDbB9b30344` owned by two owners that are not the master safe.

- `0x3273449DB03e0e1bc80586b6093911b3624556EA` which is owned by multiple accounts (including the master safe), but not owned solely by the master safe.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -15,9 +15,9 @@ yarn run networks-inject
 ```
 
 Create a gnosis-safe wallet [here-mainnet](https://gnosis-safe.io) or [here-rinkeby](https://rinkeby.gnosis-safe.io). This wallet will be called your Master Safe in the following. It is used to bundle the transactions and setup the bracket-traders.
-This Master Safe must have an additional owner `0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1`, known as the "Proposer" account. The following scripts will use this account to propose transactions to the interface. This implies that the mnemonic phrase for this "Proposer" account is stored in plain text within this project.
+This Master Safe must have an additional owner `0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1`, known as the "Proposer" account. The following scripts will use this account to propose transactions to the interface. This implies that the mnemonic phrase for this "Proposer" account is available in plain text within this project.
 
-In order to have a secure setup, make sure that _your Master Safe always requires one more signature than just the signature of the Proposer account to send a transaction_. _Otherwise, everyone can steal the funds from your account!_
+In order to have a secure setup, make sure your Master Safe always requires one more signature than just the signature of the Proposer account to send a transaction_. Otherwise, everyone can steal the funds from your account!_
 
 Setup env variables for the deployment process:
 
@@ -34,16 +34,16 @@ export MASTER_SAFE=<master safe>
 In order to deploy new bracket-trader contracts, place orders on behalf of the newly mined contracts and fund their accounts on the exchange, one only has to run the `complete_liquidity_provision` script.
 It will send one ethereum transaction and send two transactions request to the gnosis-safe interface.
 
-The ethereum transasction will create the bracket-traders. For this transaction the provided private key will be used to pay for the gas.
+The ethereum transaction will create the bracket-traders. For this transaction the provided private key will be used to pay for the gas.
 The first request of the script will generate orders on behalf of the bracket-traders.
-Please sign this transaction in the gnosis-safe interface and double check that the prices of the orders are as intended - for example in [telegram-mainnet](https://t.me/gnosis_protocol) or [telegram-rinkeby](https://t.me/gnosis_protocol_dev) channels.
+Please sign this transaction in the gnosis-safe interface and double check the order prices are intended - for example in [telegram-mainnet](https://t.me/gnosis_protocol) or [telegram-rinkeby](https://t.me/gnosis_protocol_dev) channels.
 The second request generates a transaction funding the bracket-traders' accounts on the exchange.
 Making the requests to the gnosis-interfaces does not cost any gas. However, signing and executing the transactions in the gnosis-safe interface will incur gas costs.
 
 Here is an example script invocation:
 
 ```js
-npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --investmentTargetToken=10 --investmentStableToken=1000 --fleetSize=20 --network=$NETWORK_NAME
+npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --investmentTargetToken=10 --investmentStableToken=1000 --fleetSize=10 --network=$NETWORK_NAME
 ```
 
 The prices must be specified in terms of 1 target token = x stable tokens.
@@ -59,7 +59,7 @@ Instead of doing all the steps with one script, the different steps can also be 
 
 ### Deploy safes
 
-Requires that Master is already deployed.
+Requires that Master Safe has already been deployed.
 
 An example of the usage would be:
 
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -87,7 +87,7 @@ module.exports = async (callback) => {
     const investmentTargetToken = toErc20Units(argv.investmentTargetToken, targetTokenDecimals)
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
 
-    if (argv.brackets.length > 0) {
+    if (argv.brackets) {
       assert(argv.fleetSize === argv.brackets.length, "Please ensure fleetSize equals number of brackets")
     }
     assert(argv.fleetSize % 2 === 0, "Fleet size must be a even number for easy deployment script")
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -117,7 +117,7 @@ module.exports = async (callback) => {
     }
 
     let bracketAddresses
-    if (argv.brackets.length > 0) {
+    if (argv.brackets) {
       console.log("==> Skipping safe deployment and using brackets safeOwners")
       bracketAddresses = argv.brackets
       // Ensure that safes are all owned solely by masterSafe
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -7,6 +7,7 @@ const {
   buildTransferApproveDepositFromOrders,
   buildOrders,
   checkSufficiencyOfBalance,
+  isOnlySafeOwner,
 } = require("./utils/trading_strategy_helpers")(web3, artifacts)
 const { isPriceReasonable, areBoundsReasonable } = require("./utils/price_utils")(web3, artifacts)
 const { signAndSend } = require("./utils/sign_and_send")(web3, artifacts)
@@ -87,8 +88,10 @@ module.exports = async (callback) => {
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
 
     if (argv.brackets.length > 0) {
-      // Override fleetSize with num brackets when provided.
-      argv.fleetSize = argv.brackets.length
+      assert(
+        argv.fleetSize === argv.brackets.length, 
+        "Please ensure fleetSize equals number of brackets"
+      )
     }
     assert(argv.fleetSize % 2 === 0, "Fleet size must be a even number for easy deployment script")
 
@@ -121,22 +124,21 @@ module.exports = async (callback) => {
       console.log("==> Skipping safe deployment and using brackets safeOwners")
       bracketAddresses = argv.brackets
       // Ensure that safes are all owned solely by masterSafe
-      for (const safeAddr of argv.brackets) {
-        const safeOwners = await (await GnosisSafe.at(safeAddr)).getOwners()
-        // TODO: write function like safeOwnedSolelyBy(safe, owner) -> Bool
-        if (!(safeOwners.length === 1 && safeOwners.includes(masterSafe.address))) {
-          callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
-        }
-      }
+      await Promise.all(
+        bracketAddresses.map(async (safeAddr) => {
+          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
+            callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
+          }
+        })
+      )
     } else {
       console.log(`==> Deploying ${argv.fleetSize} trading brackets`)
       bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
       console.log("List of bracket traders in one line:", bracketAddresses.join())
+      // Sleeping for 3 seconds to make sure Infura nodes have processed 
+      // all newly deployed contracts so they can be awaited.
+      await sleep(3000)
     }
-
-    // Sleeping for 3 seconds to make sure Infura nodes have processed all newly deployed contracts so that
-    // they can be awaited.
-    await sleep(3000)
 
     console.log("==> Building orders and deposits")
     const orderTransaction = await buildOrders(
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -88,10 +88,7 @@ module.exports = async (callback) => {
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
 
     if (argv.brackets.length > 0) {
-      assert(
-        argv.fleetSize === argv.brackets.length, 
-        "Please ensure fleetSize equals number of brackets"
-      )
+      assert(argv.fleetSize === argv.brackets.length, "Please ensure fleetSize equals number of brackets")
     }
     assert(argv.fleetSize % 2 === 0, "Fleet size must be a even number for easy deployment script")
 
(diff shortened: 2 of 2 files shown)


scripts/README.md
```diff
@@ -32,32 +36,34 @@ export MASTER_SAFE=<master safe>
...
+Please sign this transaction in the gnosis-safe interface and double check the order prices are as expected before signing the next - for example in [telegram-mainnet](https://t.me/gnosis_protocol) or [telegram-rinkeby](https://t.me/gnosis_protocol_dev) channels.
 The second request generates a transaction funding the bracket-traders' accounts on the exchange.
 Making the requests to the gnosis-interfaces does not cost any gas. However, signing and executing the transactions in the gnosis-safe interface will incur gas costs.
 
 Here is an example script invocation:
 
 ```js
-npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --investmentTargetToken=10 --investmentStableToken=1000 --fleetSize=20 --network=$NETWORK_NAME
+npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --investmentTargetToken=10 --investmentStableToken=1000 --fleetSize=10 --network=$NETWORK_NAME
```
bh2smith, line 48: I changed this because the default is already 20.

scripts/complete_liquidity_provision.js
```diff
@@ -79,9 +86,13 @@ module.exports = async (callback) => {
     const investmentTargetToken = toErc20Units(argv.investmentTargetToken, targetTokenDecimals)
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
 
-    assert(argv.fleetSize % 2 == 0, "Fleet size must be a even number for easy deployment script")
+    if (argv.brackets.length > 0) {
+      // Override fleetSize with num brackets when provided.
+      argv.fleetSize = argv.brackets.length
+    }
+    assert(argv.fleetSize % 2 === 0, "Fleet size must be a even number for easy deployment script")
```
fleupold, line 93: It seems like fleetSize makes only sense to specify if you don't specify `--brackets`. Should we move all fleet size checks (<= 23, event etc.) into the if branch that actually uses this value (the one that calls deployFleetOfSafes)

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,29 @@ module.exports = async (callback) => {
...
+      console.log("==> Skipping safe deployment and using brackets safeOwners")
+      bracketAddresses = argv.brackets
+      // Ensure that safes are all owned solely by masterSafe
+      for (const safeAddr of argv.brackets) {
+        const safeOwners = await (await GnosisSafe.at(safeAddr)).getOwners()
+        // TODO: write function like safeOwnedSolelyBy(safe, owner) -> Bool
+        if (!(safeOwners.length === 1 && safeOwners.includes(masterSafe.address))) {
+          callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
+        }
+      }
```
fleupold, line 130: Should we extract all these check into one method? I think Federico wanted to verify as well that there are not yet any orders.

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,29 @@ module.exports = async (callback) => {
...
-    console.log(`2. Deploying ${argv.fleetSize} trading brackets`)
-    const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
-    console.log("List of bracket traders in one line:", bracketAddresses.join())
+    let bracketAddresses
+    if (argv.brackets.length > 0) {
+      console.log("==> Skipping safe deployment and using brackets safeOwners")
+      bracketAddresses = argv.brackets
+      // Ensure that safes are all owned solely by masterSafe
+      for (const safeAddr of argv.brackets) {
+        const safeOwners = await (await GnosisSafe.at(safeAddr)).getOwners()
```
fleupold, line 125: Can we use the existing masterSafe var instead of calling .at again?

scripts/complete_liquidity_provision.js
josojo: Federico has good examples of checking the argv arguments in the withdraw script. Especially, he has several checks that check that either the one or the other parameter is set. You can take a look over there...

scripts/complete_liquidity_provision.js
josojo: For the actual checks, we have already code for it:

```
assert(await isOnlySafeOwner(masterAddress, bracketAddress), "each bracket should be owned only by the master Safe")

        assert(await checkNoOrdersPlaced(bracketAddress, exchange), "each bracket should be owned only by the master Safe")
```
Take also a look into this PR:
https://github.com/gnosis/dex-liquidity-provision/pull/196/files#

Here Federico requests to put the orderchecks into this PR.

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,29 @@ module.exports = async (callback) => {
...
+      }
+    } else {
+      console.log(`==> Deploying ${argv.fleetSize} trading brackets`)
+      bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
+      console.log("List of bracket traders in one line:", bracketAddresses.join())
+    }
 
     // Sleeping for 3 seconds to make sure Infura nodes have processed all newly deployed contracts so that
     // they can be awaited.
     await sleep(3000)
```
josojo, line 139: The sleep statement can go into the else branch of above

scripts/complete_liquidity_provision.js
bh2smith: I have explicitly set `fleetSize`equal to the bracket array length so the equivalent checks can be performed without requiring any other changes. This was the least invasive and still does the same thing. I didn't really want to get involved with the business of putting a bunch of logical constraints on the command line arguments.

scripts/complete_liquidity_provision.js
bh2smith: > Can we use the existing masterSafe var instead of calling .at again?

These are not master safes. They are the slave safes.

bh2smith: @josojo are you suggesting that I wait until #196 is merged? Or include this suggestion as an addition here? I imagine it wouldn't hurt to ensure no orders exist for these brackets, but I really don't like this strict enforcement. What if I wanted to reuse the same brackets with renewed orders? I could make a "proceedAnyway" verification, but first please clarify what it was I was I am supposed to use from #196 or if I am to wait until it is merged.

scripts/complete_liquidity_provision.js
bh2smith: > checking the argv arguments in the withdraw script. 

I will take a look, but I don't exactly feel the need to impose such complex logic unless someone would really like it. Please let me know if this is a suggestion, or a must-have.

scripts/complete_liquidity_provision.js
bh2smith: Yes good catch! I had actually meant to do this earlier.

scripts/complete_liquidity_provision.js
fedgiac: With the current usage of the script it makes sense to have both arguments set, then you'd also check whether the length of `argv.brackets` coincides with `argv.fleetSize` (notes on usage [here](https://github.com/gnosis/dex-liquidity-provision/issues/193#issuecomment-624198288)).

scripts/complete_liquidity_provision.js
fedgiac: Every safe is new, so you need to await it at some point. As it is right now it would be very slow however because for 20 safes you'd be sending 40 requests in series to Infura, so everything should be wrapped at least by a `Promise.all()`.
I'd consider simply calling `verifyBracketsWellFormed` from `verify_script.js` and not add new ad hoc checks here.

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,29 @@ module.exports = async (callback) => {
...
-    const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
-    console.log("List of bracket traders in one line:", bracketAddresses.join())
+    let bracketAddresses
+    if (argv.brackets.length > 0) {
+      console.log("==> Skipping safe deployment and using brackets safeOwners")
+      bracketAddresses = argv.brackets
+      // Ensure that safes are all owned solely by masterSafe
+      for (const safeAddr of argv.brackets) {
+        const safeOwners = await (await GnosisSafe.at(safeAddr)).getOwners()
+        // TODO: write function like safeOwnedSolelyBy(safe, owner) -> Bool
```
fedgiac, line 126: Like the function `isOnlySafeOwner` from `trading_strategy_helpers.js`? :stuck_out_tongue:

scripts/complete_liquidity_provision.js
josojo: Yes incorporating the ordercheck from PR #196 is the best way in my opinion.

Additionally, you can include the whole function : `verifyBracketsWellFormed` then we should be quite safe.

> but I really don't like this strict enforcement. 
fair enough, a prompt would be the best solution for sure

scripts/complete_liquidity_provision.js
bh2smith: So instead of simply resetting the value to `brackets.length` I should just assert that the length matches the `fleetSize`. I can do this and it would make thing a bit cleaner here, but also seems superfluous. 

I looked at the link you provided, but not exactly sure what relevance it has in this discussion.

Overall, I would be happy to go with the suggestion that `fleetSize` agree with `brackets.length`. It is a simple change, it makes the script more safe and does not impeded on any existing code or impose unnecessarily complex business logic. Unless otherwise stated, I will implement this option.

scripts/complete_liquidity_provision.js
bh2smith: Does this `verifyBracketsWellFormed` already exist and does it do both of the checks we would like to have here? 

That is, that each safe is solely owned by master and that there are not yet any orders?

My new problem with this is that having both ownership and non-existent orders tied up in the same constraint means that I cannot reuse safes for different orders. I would prefer if we have a hard constraint on ownership (because we must) but have an option to "proceed anyway" for the orders check. 

Any opinions?

bh2smith:
> An important check missing is whether there are already orders standing.

I would like to discuss with some others that this "check" still be allowed to proceed if it fails.

scripts/complete_liquidity_provision.js
bh2smith: I have another problem with this. Because I was hoping to avoid "fetching" the master safe again. Unfortunately, this `isOnlySafeOwner` only takes addresses and has to fetch both safes... Is there something that can be done about this or it is really worth the extra fetch?

scripts/complete_liquidity_provision.js
bh2smith: Actually, I don't really see the difference apart from inconvenience from the usability perspective from doing

```js
    if (argv.brackets.length > 0) {
      // Override fleetSize with num brackets when provided.
      argv.fleetSize = argv.brackets.length
    }
```

over

```
    if (argv.brackets.length > 0) {
      // Override fleetSize with num brackets when provided.
      assert.equal(argv.fleetSize, argv.brackets.length, "Please ensure fleetSize == number of brackets")
    }
```

scripts/complete_liquidity_provision.js
bh2smith: @josojo - Could we merge this as is (i.e. without the check for orders) or can we get the order check merged beforehand. I feel like this would be an easy fix/feature to include once both of these are merged rather than have this hanging blocked by the pending PR.

scripts/complete_liquidity_provision.js
bh2smith: @fedgiac - would you agree like to see that the following snippet of code be made into its own function?

```js
      // Ensure that safes are all owned solely by masterSafe
      await Promise.all(
        bracketAddresses.map(async (safeAddr) => {
          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
            callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
          }
        })
      )
```

scripts/complete_liquidity_provision.js
```diff
@@ -79,9 +87,12 @@ module.exports = async (callback) => {
     const investmentTargetToken = toErc20Units(argv.investmentTargetToken, targetTokenDecimals)
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
 
-    assert(argv.fleetSize % 2 == 0, "Fleet size must be a even number for easy deployment script")
+    if (argv.brackets.length > 0) {
```
josojo, line 90: nit-pick: if(argv.brackets) would be shorter.

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,28 @@ module.exports = async (callback) => {
...
-    // they can be awaited.
-    await sleep(3000)
+    let bracketAddresses
+    if (argv.brackets.length > 0) {
+      console.log("==> Skipping safe deployment and using brackets safeOwners")
+      bracketAddresses = argv.brackets
+      // Ensure that safes are all owned solely by masterSafe
+      await Promise.all(
+        bracketAddresses.map(async (safeAddr) => {
+          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
```
josojo, line 126: nit-pick: assert(await isOnlySafeOwner would be shorter.

bh2smith: Yes, I like the shorted, but don't we want to be using callbacks here?

bh2smith:
> In the next PR containing the ordercheck and `verifyBracketsWellFormed`, we could put these ... Into a separate function called: new-bracket-checks and then write unit tests for this function.


My only concern with this idea is that well formed brackets should be strictly enforced while we have all agreed that the ordercheck failure should be allowed to proceed anyway. Of course we can discuss this more on the next PR, but I want to mention that combining these two checks would not be so natural to bind together.

scripts/complete_liquidity_provision.js
bh2smith:
```suggestion
    if (argv.brackets) {
```

scripts/complete_liquidity_provision.js
```diff
@@ -105,15 +116,28 @@ module.exports = async (callback) => {
...
 
-    console.log(`2. Deploying ${argv.fleetSize} trading brackets`)
-    const bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, true)
-    console.log("List of bracket traders in one line:", bracketAddresses.join())
-
-    // Sleeping for 3 seconds to make sure Infura nodes have processed all newly deployed contracts so that
-    // they can be awaited.
-    await sleep(3000)
+    let bracketAddresses
+    if (argv.brackets.length > 0) {
```
bh2smith, line 120:
```suggestion
    if (argv.brackets) {
```

bh2smith:
> The concept looks good. An important check missing is whether there are already orders standing.

This check will be included as part of #196, since they are separate tasks.

bh2smith:
Merging this as is and introducing two additional checks as separate PR:

- `verifyBracketsWellFormed` which is currently unusable (cf #217)
- `checkExisitingOrders`- to be introduced in #196

PR #211, opened 2020-05-06
[GasEstimate] Use safe method to estimate gas for internal call

### LINKED ISSUE #154: Gas amount per gnosis-safe transaction
Currently, we still need to adjust the gas amount in metamask, as the gas estimation for the bundled transactions is not working.

One idea to fix it was to use always a very high gas limit for the transactions:

```
const endpoint = `https://safe-transaction.${network}.gnosis.io/api/v1/safes/${masterSafe.address}/transactions/`
    const postData = {
      to: transaction.to,
      value: transaction.value,
      data: transaction.data,
      operation: transaction.operation,
      safeTxGas: 0, // TODO: magic later <---- modify the number here to 5 million
      baseGas: 0,
      gasPrice: 0, // important that this is zero
      gasToken: ADDRESS_0,
      refundReceiver: ADDRESS_0,
      nonce: nonce.toNumber(),
      contractTransactionHash: transactionHash,
      sender: web3.utils.toChecksumAddress(account),
      signature: sigs,
    }
    const ans = await axios.post(endpoint, postData)
```
(it is the function here: https://github.com/gnosis/dex-liquidity-provision/blob/master/scripts/utils/sign_and_send.js#L18)

However, if we set another value bigger than 0, we always get this error:
```
Error: Request failed with status code 422
    at createError (/Users/alexherrmann/gnosis/dex-liquidity-provision/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/Users/alexherrmann/gnosis/dex-liquidity-provision/node_modules/axios/lib/core/settle.js:17:12)
    at IncomingMessage.handleStreamEnd (/Users/alexherrmann/gnosis/dex-liquidity-provision/node_modules/axios/lib/adapters/http.js:236:11)
    at IncomingMessage.emit (events.js:203:15)
    at endReadableNT (_stream_readable.js:1145:12)
```

Right now we don't have a proper way to debug it. Our current requests look like this:
```
data:
   '0x8d80ff0a0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000357a00bc935bdde7c30ee228b431a602068c1466938698000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000075f8ebf1d7e3affffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff00000000000000000000000000000000070d2ba17b9ee57fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b99000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000037a2ecff63de3d35bd15bcf1f0d2999437edf9b4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000070d2ba17b9ee57fffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000006be6116ead0da7fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b9900000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000005f239a6671bc6d2baef6d7cd892296e678810882000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000006be6116ead0da7fffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff00000000000000000000000000000000067306f123d2ae7fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b99000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000067367bcbb0289b5c588bfabafe1328f35405da9a000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000067306f123d2ae7fffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff00000000000000000000000000000000062af6c2d7cae8bfffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b99000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000007c8eabadd0d1d791b82cafa241b0ccbebff65cc000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000062af6c2d7cae8bfffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000005e60bcc17982f7fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b990000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000dc4770bb2446efdfbbfe78beb661395eec5976f2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000005e60bcc17982fbfffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000005a422e79618b83fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b990000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000772420c7c911adda17223d49ed383d6bafe4aa33000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000005a422e79618b7ffffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000005651a78bbda48bfffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b990000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000c60f18e4158c50cc188d6e84cfe8a14ad3911930000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000005651a78bbda48bfffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000528d25a830ef27fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b99000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000089845a5c9f1ea30da4ed269ef96fc003c00e8052000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000528d25a830ef27fffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000004ef2bcf1308ba7fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b990000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000cf5188d539794387c24a37ee881ad7c0f83386a7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005046a761202000000000000000000000000c576ea7bd102f7e476368a5e98fa455d1ea34de2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000030465cc3e7800000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000024000000000000000000000000000000000000000000000000000000000000002a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000050bf4f000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000fffffffe00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000004ef2bcf1308babfffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000ffffffffffffffffffffffffffffffff0000000000000000000000000000000004b8097012e0267fffffffffffffffff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000041000000000000000000000000ec9279204c734e335f8b7f7f660de19968f15b9900000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000',
  operation: 1,
  safeTxGas: 500000,
  baseGas: 0,
  gasPrice: 0,
  gasToken: '0x0000000000000000000000000000000000000000',
  refundReceiver: '0x0000000000000000000000000000000000000000',
  nonce: 40,
  contractTransactionHash:
   '0x3e11f97c90ca796fb7d4e84924b7c5ccae9325633f55b9fa6c892238b665cdb1',
  sender: '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1',
  signature:
   '0x4b0d61a5a589c3229080f7bc851dd4caca57ba4b0e6847d7af152a1bf35ca9084bbfaf8eb7ae5e3dee189b475b4ae567c108a56d3fef0be644df5c7845ebff4f1b' }
```

fleupold:
Fixes #154 

Instead of hardcoding an internal gas limit (which already turned out to be brittle, cf #209), we should use the Safe's functionality to estimate the internal call appropriately.

This is also done in the [safe-proxy-kit](https://github.com/gnosis/contract-proxy-kit/pull/49/files) and an example of how to do it can be found in the safe contracts [unit tests](https://github.com/gnosis/safe-contracts/pull/188)

There are two tiny unrelated changes in this PR. One is a hotfix for #200 and the other one is a formatting issue that made only printed as:

> Error: Error while talking to the gnosis-interface: [object Object] 

### Test Plan

```
export NETWORK_NAME=rinkeby
export GAS_PRICE_GWEI=9
export MASTER_SAFE=your master safe
export PK=...

npx truffle exec scripts/complete_liquidity_provision.js --targetToken=0 --stableToken=7 --lowestLimit=0.10 --highestLimit=1 --currentPrice=0.70 --masterSafe=$MASTER_SAFE --investmentTargetToken=0.05 --investmentStableToken=0.05 --fleetSize=20 --network=$NETWORK_NAME
```

See txs go through.

Or check

https://rinkeby.etherscan.io/tx/0xbcf31282c2350f8e521123d76d243a3c8a995f21024a8b25776c98c12993b450#internal

which shows the hardcoded value of 4M **before this PR** and

https://rinkeby.etherscan.io/tx/0x37deb7cd729e52e2d29d195df6b4ac16299f1b6b90b14e6f53300e201d39d3e3#internal

for the exact estimated value **after this PR** (2.9M)

### DIFF
--- a/scripts/utils/sign_and_send.js
+++ b/scripts/utils/sign_and_send.js
@@ -19,6 +19,25 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     return new Promise((resolve) => rl.question(message, (answer) => resolve(answer)))
   }
 
+  const estimateGas = async function (masterSafe, transaction) {
+    const estimateCall = masterSafe.contract.methods
+      .requiredTxGas(transaction.to, transaction.value, transaction.data, transaction.operation)
+      .encodeABI()
+    const estimateResponse = await web3.eth.call({
+      to: masterSafe.address,
+      from: masterSafe.address,
+      data: estimateCall,
+      gasPrice: 0,
+    })
+    console.log(estimateResponse)
+    // https://docs.gnosis.io/safe/docs/contracts_tx_execution/#safe-transaction-gas-limit-estimation
+    // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
+    // encoded uint value the first 68 bytes of the error message need to be removed.
+    const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
+    console.log(txGasEstimate)
+    return txGasEstimate
+  }
+
   /**
    * Signs and sends the transaction to the gnosis-safe UI
    * @param {Address} masterAddress Address of the master safe owning the brackets
@@ -28,9 +47,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     if (nonce === null) {
       nonce = (await masterSafe.nonce()).toNumber()
     }
-    const safeTxGas = 4000000 // Since the gas can not yet be estimated reliably, we are hard coding it.
-    // 4000000 is a magic value: This value ensures that the user gets a gas limit proposal in Metask
-    // of roughly 6m - MetaMask adds a buffer of roughly 50% here.
+    const safeTxGas = await estimateGas(masterSafe, transaction)
     const baseGas = 0
     console.log("Aquiring Transaction Hash")
     const transactionHash = await masterSafe.getTransactionHash(
@@ -45,7 +62,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       ADDRESS_0,
       nonce
     )
-    const privateKey = process.env.PK
+    const privateKey = "0x" + process.env.PK
     const account = web3.eth.accounts.privateKeyToAccount(privateKey)
     console.log(`Signing and posting multi-send transaction request from proposer account ${account.address}`)
     const sigs = signHashWithPrivateKey(transactionHash, privateKey)
@@ -67,7 +84,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       signature: sigs,
     }
     await axios.post(endpoint, postData).catch(function (error) {
-      throw new Error("Error while talking to the gnosis-interface: " + error.response.data)
+      throw new Error("Error while talking to the gnosis-interface: " + JSON.stringify(error.response.data))
     })
     const interfaceLink = `https://${linkPrefix[network]}gnosis-safe.io/app/#/safes/${masterSafe.address}/transactions`
     console.log("Transaction awaiting execution in the interface", interfaceLink)
--- a/scripts/utils/sign_and_send.js
+++ b/scripts/utils/sign_and_send.js
@@ -35,7 +35,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
     // encoded uint value the first 68 bytes of the error message need to be removed.
     const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
-    // Multiply with 64/63 due to EIP-150
+    // Multiply with 64/63 due to EIP-150 (https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md)
     return Math.ceil((txGasEstimate * 64) / 63)
   }
 
--- a/scripts/utils/sign_and_send.js
+++ b/scripts/utils/sign_and_send.js
@@ -29,12 +29,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       data: estimateCall,
       gasPrice: 0,
     })
-    console.log(estimateResponse)
     // https://docs.gnosis.io/safe/docs/contracts_tx_execution/#safe-transaction-gas-limit-estimation
     // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
     // encoded uint value the first 68 bytes of the error message need to be removed.
     const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
-    console.log(txGasEstimate)
     return txGasEstimate
   }
 
--- a/scripts/utils/sign_and_send.js
+++ b/scripts/utils/sign_and_send.js
@@ -35,7 +35,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
     // encoded uint value the first 68 bytes of the error message need to be removed.
     const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
-    return txGasEstimate
+    // Multiply with 64/63 due to EIP-150
+    return Math.ceil((txGasEstimate * 64) / 63)
   }
 
   /**


scripts/utils/sign_and_send.js
```diff
@@ -45,7 +60,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       ADDRESS_0,
       nonce
     )
-    const privateKey = process.env.PK
+    const privateKey = "0x" + process.env.PK
```
josojo, line 63: Could you implement here your proposed solution from the chat, so that no matter how the user is providing the PK?

fleupold: This is just a placeholder to make the testplan work for the time being. I can do the proposed change in a separate PR.

fleupold: https://github.com/gnosis/dex-liquidity-provision/pull/212

fedgiac:
I'm surprised this approach works: an issue I expect to see with this method is that `requiredTxGas` may compute the exact amount spent in a transaction, while our gas limit must also take into account the gas refunds, of which we have many. I've checked the Gnosis Safe code and it says "Estimate without refunds", so maybe the example works because Metamask does the job of taking care of the refunds?
It's a very welcome change if this is not an issue, let's wait for Richard's opinion on this.

scripts/utils/sign_and_send.js
```diff
@@ -19,6 +19,23 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+      to: masterSafe.address,
+      from: masterSafe.address,
+      data: estimateCall,
+      gasPrice: 0,
+    })
+    // https://docs.gnosis.io/safe/docs/contracts_tx_execution/#safe-transaction-gas-limit-estimation
+    // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
+    // encoded uint value the first 68 bytes of the error message need to be removed.
+    const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
+    return txGasEstimate
```
rmeissner, line 36: I would recomment to up this value by at least 1/64 ... due to EIP-150 ;)

rmeissner: we experienced some issues when performing gas intensive transactions, therefore we now do a more complex estimation where we double check if the transaction can be performed with the set safeTxGas limit.

rmeissner:
@fedgiac as far as I can see you do not make use of the refunde feature. Also since the refund would revert if there is not enough gas it will indeed be accounted for in the estimate from Metamask (same for the gas required to send the data to the contract).

scripts/utils/sign_and_send.js
rmeissner: A little more details to [EIP-150](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md). There is a section that says 

> If a call asks for more gas than the maximum allowed amount (i.e. the total amount of gas remaining in the parent after subtracting the gas cost of the call and memory expansion), do not return an OOG error; instead, if a call asks for more gas than all but one 64th of the maximum allowed amount, call with all but one 64th of the maximum allowed amount of gas (this is equivalent to a version of EIP-901 plus EIP-1142). CREATE only provides all but one 64th of the parent gas to the child call.

scripts/utils/sign_and_send.js
bh2smith: Prima!

fedgiac:
Ah, I thought that the refund mentioned in the code was the gas refund you receive when for example freeing memory in a transaction or selfdestructing a contract. So `requiredTxGas` returns the estimated gas limit without subtracting the amount you get back from the EVM gas refunds @rmeissner?
Otherwise, my worry is that if Metamask's gas estimation falls between the estimated gas amount and the estimated gas amount plus refunds it would still fail.

rmeissner:
It is not possible to see gas refunds (triggered by contract deletion or storage freeing) in a solidity contract afaik. As refund is only applied once the transaction is completely executed, since it depends on the total used gas.

bh2smith:
I resolved the conflict, because I "introduced" it. This sexy PR is ready for merge!

scripts/utils/sign_and_send.js
```diff
@@ -35,7 +35,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     // The value returned by requiredTxGas is encoded in a revert error message. For retrieving the hex
     // encoded uint value the first 68 bytes of the error message need to be removed.
     const txGasEstimate = parseInt(estimateResponse.substring(138), 16)
-    return txGasEstimate
+    // Multiply with 64/63 due to EIP-150
```
bh2smith, line 38: Could we put a link to this EIP here?

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-150.md

PR #213, opened 2020-05-06
Change amount withdrawn with --brackets

### LINKED ISSUE #159: Revise caluclation of the withdrawal amount
With the current withdrawal logic, users might only be able to withdraw less than the amount in the pending withdrawal request, and thus the multisend withdraw+transferFundsToMaster would fail.

We should try to use web3.getStorageAt and manually compute the storage position of balanceState[braketAddress] to calculate the right amount.

fedgiac:
Implements the changes described in #159.
When executing `withdraw.js` with `--brackets`:
- `--requestWithdraw` uses the maximum value for `uint256`. The idea is that a user who doesn't specify the amounts with a deposit file simply wants to withdraw everything in the brackets. Before this PR the script would only request to withdraw the amount that was available at the time of the request, meaning that amounts deriving from future trades would not be withdrawn. The new behavior is also useful for an emergency shutdown of the brackets.
- `--withdraw` and `--withdraw --transferFundsToMaster` compute the actual amount available for withdrawing. Before they relied on the amount specified in the withdraw request.
- `--transferFundsToMaster` is unchanged.

Downside of the changes is that the user is not warned anymore when using `--withdraw` if there are funds available to be withdrawn but no withdraw was requested or the withdraw is not yet valid.

Closes #159.

Test plan: unit tests.

### DIFF
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -1,5 +1,6 @@
 module.exports = function (web3, artifacts) {
   const fs = require("fs").promises
+  const { getWithdrawableAmount } = require("@gnosis.pm/dex-contracts/src")
 
   const { fromErc20Units, shortenedAddress } = require("../utils/printing_tools")
   const {
@@ -41,35 +42,14 @@ module.exports = function (web3, artifacts) {
     }
   }
 
-  const getMaxWithdrawableAmount = async function (
-    argv,
-    bracketAddress,
-    tokenData,
-    exchange,
-    currentBatchId,
-    printOutput = false
-  ) {
-    const log = printOutput ? (...a) => console.log(...a) : () => {}
+  const getMaxWithdrawableAmount = async function (argv, bracketAddress, tokenData, exchange) {
     let amount
     const token = tokenData.instance
     if (argv.requestWithdraw) {
       amount = bnMaxUint.toString()
     } else {
       if (argv.withdraw) {
-        // replace with getWithdrawableAmount(bracketAddress, tokenData.address, exchange, web3)
-        const pendingWithdrawal = await exchange.getPendingWithdraw(bracketAddress, tokenData.address)
-        amount = pendingWithdrawal[0].toString()
-        if (pendingWithdrawal[1].toNumber() >= currentBatchId) {
-          const batchIdsLeft = pendingWithdrawal[1].toNumber() - currentBatchId + 1
-          log(
-            `Warning: requested withdrawal of ${amount} ${
-              tokenData.symbol
-            } for bracket ${bracketAddress} cannot be executed until ${batchIdsLeft} ${
-              batchIdsLeft == 1 ? "batch" : "batches"
-            } from now, skipping`
-          )
-          amount = "0"
-        }
+        amount = await getWithdrawableAmount(bracketAddress, tokenData.address, exchange, web3)
       }
       if (argv.transferFundsToMaster) {
         amount = amount || (await token.balanceOf(bracketAddress)).toString()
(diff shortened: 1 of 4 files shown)


scripts/wrapper/withdraw.js
```diff
@@ -40,34 +42,18 @@ module.exports = function (web3, artifacts) {
...
-  const getMaxWithdrawableAmount = async function (
-    argv,
-    bracketAddress,
-    tokenData,
-    exchange,
-    currentBatchId,
-    printOutput = false
-  ) {
-    const log = printOutput ? (...a) => console.log(...a) : () => {}
+  const determineAmountToWithdraw = async function (argv, bracketAddress, tokenData, exchange) {
```
fleupold, line 45: This method is a little bit confusing to me. It does three very different things depending on some fields on the very general argv argument.

It seems that this logic is very tightly coupled to the method that gets invoked with this amount in the end (`buildRequestWithdraw`, `buildWithdraw`, `buildWithdrawAndTransferFundsToMaster`, etc).

Maybe having three different method and calling them inside this other if else block or even inside the `build...` methods would make this coupling more explicit?

test/scripts/withdraw.js
```diff
@@ -18,6 +19,8 @@ const { waitForNSeconds, execTransaction } = require("../../scripts/utils/intern
 const prepareWithdraw = require("../../scripts/wrapper/withdraw")(web3, artifacts)
 const { toErc20Units, fromErc20Units } = require("../../scripts/utils/printing_tools")
 
+const bnMaxUint256 = new BN(2).pow(new BN(256)).subn(1)
```
fleupold, line 22: ubernit: why not reuse the one from printing tools?

test/scripts/withdraw.js
```diff
@@ -433,6 +436,25 @@ contract("Withdraw script", function (accounts) {
...
+        masterSafe: masterSafe.address,
+        brackets: bracketAddresses,
+        tokens: [tokenInfo[0].address],
+        withdraw: true,
+        transferFundsToMaster: true,
+      }
+      const transaction = await prepareWithdraw(argv)
+      // if the built transaction used the token balance of the bracket instead of zero, then
+      // the following transaction would fail.
+      await await execTransaction(masterSafe, safeOwner.privateKey, transaction)
```
josojo, line 456: one await too much

bh2smith:
```suggestion
      await execTransaction(masterSafe, safeOwner.privateKey, transaction)
```

test/scripts/withdraw.js
bh2smith: Agree that this is probably already somewhere in the project, exported and ready for import.

scripts/wrapper/withdraw.js
fedgiac: That's a good idea, it looks like there wouldn't be too much code duplication by doing that. I'll open an issue for that.

test/scripts/withdraw.js
fedgiac: I thought about it, but since this is a test and that is part of the logic of the code I think it makes sense to define them independently. I don't have a strong opinion on that, I'm fine with importing it from the printing tools. Or to add this constant to `test_utils.js` and import it from there.

test/scripts/withdraw.js
fedgiac: Code suggestions! How do you do that?

fleupold: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request

Point 6

scripts/wrapper/withdraw.js
fleupold: Isn't this something that could be tackled in this PR. It seems like a bunch of special casing is only introduced here which creates this weird coupling.

fedgiac: Maybe I don't understand your proposal, but as I read it I'd have to refactor most of the file to do what you want. It's not a large file, but it would multiply by three at least the number of lines changed.

fedgiac: I'll leave this up as #222 since it's a nice small task for a cleanup session.

PR #214, opened 2020-05-06
Script for getting already deployed brackets

josojo:
For better documentation and user research, we need a simple script to extract all already deployed addresses.

This PR provides this script and writes the brackets into a csv file, as this should be our current platform, as discussed with @anxolin

testplan:
```
npx truffle exec scripts/getDeployedBrackets.js --masterSafe=$MASTER_SAFE --network=$NETWORK_NAME
```


scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,34 @@
...
+module.exports = async (callback) => {
+  try {
+    const fleetFactory = await FleetFactory.deployed()
+    const events = await fleetFactory.getPastEvents("FleetDeployed", {
+      filter: { owner: argv.masterSafe },
+      fromBlock: 0,
+      toBlock: "latest",
+    })
+    const bracketsAsObjects = events.map((object) => object.returnValues.fleet)
+    const bracketAddresses = [].concat(...bracketsAsObjects)
```
fedgiac, line 18: I think this should be its own function, we are likely to reuse it at a certain point for the view script.

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,34 @@
...
+
+    // writing the brackets into a csv file
+    const createCsvWriter = require("csv-writer").createObjectCsvWriter
+    const csvWriter = createCsvWriter({
+      path: "./bracket-addresses.csv",
+      header: [{ id: "bracketAddress", title: "bracket-address" }],
+    })
+    const bracketsFormatted = bracketAddresses.map((bracketAddress) => [{ bracketAddress: bracketAddress }])
+    const records = [].concat(...bracketsFormatted)
```
fedgiac, line 27: Why is this line needed? As I understand it it's a way to create a shallow copy of `bracketsFormatted`, but even if `csvWriter.writeRecords` were to change the object you don't need it anymore.

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,34 @@
...
+      fromBlock: 0,
+      toBlock: "latest",
+    })
+    const bracketsAsObjects = events.map((object) => object.returnValues.fleet)
+    const bracketAddresses = [].concat(...bracketsAsObjects)
+
+    // writing the brackets into a csv file
+    const createCsvWriter = require("csv-writer").createObjectCsvWriter
+    const csvWriter = createCsvWriter({
+      path: "./bracket-addresses.csv",
```
fleupold, line 23: nit: Given it's simple, I would allow this to be optionally passed in via the cmd line.

scripts/get_deployed_brackets.js
fleupold: You mean a helper like `getAllBrackets`? Sounds like a good idea.

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,34 @@
...
+    const events = await fleetFactory.getPastEvents("FleetDeployed", {
+      filter: { owner: argv.masterSafe },
+      fromBlock: 0,
+      toBlock: "latest",
+    })
+    const bracketsAsObjects = events.map((object) => object.returnValues.fleet)
+    const bracketAddresses = [].concat(...bracketsAsObjects)
+
+    // writing the brackets into a csv file
+    const createCsvWriter = require("csv-writer").createObjectCsvWriter
```
fleupold, line 21: ubernit: We usually require at the top of the file. Why inline it here?

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+          tradingPair = " - not yet defined -"
+        }
+        return {
+          Address: bracketAddress,
+          Type: "liquidity",
+          Description: "bracket-strategy",
+          Tags: "bracket-strategy on the pair " + tradingPair,
+          EL: "https://etherscan.io/address/" + bracketAddress,
+          CP: '"' + bracketAddress + '"',
```
fleupold, line 55: What is the value of this field if it's the exact same as address?

Syntactic nit: you can use `"${bracketAddress}"`

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,65 @@
...
+          tradingPair =
+            (await tokenInfoPromises[orders[0].sellToken]).symbol + " - " + (await tokenInfoPromises[orders[0].buyToken]).symbol
+        } else {
+          tradingPair = " - not yet defined -"
+        }
+        return {
+          Address: bracketAddress,
+          Type: "liquidity",
+          Description: "bracket-strategy",
+          Tags: "bracket-strategy on the pair " + tradingPair,
```
fleupold, line 53: Invert? The description sounds more like a tag and the tag reads more like a description.

scripts/get_deployed_brackets.js
josojo: thanks for syntactic nit.

Well,  @anxolin proposed this, and I followed this proposal.

anxolin: the Etherscan link and CP is not needed. Those are columns in the Google Sheets, but are autogenerated

Important are: address, type, description and tags

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,65 @@
...
+          const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, [orders[0].buyToken, orders[0].sellToken])
+          tradingPair =
+            (await tokenInfoPromises[orders[0].sellToken]).symbol + " - " + (await tokenInfoPromises[orders[0].buyToken]).symbol
+        } else {
+          tradingPair = " - not yet defined -"
+        }
+        return {
+          Address: bracketAddress,
+          Type: "liquidity",
+          Description: "bracket-strategy on the pair " + tradingPair,
```
anxolin, line 52: Given that all the addresses come from the same master contract, do you think is valuble to ad that as part of the description?

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,65 @@
...
+
+    const exchange = await getExchange(web3)
+    const records = await Promise.all(
+      bracketAddresses.map(async (bracketAddress) => {
+        const orders = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders.call(bracketAddress))
+        let tradingPair
+        if (orders.length > 0) {
+          const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, [orders[0].buyToken, orders[0].sellToken])
+          tradingPair =
+            (await tokenInfoPromises[orders[0].sellToken]).symbol + " - " + (await tokenInfoPromises[orders[0].buyToken]).symbol
```
anxolin, line 45: A bit of replication here. Maybe you can simplify, and wait just once (Promise.all). Also, destructuring the array/objects may come handy, together with using
Good idea to add auxiliary consts, like `const firstOrder` or  `tokenInfo`.

scripts/get_deployed_brackets.js
```diff
@@ -0,0 +1,65 @@
...
+      ],
+    })
+
+    const exchange = await getExchange(web3)
+    const records = await Promise.all(
+      bracketAddresses.map(async (bracketAddress) => {
+        const orders = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders.call(bracketAddress))
+        let tradingPair
+        if (orders.length > 0) {
+          const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, [orders[0].buyToken, orders[0].sellToken])
```
anxolin, line 43: Not from this PR, but I thought the signature (params and return value) of `fetchTokenInfoFromExchange` was a bit strange when I read it. If this is working no need to change.

Also, ERC20 doesn't necessarily implement symbol or decimals, and it looks like if the contract fails to get the script would fail. So probably a token like SAI o MKR wouldn't work.

scripts/get_deployed_brackets.js
josojo: we don't have any repetition, as the promises are stored in a kinda global object 
`  
const globalTokenPromisesFromId = {}
`

scripts/get_deployed_brackets.js
fedgiac: Good point. What's the standard if the token has no decimals when printing amounts? It defaults to 18?

josojo: It's a good point, we will address it in another PR.

anxolin: Regarding decimals or name, I don't think there's an official "default"
I think some people use 18 as the default, but I don't think is part of the standard: https://eips.ethereum.org/EIPS/eip-20

Regarding the method signature, since it's a script, I wouldn't bother too much, but imo since we are discussing it 😀, if it's an async call it should return a Promise, not an object with ids as keys and promises as values. That's what I find strange.

If the reason is, to allow parallel processing, I would define it like `fetchTokenInfoFromExchange({ exchangeArtifiact, tokenId }): Promise<TokenInfo>`, so you can easily implement any async with that.

nit: Inside the implementation, it uses a `for`, but normally using a map is more common practice. --> map a list of Ids to a list of promises 

nit: It uses global const for caching, that normally should be using UPER_CASE and placed on top. 

nit: Any non-trivial params like this one should have named params, to make it clear what it receives and prevent errors. 

Recommended also to use typescrypt, to prevent issues

scripts/get_deployed_brackets.js
anxolin: With repetition I mean you can se `orders[0]` many times,
`orders[0].buyToken`

Then it's too complex construction of the string, it's a good practice to aim to do one thing in each line (for readability), here you are concatenating, so better not to await two promise, get the first order (2 times) and extract the symbol from the resulting object too

No shame in having more lines and aux const/vars:
```js
const { sellToken, buyOrder } = orders[0]
[ sellTokenInfo, buyTokenInfo ] = await Promise.all([
   tokenInfoPromises[sellToken]),
   tokenInfoPromises[buyToken]) 
])
tradingPair = `${sellTokenInfo.symbol}-${buyTokenInfo.symbol}`
```

Each line above have a single purpose, it's not bad in my view even to have:
```
const firstOrder = orders[0]
const { sellToken, buyOrder } = firstOrder
```
But also, it's fair to do it all at once

PR #219, opened 2020-05-07
Allow to verify liquidity provision tx hashes before executing

### LINKED ISSUE #193: Verification scripts for signer

fleupold:
Closes #193

This PR makes it so that we can verify the tx hash of a liquidity provisioning transaction before signing it in the safe interface.

This is achieved by adding an optional `--dry-run` arguments which instead of sending out transactions only simulates them. In combination with the `--bracket ` argument introduced from #210 this allows reproducing the exact transaction hash that the original proposer created.

I also added another optional `--nonce` argument to allow verifying hashes even when the safe has moved on signing other txs and thus gotten a new nonce. In the regular case where we verify the tx before signing this shouldn't be needed.

Note, that this verification method is still not 100% fool proof, since the web interface doesn't seem to show the tx hash anywhere. Therefore we still trust that the proposer is sharing the hash with use that corresponds to the calldata that was submitted in the interface.

### Test Plan

Run the regular liquidity provision command:

```
origin/191/use_existing_bracketsnpx truffle exec scripts/complete_liquidity_provision.js --targetToken=0 --stableToken=7 --lowestLimit=0.10 --highestLimit=1 --currentPrice=0.70 --masterSafe=$MASTER_SAFE --investmentTargetToken=0.05 --investmentStableToken=0.05 --fleetSize=20 --network=$NETWORK_NAME
```

Then re-run is ammended with the arguments provided as a last statement of the first command (`--nonce` is not required).

See that the tx hashes are the same in both commands.

### DIFF
--- a/networks.json
+++ b/networks.json
@@ -7,8 +7,8 @@
     },
     "4": {
       "links": {},
-      "address": "0x1b94BF378B910e5be70b59e774867a1c919821C7",
-      "transactionHash": "0xf8256a6b6c92a95f9342d0dbb8afee83d1dd6bec8510278893f568ae0114ee3f"
+      "address": "0x9ebF59B1F7F58Db871189a17c078B456014fe515",
+      "transactionHash": "0x992d706340cae916c667cb0733b8dc4e80c66afbde56a5d81ee6af46ba233838"
     }
   },
   "GnosisSafe": {
--- a/networks.json
+++ b/networks.json
@@ -7,8 +7,8 @@
     },
     "4": {
       "links": {},
-      "address": "0x9ebF59B1F7F58Db871189a17c078B456014fe515",
-      "transactionHash": "0x992d706340cae916c667cb0733b8dc4e80c66afbde56a5d81ee6af46ba233838"
+      "address": "0x1b94BF378B910e5be70b59e774867a1c919821C7",
+      "transactionHash": "0xf8256a6b6c92a95f9342d0dbb8afee83d1dd6bec8510278893f568ae0114ee3f"
     }
   },
   "GnosisSafe": {
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -67,7 +67,7 @@ const argv = require("./utils/default_yargs")
     type: "float",
     describe: "Price for the bracket selling at the highest price",
   })
-  .option("dry-run", {
+  .option("verify", {
     type: "boolean",
     default: false,
     describe: "Do not actually send transactions, just simulate their submission",
@@ -93,8 +93,6 @@ module.exports = async (callback) => {
     const stableTokenData = await tokenInfoPromises[stableTokenId]
     const { instance: targetToken, decimals: targetTokenDecimals } = targetTokenData
     const { instance: stableToken, decimals: stableTokenDecimals } = stableTokenData
-
-    const dry_run = argv["dry-run"]
 
     const investmentTargetToken = toErc20Units(argv.investmentTargetToken, targetTokenDecimals)
     const investmentStableToken = toErc20Units(argv.investmentStableToken, stableTokenDecimals)
@@ -154,8 +152,9 @@ module.exports = async (callback) => {
         callback("Error: Existing order verification failed.")
       }
     } else {
+      assert(!argv.verify, "Trading Brackets need to be provided via --brackets when verifying a transaction")
       console.log(`==> Deploying ${argv.fleetSize} trading brackets`)
-      bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize, dry_run)
+      bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize)
       // Sleeping for 3 seconds to make sure Infura nodes have processed
       // all newly deployed contracts so they can be awaited.
       await sleep(3000)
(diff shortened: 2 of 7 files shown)


networks.json
```diff
@@ -7,8 +7,8 @@
     },
     "4": {
       "links": {},
-      "address": "0x1b94BF378B910e5be70b59e774867a1c919821C7",
-      "transactionHash": "0xf8256a6b6c92a95f9342d0dbb8afee83d1dd6bec8510278893f568ae0114ee3f"
+      "address": "0x9ebF59B1F7F58Db871189a17c078B456014fe515",
+      "transactionHash": "0x992d706340cae916c667cb0733b8dc4e80c66afbde56a5d81ee6af46ba233838"
```
fleupold, line 11: Note, I will also have to redeploy mainnet if this change is accepted.

bh2smith: I thought the changes to `fleetFactory` were simply just lint fixes? Did you change something else that meant we had to redeploy?

scripts/complete_liquidity_provision.js
```diff
@@ -84,6 +92,8 @@ module.exports = async (callback) => {
     const { instance: targetToken, decimals: targetTokenDecimals } = targetTokenData
     const { instance: stableToken, decimals: stableTokenDecimals } = stableTokenData
 
+    const dry_run = argv["dry-run"]
```
bh2smith, line 95: Any particular reason you chose to access like a dict instead of like a JS object?

scripts/complete_liquidity_provision.js
```diff
@@ -65,6 +65,14 @@ const argv = require("./utils/default_yargs")
   .option("highestLimit", {
     type: "float",
     describe: "Price for the bracket selling at the highest price",
+  })
+  .option("dry-run", {
+    type: "boolean",
+    describe: "Do not actually send transactions, just simulate their submission",
+  })
```
bh2smith, line 72: Could we add an explicit default of false here?

bh2smith:
Build failing because of commited merge conflict.


Screenshot 2020-05-07 at 11 50 29 AM

scripts/complete_liquidity_provision.js
fleupold: I actually made it return the fleet address on top of emitting them as an event. This way we can "call" the method in the dry run mode.

fleupold:
> I like this very much and approve, but I do not see how this closes the PR #196

You are right, it closes #193

networks.json
fleupold: I actually made it return the fleet address on top of emitting them as an event. This way we can "call" the method in the dry run mode.

scripts/complete_liquidity_provision.js
fleupold: multi word command line arguments are by convention kebab-case and that cannot be directly accessed on the argv as far as I know.

networks.json
bh2smith: Noice! Smort!

scripts/complete_liquidity_provision.js
bh2smith: Interesting. Does this mean we should consider changing our other "multi-word" arguments from camel case?

fleupold:
> I would prefer if we keep the old contract and you would grep the bracket addresses via a function reading the events

See the discussion on #193 This would also rely on an assumption that the verification is done immediately after the original creation.

I was trying to introduce the concept of a dry-run in this PR which can be used to verify the liquidity provision transaction. It turns out however that we cannot dry run it liquidity provision without actually have the safes created. In this case I guess we have two options:

1) require --brackets to be set on --dry-run (fail otherwise)
2) default to querying the last deployed fleet (in which case dry-run is maybe the wrong word and it should rather be verify)

scripts/complete_liquidity_provision.js
```diff
@@ -161,11 +171,20 @@ module.exports = async (callback) => {
...
+    }
+    await signAndSend(masterSafe, orderTransaction, argv.network, nonce, dry_run)
 
     console.log("==> Sending the funds transferring transaction, please execute this transaction second")
-    await signAndSend(masterSafe, bundledFundingTransaction, argv.network, nonce + 1)
+    await signAndSend(masterSafe, bundledFundingTransaction, argv.network, nonce + 1, dry_run)
+
+    if (!dry_run) {
+      console.log(
+        `To verify the transactions run the same script with --dry-run --nonce ${nonce} --brackets ${bracketAddresses.join()}`
```
fedgiac, line 185: Nitpick: I'd use the `--option=value` format since it's the one we (tend to) use in the readme. 
```suggestion
        `To verify the transactions run the same script with --dry-run --nonce=${nonce} --brackets=${bracketAddresses.join()}`
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -189,22 +189,19 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
-    const transcript = await fleetFactory.deployFleet(masterAddress, fleetSize, gnosisSafeMasterCopy.address)
-    const createdSafes = transcript.logs[0].args.fleet
-    log("New Safes created:")
-    createdSafes.forEach((safeAddress, index) => {
-      log("Safe " + index + ":", safeAddress)
-    })
-
-    return createdSafes
+    if (dry_run) {
+      return await fleetFactory.contract.methods.deployFleet(masterAddress, fleetSize, gnosisSafeMasterCopy.address).call()
```
fedgiac, line 200: Nitpick: the `await` is not needed.
```suggestion
      return fleetFactory.contract.methods.deployFleet(masterAddress, fleetSize, gnosisSafeMasterCopy.address).call()
```

josojo:
> > I would prefer if we keep the old contract and you would grep the bracket addresses via a function reading the events
> 
> See the discussion on #193 This would also rely on an assumption that the verification is done immediately after the original creation.

Ah nice, so with events we could be quite sure that we get the right safes, if we are filtering for safe creations with 

> > I would prefer if we keep the old contract and you would grep the bracket addresses via a function reading the events
> 
> See the discussion on #193 This would also rely on an assumption that the verification is done immediately after the original creation.
> 
> I was trying to introduce the concept of a dry-run in this PR which can be used to verify the liquidity provision transaction. It turns out however that we cannot dry run it liquidity provision without actually have the safes created. In this case I guess we have two options:
> 
> 1. require --brackets to be set on --dry-run (fail otherwise)
> 2. default to querying the last deployed fleet (in which case dry-run is maybe the wrong word and it should rather be verify)

While option one is more copy and paste, I would still prefer it, as option 2 can go wrong if the proposer deployed many safes in different transactions and hence introduces more complexity.

scripts/complete_liquidity_provision.js
bh2smith: I agree it has become the standard but I prefer without. Of course this is a matter of opinion and we should probably remain consistent.

bh2smith:
I resolved the conflicts resulting from the merge of #196

bh2smith:
Looks ready for shipping!

josojo:
You might wanna mention this mighty feature in the readme...

PR #224, opened 2020-05-07
Replace unlimited orders with limited but large orders

### LINKED ISSUE #203: Investigate multiplication overflow issue
Because we can run into multiplication overflows during the utility-calculation, we should no longer use them

fedgiac:
Removes unlimited order and uses large orders instead. Closes #203.

Before merging I'd run from a master safe that has 10USDC and 1WETH:
```
npx truffle exec scripts/complete_liquidity_provision.js --targetToken=1 --stableToken=6 --lowestLimit=180 --highestLimit=220 --currentPrice=190 --masterSafe=$MASTER_SAFE --investmentTargetToken=1 --investmentStableToken=10 --fleetSize=20 --network=rinkeby
```
and double check the orders on the Telegram bot [here](https://t.me/s/gnosis_protocol_dev).

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -6,7 +6,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const exchangeUtils = require("@gnosis.pm/dex-contracts")
   const { Fraction } = require("@gnosis.pm/dex-contracts/src")
 
-  const max128 = new BN(2).pow(new BN(128)).subn(1)
+  const { MAX_ORDER_AMOUNT } = require("./constants.js")
 
   const checkCorrectnessOfDeposits = async (
     currentPrice,
@@ -179,11 +179,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @param {integer} stableTokenDecimals number of decimals of the stable token
    * @return {BN[2]} amounts of stable token and target token for an unlimited order at the input price
    */
-  const getUnlimitedOrderAmounts = function (price, targetTokenDecimals, stableTokenDecimals) {
-    let targetTokenAmount = max128.clone()
+  const getLargeOrderAmounts = function (price, targetTokenDecimals, stableTokenDecimals) {
+    let targetTokenAmount = MAX_ORDER_AMOUNT.clone()
     let stableTokenAmount = getOutputAmountFromPrice(price, targetTokenAmount, targetTokenDecimals, stableTokenDecimals)
     if (stableTokenAmount.gt(targetTokenAmount)) {
-      stableTokenAmount = max128.clone()
+      stableTokenAmount = MAX_ORDER_AMOUNT.clone()
       targetTokenAmount = getOutputAmountFromPrice(1 / price, stableTokenAmount, stableTokenDecimals, targetTokenDecimals)
       assert(stableTokenAmount.gte(targetTokenAmount), "Error: unable to create unlimited order")
     }
@@ -225,9 +225,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     areBoundsReasonable,
     checkCorrectnessOfDeposits,
     getOutputAmountFromPrice,
-    getUnlimitedOrderAmounts,
+    getLargeOrderAmounts,
     getDexagPrice,
     checkNoProfitableOffer,
-    max128,
   }
 }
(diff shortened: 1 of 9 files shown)


fleupold:
I was hoping we could make a capped trading limit optional (by passing it via the command line if desired) as it seems only necessary on certain trades and doesn't have to be the default.

fedgiac:
That's a very technical option and I think we should be making the choice for the user here.
1. the reason why a capped amount exists is a smart contract bug. The user should be understanding the bug.
2. even then, it's still very unclear even to us what the right value to use would be, if you are so familiar with the issue it's probably fine for you to change a constant in the source code. I don't think we should be changing that constant on deployments ever after this PR, by the way.
3. the capped amount is computed taking into account the number of decimals of both tokens and the price in some weird way.

We could maybe add a flag `--unlimited` that makes the order truly unlimited, but it's unclear why the user would bother. The tradeoff is having to redo the orders in the far future vs having the orders ignored sometimes. Anyway, I think it's worth discussing now how we want our future deployments to look like, what do Ben and Alex think about this?

bh2smith:
Sorry @fedgiac if I did not clarify what the intention of the PR was meant to be. We have not yet decided for sure that we will reduce the unlimited orders for the moment, but rather change the function `getUnlimitedOrderAmount` to something like `getOrderAmounts` and introduce a new parameter (not a constant) called `maxSellAmount` This way, the function could be used for BOTH _unlimited orders_ and _configurably large_ orders.

I meant to update the issue description, but made the silly assumption that we were all on the same page.

I think also this would be an appropriate place to make this function more general and intuitive to use (use terms like `targetToken` and `relativeToken` instead of `stableToken`) and then making it so that the user must explicitly use the multiplicative inverse price when swapping the tokens to get the inverse prices.

I would imagine, perhaps something like this

`fn getBuyAmountTarget(priceTargetInRelative, targetToken, relativeToken, relativeSellAmount) -> targetBuyAmount`

fedgiac:
@bh2smith, @fleupold, what do you think about the new changes? Is this how you envisioned this PR?
Ben, my understanding of `getBuyAmt` is that it's equivalent to `getOutputAmountFromPrice` up to renaming variables, so I didn't create a new function. Is there a difference I'm missing?
To send unlimited orders you can append `--orderSizeLimit=340282366920938463463374607431768211455` to the command you are running. I maintain that I don't see why any user would want to use this option ever.

scripts/utils/price_utils.js
```diff
@@ -179,11 +179,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @param {integer} stableTokenDecimals number of decimals of the stable token
    * @return {BN[2]} amounts of stable token and target token for an unlimited order at the input price
    */
-  const getUnlimitedOrderAmounts = function (price, targetTokenDecimals, stableTokenDecimals) {
-    let targetTokenAmount = max128.clone()
+  const getLargeOrderAmounts = function (price, targetTokenDecimals, stableTokenDecimals, orderSizeLimit = MAX_ORDER_AMOUNT) {
```
bh2smith, line 182: I was expecting this function to be called `getOrderAmounts` and not to use a global variable, but rather pass the desired amount in.

scripts/utils/price_utils.js
```diff
@@ -179,11 +179,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+    let targetTokenAmount = orderSizeLimit.clone()
     let stableTokenAmount = getOutputAmountFromPrice(price, targetTokenAmount, targetTokenDecimals, stableTokenDecimals)
     if (stableTokenAmount.gt(targetTokenAmount)) {
```
bh2smith, line 185: I was also expecting that this multiplicative inverse logic would no longer be a part of the function, but rather the exact price be specified by the function caller.

bh2smith:
> Ben, my understanding of getBuyAmt is that it's equivalent to getOutputAmountFromPrice up to renaming variables, so I didn't create a new function. Is there a difference I'm missing?

If there is already a function `getOutputAmountFromPrice`, the I might propose that our scripts start using this.

fleupold:
> To send unlimited orders you can append --orderSizeLimit=340282366920938463463374607431768211455 to the command you are running. I maintain that I don't see why any user would want to use this option ever.

This value is a magic number in our smart contract which makes it so that you will never have to replenish your order and including your order in a solution is slightly cheaper for the solver (which might as some point become a criteria in the economic viability).

But you are right that at the moment there is not a huge practical difference that incentivizes the liquidity provider (unless they also pay for a solver like us).

Leaving it up to you to decide what should be the default.

PR #225, opened 2020-05-07
[Hotfix] Checks not performed in parallel

bh2smith:
It was noticed that  The bracket safe verification "Ensure that safes are all owned solely by masterSafe" was taking a long time and not submitted in parallel. This fixes it.

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -122,13 +122,15 @@ module.exports = async (callback) => {
       console.log("==> Skipping safe deployment and using brackets safeOwners")
       bracketAddresses = argv.brackets
       // Ensure that safes are all owned solely by masterSafe
-      await Promise.all(
+      const masterNotOnlyOwner = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
-          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
-            callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
-          }
+          !(await isOnlySafeOwner(masterSafe.address, safeAddr))
         })
       )
+      const badSafes = masterNotOnlyOwner.filter((_, i) => masterNotOnlyOwner[i] == true)
+      if (badSafes.some((t) => t)) {
+        callback(`Error: Brackets ${badSafes.join()} is/are not owned (or at least not solely) by master safe ${masterSafe.address}`)
+      }
       // Detect if provided brackets have existing orders.
       const existingOrders = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -124,7 +124,7 @@ module.exports = async (callback) => {
       // Ensure that safes are all owned solely by masterSafe
       const masterNotOnlyOwner = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
-          !(await isOnlySafeOwner(masterSafe.address, safeAddr))
+          return !(await isOnlySafeOwner(masterSafe.address, safeAddr))
         })
       )
       const badSafes = masterNotOnlyOwner.filter((_, i) => masterNotOnlyOwner[i])
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -127,7 +127,7 @@ module.exports = async (callback) => {
           !(await isOnlySafeOwner(masterSafe.address, safeAddr))
         })
       )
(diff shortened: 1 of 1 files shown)


scripts/complete_liquidity_provision.js
```diff
@@ -122,13 +122,17 @@ module.exports = async (callback) => {
...
-      await Promise.all(
+      const masterNotOnlyOwner = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
-          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
-            callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
-          }
+          !(await isOnlySafeOwner(masterSafe.address, safeAddr))
         })
       )
+      const badSafes = masterNotOnlyOwner.filter((_, i) => masterNotOnlyOwner[i] == true)
```
josojo, line 130: diminutive warning: in js I would never use ==, but always ===. but here you can even leave it out.

fleupold: Is there a reason to leave it out here specifically?

josojo: With leaving out, I mean:
```
const badSafes = masterNotOnlyOwner.filter((_, i) => masterNotOnlyOwner[i])
```
But I am braindead right now...

scripts/complete_liquidity_provision.js
```diff
@@ -122,13 +122,17 @@ module.exports = async (callback) => {
       console.log("==> Skipping safe deployment and using brackets safeOwners")
       bracketAddresses = argv.brackets
       // Ensure that safes are all owned solely by masterSafe
-      await Promise.all(
+      const masterNotOnlyOwner = await Promise.all(
         bracketAddresses.map(async (safeAddr) => {
-          if (!(await isOnlySafeOwner(masterSafe.address, safeAddr))) {
-            callback(`Error: Bracket ${safeAddr} is not owned (or at least not solely) by master safe ${masterSafe.address}`)
-          }
+          !(await isOnlySafeOwner(masterSafe.address, safeAddr))
```
fedgiac, line 127: For this to work you need to return this value or better to remove the curly brackets, otherwise `masterNotOnlyOwner` is an array of `undefined`.

fedgiac:
I also don't see why this is needed.
I made a minimal working example of the code before this PR and everything works as expected (script time does not increase with the number of array entries):
```
const sleep = function (milliseconds) {
  return new Promise((r) => setTimeout(r, milliseconds))
}

const entries = 50
const tests = ",".repeat(entries).split(",")
tests.fill(false, 0, entries)
tests[entries] = true

async function main() {
  const start = Date.now()
  await Promise.all(
    tests.map(async (test, i) => {
      await sleep("1000")
      if (test) console.log("Found at index", i)
    })
  )
  console.log(`Finished in ${Date.now() - start}ms`)
}

main()
```

scripts/complete_liquidity_provision.js
bh2smith:
```suggestion
      const badSafes = masterNotOnlyOwner.filter((_, i) => masterNotOnlyOwner[i])
```

scripts/complete_liquidity_provision.js
bh2smith:
```suggestion
          return !(await isOnlySafeOwner(masterSafe.address, safeAddr))
```

bh2smith:
Thanks @fedgiac, I suppose this snippet of code demonstrates that the original code was indeed running in parallel. Lets just see if @fleupold is satisfied with this construction and we can forget about this PR. 

Although I do kinda like the way this is written over the previous.

bh2smith:
> Can you elaborate how this fixes this? I only see that this is moving a synchronous print message into an extra loop. How is this making things more parallelized?

From my understanding, I am moving an await from inside the greater promise out so that the individual entries are not being awaited one at a time. It is now clear to me from your question that I have not understood correctly and I guess I will have to go actually learn about asynchronous awaits.

It appears from @fedgiac's snippet that everything was fine in the first place... so then I'll asynchronously await someone to confirm that this was fine in the first place and that this PR can be closed.

fleupold:
>  Lets just see if @fleupold is satisfied with this construction

I'm fine with this construction and agree that this PR doesn't change anything. My original point was that providing brackets into the script makes it run for a very long time even on reasonably internet (making it impractical to use during development)

I was hoping you could look into and find out what is taking so long.

fedgiac:
While working on making requests in parallel in these scripts I've noticed that, if there are many queries (~20?) at the same time, it can happen that the running time still increases linearly with the number of queries. I didn't look into it since the effect was much smaller compared to the time it takes to the script to run if not parallelized. Maybe it's some rate limiting of the node, or maybe it's some part of the network request that cannot be run asynchronously?

bh2smith:
I am satisfied with @fedgiac's explanation here. Are we then in favour of closing or merging these changes?

PR #232, opened 2020-05-11
Fix duplicated "y" in user prompt

### LINKED ISSUE #189: Are you sure you want to send this transaction to the EVM? [yN] yy
For some reasons, user input on prompts repeats. Press "y" -> printed "yy".

bh2smith:
Changes here, move the user prompt into a more appropriate location which avoids being unavoidably imported multiple times in files that rely on it.

Closes #189 

### Test Plan:
Run this (with intentionally bad  values)

```sh
npx truffle exec scripts/complete_liquidity_provision.js --targetToken 1 --stableToken 4 --lowestLimit 300 --fleetSize 2 --highestLimit 416 --currentPrice 20 --masterSafe 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --investmentTargetToken 0.1 --investmentStableToken 1 --network rinkeby --brackets=0x725fd170B9a8c28caabB9b81Dd0dC80Fb3aa5Db6,0xeda61B1e9f414889ddE0dbe83B3BaCF814A98E87
```

and expect to get the following warnings along with their "proceed Anyway" prompts.

```
         chosen price: 20 USDC bought for 1 WETH
         dex.ag price: 188.5327344940625 USDC bought for 1 WETH
Price check failed! Continue anyway? [yN] y
Please double check your bounds. They seem to be unreasonable
Please double check your bounds. Current price is not within the bounds
Bound checks failed! Continue anyway? [yN] y
==> Skipping safe deployment and using brackets safeOwners
The following brackets have existing orders:
  0x725fd170B9a8c28caabB9b81Dd0dC80Fb3aa5Db6,0xeda61B1e9f414889ddE0dbe83B3BaCF814A98E87
 Continue anyway? [yN] y
```

### DIFF
--- a/scripts/bracket_orders.js
+++ b/scripts/bracket_orders.js
@@ -3,8 +3,8 @@ const { fetchTokenInfoFromExchange, getExchange, getSafe, buildOrders } = requir
   artifacts
 )
 const { isPriceReasonable, areBoundsReasonable } = require("./utils/price_utils.js")(web3, artifacts)
-const { signAndSend, promptUser } = require("./utils/sign_and_send")(web3, artifacts)
-const { proceedAnyways } = require("./utils/user_interface_helpers")
+const { signAndSend } = require("./utils/sign_and_send")(web3, artifacts)
+const { proceedAnyways, promptUser } = require("./utils/user_interface_helpers")
 
 const argv = require("./utils/default_yargs")
   .option("targetToken", {
--- a/scripts/utils/user_interface_helpers.js
+++ b/scripts/utils/user_interface_helpers.js
@@ -5,10 +5,12 @@ const promptUser = function (message) {
     input: process.stdin,
     output: process.stdout,
   })
-  return new Promise((resolve) => rl.question(message, (answer) => {
-    rl.close()
-    resolve(answer)
-  }))
+  return new Promise((resolve) =>
+    rl.question(message, (answer) => {
+      rl.close()
+      resolve(answer)
+    })
+  )
 }
 
 const proceedAnyways = async function (message) {
--- a/scripts/utils/user_interface_helpers.js
+++ b/scripts/utils/user_interface_helpers.js
@@ -1,11 +1,11 @@
 module.exports = function (web3 = web3, artifacts = artifacts) {
   const readline = require("readline")
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  })
 
   const promptUser = function (message) {
-    const rl = readline.createInterface({
-      input: process.stdin,
-      output: process.stdout,
-    })
     return new Promise((resolve) => rl.question(message, (answer) => resolve(answer)))
   }
 
--- a/scripts/utils/user_interface_helpers.js
+++ b/scripts/utils/user_interface_helpers.js
@@ -1,23 +1,22 @@
-module.exports = function () {
-  const readline = require("readline")
-  const rl = readline.createInterface({
-    input: process.stdin,
-    output: process.stdout,
-  })
+const readline = require("readline")
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+})
 
-  const promptUser = function (message) {
-    return new Promise((resolve) => rl.question(message, (answer) => resolve(answer)))
-  }
+const promptUser = function (message) {
+  return new Promise((resolve) => rl.question(message, (answer) => resolve(answer)))
+}
 
-  const proceedAnyways = async (message) => {
-    const answer = await promptUser(message + " Continue anyway? [yN] ")
-    if (answer === "y" || answer.toLowerCase() === "yes") {
-      return true
-    }
-    return false
+const proceedAnyways = async function (message) {
+  const answer = await promptUser(message + " Continue anyway? [yN] ")
+  if (answer === "y" || answer.toLowerCase() === "yes") {
+    return true
   }
-  return {
-    proceedAnyways,
-    promptUser,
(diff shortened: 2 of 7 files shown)


scripts/utils/user_interface_helpers.js
```diff
@@ -1,5 +1,13 @@
 module.exports = function (web3 = web3, artifacts = artifacts) {
-  const { promptUser } = require("./sign_and_send")(web3, artifacts)
+  const readline = require("readline")
+
+  const promptUser = function (message) {
+    const rl = readline.createInterface({
+      input: process.stdin,
+      output: process.stdout,
+    })
+    return new Promise((resolve) => rl.question(message, (answer) => resolve(answer)))
```
nlordell, line 9: My best guess is that once the promise resolves that maybe [`rl.close()`](https://nodejs.org/api/readline.html#readline_rl_close) might need to be called?

nlordell: Or, the same `rl` can be reused for all `promptUser` calls?

bh2smith: Would your second suggestion amount to moving the rl outside of the method. I think I will go  with your first suggestion though.

nlordell: They are just some ideas for getting it to work, I don't know enough about `readline` to know what the "correct" way to do it is.

bh2smith: Actually, it turns out I only had to move the 

```
  const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout,
  })
```

outside of the function declaration!

scripts/bracket_orders.js
```diff
@@ -3,8 +3,8 @@ const { fetchTokenInfoFromExchange, getExchange, getSafe, buildOrders } = requir
   artifacts
 )
 const { isPriceReasonable, areBoundsReasonable } = require("./utils/price_utils.js")(web3, artifacts)
-const { signAndSend, promptUser } = require("./utils/sign_and_send")(web3, artifacts)
-const { proceedAnyways } = require("./utils/user_interface_helpers")
```
fedgiac, line 7: I just realized this line should not be working in the first place since the file that is required exports a function and not a "generic" object. So why isn't destructuring failing here?

bh2smith: I am not sure I understand the question/problem here.... the file `./utils/user_interface_helpers` exports `proceedAnyways`.

scripts/utils/user_interface_helpers.js
```diff
@@ -1,5 +1,13 @@
 module.exports = function (web3 = web3, artifacts = artifacts) {
```
bh2smith, line 1: @fedgiac - I guess this is the place where we can remove the input parameters.

scripts/bracket_orders.js
bh2smith: Oh yes. I see what is happening here. Weird stuff that it was even working!

josojo:
Reading in the docs: https://nodejs.org/api/readline.html
```
Once this code is invoked, the Node.js application will not terminate until the readline.Interface is closed because the interface waits for data to be received on the input stream
```
I think that closing it would be a little bit cleaner. Why have you decided against it?

bh2smith:
> I think that closing it would be a little bit cleaner. Why have you decided against it?

I tried closing it, but since we use it again, the script just stopped.

I put the close right here.

https://github.com/gnosis/dex-liquidity-provision/blob/16a803590475f65adea61b9abe63066a23ac4af9/scripts/utils/user_interface_helpers.js#L7-L9

I don't know how or where is the appropriate place to close it and it works fine now as it is, so I don't see why we need to close it.

bh2smith:
> Looks good to me! Do you have any insights on why the import was working before? Has the original issue been fixed?

Yes, we had some weird async wrapper around the script before causing the entire script to be a single function. I guess this is some cheap hack we use to pass around the web3 and artifacts. 

In this script, since none of these items are needed, we can export/expose these functions in the usual fashion.

nlordell:
> I tried closing it, but since we use it again, the script just stopped.

What if you construct a new `readline.Interface` every time `promptUser` is called? Then you wouldn't be "reusing" the old one right?

>  I don't see why we need to close it.

I agree with @josojo here that the documentation indicates that proper use means that the `readline.Interface` should be closed when it is no longer used. I think it might be worth trying to understand why this isn't working as expected and what are the consequences of not closing it.

Also, an edge case that might cause issues: what happens in the `promptUser` is called twice? I can see this happening if we make a script that calls two separate sub-scripts that each call `promptUser`.

bh2smith:
> What if you construct a new readline.Interface every time promptUser is called? Then you wouldn't be "reusing" the old one right?

My comment on this was,

> "My initial thoughts were to move the rl into the promptUser function and close the rl right before returning. I believe I have tried this, but maybe without putting the rl declaration inside the promptUser"

> Also, an edge case that might cause issues: what happens in the promptUser is called twice? I can see this happening if we make a script that calls two separate sub-scripts that each call promptUser.

This is exactly the case provided in the PR's test plan.

bh2smith:
Ok @josojo  and @nlordell  I have gotten this to work as suggested (closing `rl` after each use) as follows:

```js
const promptUser = function (message) {
  const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout,
  })
  return new Promise((resolve) => rl.question(message, (answer) => {
    rl.close()
    resolve(answer)
  }))
}
```

let me know what ya'll think.

PR #233, opened 2020-05-11
Rework confusing comment

### LINKED ISSUE #126: Vague comment: Cant distinguish buy token from sell token.
This comment loses the meaning of what is the sell token and buy token in each of these orders. 
I would very much like to know which is which without having to read through everything again.

Does "trades standard_token against target_token" mean "sells standard_token for target_token"?

https://github.com/gnosis/dex-liquidity-provision/blob/fd01e236415e706bbe4b441c4dbf124072fc41df/scripts/utils/trading_strategy_helpers.js#L249-L252

bh2smith:
I was unable to determine what "trades A against B" meant so I made this comment more descriptive. Note that it was still unclear in the code what, exactly, was being bought and sold in each so I invoked the famous phrase "buy low, sell high" philosophy here. Assuming `targetToken` is the volatile token, then `stableToken` is the reference point. This means then then target token is being bought for the lower price and sold at the higher price.

**Test plan:** Existing unit tests.

Closes #126

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -261,8 +261,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
           targetToken.decimals,
           stableToken.decimals
         )
-        // While the first bracket-order trades standard_token against target_token, the second bracket-order trades
-        // target_token against standard_token. Hence the buyAmounts and sellAmounts are switched in the next line.
+        // While the first bracket-order sells targetToken for stableToken, the second sells stableToken for targetToken.
+        // Hence the buyAmounts and sellAmounts are switched in the next line.
         const [lowerBuyAmount, lowerSellAmount] = getUnlimitedOrderAmounts(
           lowerLimit,
           targetToken.decimals,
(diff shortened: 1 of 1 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -261,8 +261,8 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
           targetToken.decimals,
           stableToken.decimals
         )
-        // While the first bracket-order trades standard_token against target_token, the second bracket-order trades
-        // target_token against standard_token. Hence the buyAmounts and sellAmounts are switched in the next line.
+        // While the first bracket-order sells targetToken for stableToken, the second sells stableToken for targetToken.
```
fedgiac, line 264:
```suggestion
        // While the first bracket-order sells targetToken for stableToken, the second buys targetToken for stableToken at a lower price.
```

josojo:
Thanks for this PR, Ben. It was on my todo list. I actually wanted to rewrite how the function getUnlimitedOrderAmounts is used, to make it cleaner. But I guess your solution now is for sure good enough as well.

bh2smith:
Yes, @josojo  I am currently in the process of rewriting this function and it is NOT EASY AT ALL.

PR #236, opened 2020-05-11
stable to quote & target to base

### LINKED ISSUE #231: Replace stableToken with relativeToken
The term "stable token" can be misleading as it's not necessary for the stable token to be "stable" in any sense. The only difference between target token and stable token is that the price is always written in terms of amount of stable tokens against one target token. As proposed by Ben, we could use "relative token" instead.

bh2smith:
As discussed in the issue, we have chosen to with more commonly used language for our previously "target" and "stable" tokens. That is, we now use the more common, "quote" and "base" currencies for which "quote" is the one whose price is specified in relation to a "base" token. See [investopedia](https://www.investopedia.com/terms/b/basecurrency.asp) for discussion on the naming convention.

Note also that the term "investment" has been replaced with "deposit" and we now specify `TokenId` instead of just `Token` in the script arguments.

Closes #231

Test Plan:

In our case, since the brackets were originally meant to be between ETH and DAI, since we are always "quoting" the price of ETH in units of DAI, please see a few relevant changes in the comments that explicitly mention ETH and DAI to confirm that the changes made are appropriate. 

Note that "target" -> "base" and "stable" -> "quote"

Also, unit tests should be passing.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -47,7 +47,7 @@ Making the requests to the gnosis-interfaces does not cost any gas. However, sig
 Here is an example script invocation:
 
 ```js
-npx truffle exec scripts/complete_liquidity_provision.js --baseToken=1 --quoteToken=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --depositBaseToken=10 --depositQuoteToken=1000 --fleetSize=10 --network=$NETWORK_NAME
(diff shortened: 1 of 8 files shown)


bh2smith:
I will do both of these suggestions.

bh2smith:
@josojo  I just realized that the readme must not contain an example of `complete_liquidity_provision` usage because the `depositBaseToken` is not used anywhere in the readme and its a demanded argument.

PR #237, opened 2020-05-12
[Easy] Rework unlimited order function usage

bh2smith:
Previously, we were taking advantage of some non-intuitive symmetric properties emitted by the `unlimitedOrderAmounts` function that were difficult to interpret. This PR simply uses the function directly as was intended rather than assuming any prior knowledge about the function's "special properties".


bh2smith:
I was also struggling to make it work, but I was forgetting to swap the tokens. I had swapped the output and used `1 / price`, but without the tokens swapped it just wasn't working! Thanks to @fedgiac for pointing that out for me!

PR #240, opened 2020-05-12
[Trivial] Replace lost console log

bh2smith:
This log statement is valuable and had recently been removed.

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -153,6 +153,7 @@ module.exports = async (callback) => {
       assert(!argv.verify, "Trading Brackets need to be provided via --brackets when verifying a transaction")
       console.log(`==> Deploying ${argv.fleetSize} trading brackets`)
       bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize)
+      console.log("List of bracket traders in one line:", bracketAddresses.join())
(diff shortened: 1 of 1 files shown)


scripts/complete_liquidity_provision.js
```diff
@@ -153,6 +153,7 @@ module.exports = async (callback) => {
       assert(!argv.verify, "Trading Brackets need to be provided via --brackets when verifying a transaction")
       console.log(`==> Deploying ${argv.fleetSize} trading brackets`)
       bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.fleetSize)
+      console.log("List of bracket traders in one line:", bracketAddresses.join())
```
fleupold, line 156: I was under the impression it's not valuable as the list of brackets in one line will also be printed again at the end of the run when we print the verify instructions.

bh2smith: I don't believe this is the case at all. Also, if the script fails halfway through, one will have deployed safes without knowing the addresses and have to redeploy (or at least dig to find them).

bh2smith: This scenario I describe is exactly what happened to me yesterday. I wound up deploying 60 safes without even knowing because I was not provided with any success statement before the latter code failed.

fleupold: 🙆‍♂️

PR #242, opened 2020-05-13
Allow uneven amount of brackets

### LINKED ISSUE #234: Allow use of odd number of brackets
I think there is no reason anymore to only use an even number of brackets. The checks with this purpose could be safely removed.
It would help users that want to deploy liquidity only with a single Safe.

josojo:
closes #234 

test plan: there is no logic depended on whether we have an even amount of brackets.  Hence, no tests are needed. It's really just a relic of the past.


bh2smith:
> test plan: there is no logic depended on whether we have an even amount of brackets. Hence, no tests are needed. It's really just a relic of the past.

By test plan, I meant provide a command I can run to try this.

josojo:
>I'd like to see the scripts running on Rinkeby with 1 bracket or 19 just to be sure everything is right.

What would you check then on the rinkeby instances? everything? I added some tests to ganache, it increases the travis run-time, but at least here we can check efficiently want is happening.

> By test plan, I meant to provide a command I can run to try this.

Since there is no single logic related to the evenness of the size, I can not provide a specific test. I just added a bunch of test cases and amend some tests

PR #245, opened 2020-05-13
Testing order expiry of buildOrders

### LINKED ISSUE #123: validFrom and expiry parameters of buildOrders are not tested
Only the default values are checked.

josojo:
closes #123 

note that customized validFrom is no longer an option in the code.

### DIFF
--- a/test/dfusion_multi_safes.js
+++ b/test/dfusion_multi_safes.js
@@ -524,6 +524,39 @@ contract("GnosisSafe", function (accounts) {
         assert.equal(buyOrder.validFrom, currentBatch)
       }
     })
+    it("Places bracket orders and checks explicitly the expiry", async () => {
+      const masterSafe = await GnosisSafe.at(await deploySafe(gnosisSafeMasterCopy, proxyFactory, [safeOwner.account], 1))
+      const bracketAddresses = await deployFleetOfSafes(masterSafe.address, 6)
+      const baseToken = 0 // ETH
+      const quoteToken = 1 // DAI
+      const lowestLimit = 90
+      const highestLimit = 120
+      await prepareTokenRegistration(accounts[0], exchange)
+
+      await exchange.addToken(testToken.address, { from: accounts[0] })
+
+      const customExpiry = (await exchange.getCurrentBatchId.call()).toNumber() + 50
+      const transaction = await buildOrders(
+        masterSafe.address,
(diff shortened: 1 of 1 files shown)


bh2smith:
The new test appears to be failing:

```
  1 failing

  1) Contract: GnosisSafe

       bracket order placement test:

         Places bracket orders and checks explicitly the expiry:

      Got [object Object]

      + expected - actual

      -4294967294

      +5297946
```

test/dfusion_multi_safes.js
```diff
@@ -524,6 +524,39 @@ contract("GnosisSafe", function (accounts) {
...
+      const bracketAddresses = await deployFleetOfSafes(masterSafe.address, 6)
+      const baseToken = 0 // ETH
+      const quoteToken = 1 // DAI
+      const lowestLimit = 90
+      const highestLimit = 120
+      await prepareTokenRegistration(accounts[0], exchange)
+
+      await exchange.addToken(testToken.address, { from: accounts[0] })
+
+      const customExpiry = (await exchange.getCurrentBatchId.call()).toNumber() + 50
```
fedgiac, line 538: Is there a reason why you don't do `exchange.getCurrentBatchId()` but you add `.call()`?

test/dfusion_multi_safes.js
```diff
@@ -524,6 +524,39 @@ contract("GnosisSafe", function (accounts) {
...
+      await execTransaction(masterSafe, safeOwner.privateKey, transaction)
+
+      // Correctness assertions
+      for (const bracketAddress of bracketAddresses) {
+        const auctionElements = exchangeUtils.decodeOrdersBN(await exchange.getEncodedUserOrders(bracketAddress))
+        assert.equal(auctionElements.length, 2)
+        const [buyOrder, sellOrder] = auctionElements
+
+        assert.equal(buyOrder.validUntil, customExpiry, `Got ${sellOrder}`)
+        assert.equal(sellOrder.validUntil, customExpiry, `Got ${sellOrder}`)
```
fedgiac, line 558: The printed message is always `Got [object Object]`, also in the first line it should be `buyOrder` instead of `sellOrder`.

test/dfusion_multi_safes.js
josojo: since it is a view function, it does not make a difference.

PR #246, opened 2020-05-13
Checks for duplicates in --brackets

### LINKED ISSUE #188: Enforce unique input values in command-line options taking comma-separated lists
Some scripts take comma-separated list of values, most notably lists of brackets. In most cases, these values are supposed to be unique, and this should be enforced.

josojo:
closes #188 

testplan:
```
npx truffle exec scripts/bracket_orders.js --masterSafe $MASTER_SAFE --baseTokenId 1 --quoteTokenId 7 --brackets 0xbe3b1a11194f09cd471bb4Ab620807EF58cA4198,0xbe3b1a11194f09cd471bb4Ab620807EF58cA4198 --currentPrice 4 --network rinkeby 
```
and observe the error


scripts/utils/user_interface_helpers.js
```diff
@@ -13,6 +13,12 @@ const promptUser = function (message) {
   )
 }
 
+function checkNoDuplicatedBracket(argv) {
+  if (new Set(argv.brackets).size !== argv.brackets.length)
+    throw new Error("the parameter --brackets is not supposed to have duplicated entries")
+  return true
+}
+
```
bh2smith, line 21: This is cool, but I wonder if it could be made more generic like just `checkNoDuplicates` and also be used on other lists that might be provided to our scripts.

bh2smith: Also, I would prefer if this function was located in the `default_yargs` file that is more geared toward code that involved the arguments.

fedgiac: I'd also like to see a `checkNoDuplicates`, this function could be used in `script/withdraw.js` for the arguments `tokens` and `tokenIds`, for example.

fedgiac:
Is the code still working when `--brackets` is not set? I'd expect the line:
```
   if (new Set(argv.brackets).size !== argv.brackets.length) 
```
to throw since `undefined` has no property `.length` .

scripts/utils/default_yargs.js
```diff
@@ -10,3 +10,18 @@ module.exports = require("yargs")
...
+
+const checkNoDuplicate = function (array) {
+  return new Set(array).size !== array.length
+}
+
+function checkBracketsForDuplicate(argv) {
+  if (argv.brackets && checkNoDuplicate(argv.brackets))
+    throw new Error("the parameter --brackets is not supposed to have duplicated entries")
+  return true
+}
```
bh2smith, line 22: This is looking a lot better, but wonder if we could get away without a dedicated single use function like `checkBracketsForDuplicate`. For example, the more general function could 

```js 
return array && new Set(array).size !== array.length
```

but then the error message would have to go up in the `.check(checkBracketsForDuplicate).argv` I would imagine that `.check` since it is apparently a feature of yargs already has a way to include custom error message upon failure.

josojo: I checked the documentation and could not find a custom error message for .check

See also here for a search:
https://stackoverflow.com/questions/38419991/yargs-check-error-handling

josojo: Also, I don't wanna write the same error msg into all script files, which have the parameter brackets

PR #258, opened 2020-05-15
Sign with web3-accounts

### LINKED ISSUE #235: Scripts should gracefully fail when private key is not available
Running for example `complete_liquidity_provision.js` when the private key has not be added as an env variable makes the script run for a while and then fail towards the end with:
```
TypeError: Cannot read property 'startsWith' of undefined
    at withHexPrefix (/path/dex-liquidity-provision/scripts/utils/sign_and_send.js:18:45)
    at signAndSend (/path/dex-liquidity-provision/scripts/utils/sign_and_send.js:66:24)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
All scripts that need a private key should tell the users that no private key is available.

bh2smith:
We recently cut ties with our dependency on `lightwallet` in PR #200, but at the expense of signing via privatekey using

https://github.com/gnosis/dex-liquidity-provision/blob/baa915d5a17cc13ee8eabd2ba9ce1b717ccd65a2/scripts/utils/sign_and_send.js#L62-L66

https://github.com/gnosis/dex-liquidity-provision/blob/baa915d5a17cc13ee8eabd2ba9ce1b717ccd65a2/scripts/utils/internals.js#L83-L87

This was causing very bizarre and inconsistent problems with the `complete_liquidity_provision.js` script (for example) since there are several steps, some of which using the web3 account (like `deploy_safes`) and others using `sign_and_send` near the end of the script.

Luckily, `web3` provides a signing tool that allows to pass the unlocked accounts specified in truffle config [web3.eth.sign](https://web3js.readthedocs.io/en/v1.2.1/web3-eth.html#sign)

Ideally, this PR consolidates these accounts and localizes the private key juggling back in truffle config. Still working on a test plan, but I would assume that running the complete liquidity provision script with and without `PK` in `process.env` would be a good indication that this is working as intended.

 Closes #235

Test Plan: Unit test should be passing and the following script execution should also run smoothly:

```sh
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=190 --highestLimit=210 --currentPrice=202 --masterSafe=0x8990c564ec303C7b26d3d5556ef0910E58Be08Ce --depositBaseToken=0.1 --depositQuoteToken=10 --numBrackets=2 --network=rinkeby
```


bh2smith:
The problem we are now facing is that the signing proposed here does not agree with the signature we need:


```js
    function signHashWithPrivateKey(hash, privateKey) { 
      const msgBuff = new Buffer(ethUtil.stripHexPrefix(hash), "hex") 
      const sig = ethUtil.ecsign(msgBuff, new Buffer(ethUtil.stripHexPrefix(privateKey), "hex")) 
      return "0x" + sig.r.toString("hex") + sig.s.toString("hex") + sig.v.toString(16) 
    }
    // const signer = "0x2a9253b16d875bb8a7b9dc5304d35e5d2945c8f7"
    const privateKey = "0xee2d012ef73cfe974d8b9351a068304137ab8c6268550d44fc54a03da2ff47cd"
    const account = web3.eth.accounts.privateKeyToAccount(privateKey) 
    console.log("Account from PK agrees with signer?", account.address === signer)

    const sigs = await web3.eth.sign(transactionHash, signer)
    const x = signHashWithPrivateKey(transactionHash, privateKey)
    console.log(x)
    console.log(sigs)
```


Apparently, `web3.eth.sign` prepends the string "ethereum Message:" before the given message. This is crap.

fleupold:
> Apparently, web3.eth.sign prepends the string "ethereum Message:" before the given message. This is crap.

This should be supported by the Safe contract: https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafe.sol#L251

Maybe @rmeissner or @Uxio0 can give you a pointer of what we need to do in order to use the native `eth_sign` to sign Safe transactions proposals.

scripts/utils/sign_and_send.js
```diff
@@ -59,11 +54,9 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
     }
 
-    assert(process.env.PK != null, "This script requires a private key be explicitly provided. Please export PK")
-    const privateKey = withHexPrefix(process.env.PK)
-    const account = web3.eth.accounts.privateKeyToAccount(privateKey)
-    console.log(`Signing and posting multi-send transaction ${transactionHash} from proposer account ${account.address}`)
-    const sigs = signHashWithPrivateKey(transactionHash, privateKey)
+    const signer = (await web3.eth.getAccounts())[0]
+    console.log(`Signing and posting multi-send transaction ${transactionHash} from proposer account ${signer}`)
+    const sigs = web3.eth.sign(transactionHash, signer)
```
fleupold, line 59: This needs an await doesn't it?

Uxio0:
> Maybe @rmeissner or @Uxio0 can give you a pointer of what we need to do in order to use the native `eth_sign` to sign Safe transactions proposals.

I had a debate regarding this: https://github.com/ethereum/eth-account/issues/85

In Python using the `eth-acount` library you can still sign raw hashes

scripts/utils/sign_and_send.js
bh2smith: OH man! Is that the problem? Yea, it does need await, but I don' think I forgot it in the other location where this functionality is tested. The new problem I am facing is a `require` statement in Gnosis Safe

```
// Hashes are automatically approved by the sender of the message or when they have been pre-approved via a separate transaction
require(msg.sender == currentOwner || approvedHashes[currentOwner][dataHash] != 0, "Hash has not been approved");
```

scripts/utils/internals.js
```diff
@@ -76,16 +76,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       0,
       ADDRESS_0,
       ADDRESS_0,
-      sigs
+      [sigs]
```
bh2smith, line 79:
```suggestion
      sigs
```

scripts/utils/sign_and_send.js
bh2smith:
```suggestion
    const sigs = await web3.eth.sign(transactionHash, signer)
```

bh2smith:
Turns out, the following Safe Contract Documentation and @rmeissner saved the day!

https://docs.gnosis.io/safe/docs/contracts_signatures/ (cf eth_sign).

scripts/utils/sign_and_send.js
```diff
@@ -78,7 +71,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
       refundReceiver: ZERO_ADDRESS,
       nonce: nonce,
       contractTransactionHash: transactionHash,
-      sender: web3.utils.toChecksumAddress(account.address),
+      sender: web3.utils.toChecksumAddress(signer.address),
```
bh2smith, line 74:
```suggestion
      sender: web3.utils.toChecksumAddress(signer),
```

scripts/utils/sign_and_send.js
```diff
@@ -1,17 +1,13 @@
 module.exports = function (web3 = web3, artifacts = artifacts) {
-  const assert = require("assert")
   const axios = require("axios")
-
-  const { signHashWithPrivateKey } = require("../utils/internals")(web3, artifacts)
   const { ZERO_ADDRESS } = require("./constants")
+  const { getSafeCompatibleSignature } = require("./internals")
```
fedgiac, line 4: `internals.js` requires `(web3, artifacts)` to be specified, otherwise you'd get
```
TypeError: getSafeCompatibleSignature is not a function
```
when running the script.

```suggestion

  const { getSafeCompatibleSignature } = require("./internals")(web3, artifacts)
  const { ZERO_ADDRESS } = require("./constants")
```
I also added a newline for consistency with the imports in other files, it separates internal imports to our project from external imports, like `BN`.

bh2smith:
@fedgiac, the error you are experiencing is precisely what was happening earlier when I was speaking with @rmeissner. I have verified that the `signer` is simply not the claimed sender from the error. This seems to be a problem with the post to safe backend. There is nothing I can do here in this script to fix this as the address `Sender=0x627306090abaB3A6e1400e9345bC60c78a8BEf57` seems to appear out of thin air. 

Furthermore, this error appears to be completely inconsistent. That is, even if we try to ensure that signer is an owner of the safe in question, this error will still come up completely out of the blue.

I think the best approach here is to figure out what is happening in the safe backend services that somehow gets derived the wrong signer address from the signature. I wonder if maybe this has something to do with the `modifiedSig` (i.e. since there are those two distinct cases adding "f1" and "20").

fleupold:
As discussed, this issue seems bad enough to be blocking the merge for now.

> Furthermore, this error appears to be completely inconsistent.

Have you tried printing the raw requests and checking if it is at least consistent for the same POST request? Maybe we can then use one of the failing requests to implement a regression test in the relayer service.

bh2smith:
Ok, so it turns out there are actually 4 scenarios we can expect for the last two characters of a signature. We now consider all four cases and print a log containing the last two characters being modified for the safe compatible signature. If they are ever different, we should catch it in the log!

scripts/utils/internals.js
```diff
@@ -170,10 +170,17 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
     // signature standards for Gnosis Safe execTransaction
     // https://docs.gnosis.io/safe/docs/contracts_signatures/
-    const modifiedSigs = sigs.slice(0, -2) + (sigs.slice(-2) === "00" ? "1f" : "20")
+    const nodeMap = {
+      "00": "1f",
+      "1b": "1f",
+      "01": "20",
+      "1c": "20",
+    }
+    const modifiedSigs = sigs.slice(0, -2) + nodeMap[sigs.slice(-2)]
```
fleupold, line 183: Can you explain a bit more what you are doing here? According to [the docs](https://docs.gnosis.io/safe/docs/contracts_signatures/#span-stylecolordb3a3deth_signspan-signature) we are supposed to add the number 4 to the last byte. Why are you creating a node map and why does it have exactly four entries?

bh2smith: I can only tell you to the best of my understanding about the matter which is, admittedly, very little. I have gotten most of these "facts" through @rmeissner (who may be able to expand more) and a [stackexchange post](https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v) about the value of `v` in an ECDSA signature. The recovery byte in an ECDSA signature can generally be any of 0, 1, 2 or 3. Ethereum apparently restricts itself to the "upper curve" which is then only 0 or 1. However, this also depends on which type of signing we are doing (i.e. `personal_sign` or `eth_sign`) which, in one case adds 27 to the last byte in one of these cases.

This means then that there are 4 possiblities for the last byte in a signature returned from `web3.eth.sign` (0, 1, 27 or 28) which, in hex, are the keys of this `nodeMap` shown here (should probabily consider a different name for this). I was confused by @rmeissner earlier and had the feeling these values has something to do with which types of nodes were constructing the transaction hash.

Anyway, the result we are actually looking for then is `(v % 27) + 4` which is hex-equivalent to the evaluation depicted in this mapping seen here.

bh2smith: I will run some simulations today to ensure what the possible outcomes actually happen to be. I have a suspicion that we may only ever encounter 2 of these acclaimed 4 bytes.

bh2smith: I ran this snippet in `truffle console --rinkeby`

```js
const generateRecoveryBytes = async function(n) {
  let S = new Set()
  for (let i = 0; i < n; i++) { 
    let sig = await web3.eth.sign(i, accounts[0])
    S.add(sig.slice(-2))
  } 
  return S
}
```

with `n = 1000` and the only values I am getting back are `1c`

```
truffle(rinkeby)> res = await generateRecoveryBytes(1000)
undefined
truffle(rinkeby)> res
Set { '1c' }
```

bh2smith: Then I tried two more times with 
```js
const sign = async function(msg) { return (await web3.eth.sign(msg, accounts[0])).slice(-2) }
```

and got:

```
truffle(rinkeby)> sign("some word")
'1c'
truffle(rinkeby)> sign("some word other")
'1b'
```


My new claim is that only "1b" and "1c" will be possible outcomes here.

rmeissner: The returned values depend on your node and the method of signing. As mentioned above ECDSA signatures can have 2 values for `v` when created for Ethereum: 27 and 28 (in hex '1b' or '1c'). This is the recId + 27. The recId can be 0 or 1. When using `eth_sign` (which web3.eth.sign should do) then the '+27' **should** be included. When using `personal_sign` then it **should** be the plain recId. But this might slighly differ on your js lib and on the node you are using. That is why I proposed to handle all 4 cases.

rmeissner: Looking at the `web3js` docs example you see that they also return `v` without the '+ 27' for `eth_sign`: https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#sign ... so yeah ... testing that locally doesn't really represent what you find in the wild ;)

bh2smith: Then I will revert my recent commit and put back the mapping. I am getting kinda sick of this PR and would like to get it merged. If there are no more questions, this would be great.

bh2smith: If desired, I could put some of the "facts" stated in this thread as part of the comments here next to the mapping definition. Have any of the statements made here been helpful to further  understanding @fedgiac and/or @fleupold?

fleupold: I would prefer if instead of using the lookup map we actually perform the computation `v + 4` and link to the safe docs that tell us to do so. If we must, we can have a 

```
if (v === 0 || v === 1) {
  v += 27
}
```

before, but we should link a concrete library/system that doesn't work as expected to justify this clown-town.

I'm still confused why it sometimes would be 0,1.
We are only using eth_sign here. I cannot find a link to actually say 27, 28 is the standard, but if it is implementations that don't comply with it should be filed a bug and fixed. If it's not the forced standard, I'd argue that the safe backend should be able to handle values 4 and 5 as well.

Locally, Web3 v1.2.1 gives me

```
await web3.eth.sign("Hello world", accounts[0])
0x...1c
```

indicating that the web3js example is outdated. Also it looks like parity and geth both return 27, 28 (https://github.com/ethereum/go-ethereum/issues/2053, https://github.com/openethereum/openethereum/issues/5490)

Also, to better understand what is happening, can you please explain the flaky behavior you were seeing? On the same machine two consecutive runs of the code would sometimes work and sometimes not work even when adding 4 to either both 0,1 or both 27,28? Maybe you can share the commit hash that was showing the flaky behavior.

I realise you are a bit worn out by this task. I believe there is a ton of interesting historical/technical knowledge to learn here and we should try to solve this in a proper way.

fedgiac: My two cents: there are two parties in our scripts that take care of `web3.eth.sign`: either `ganache-cli` or `@truffle/hdwallet-provider`. I believe that the former returns signatures without adding `27` while the latter adds `27`. This can be tested as follows:
- execute `npx truffle console` while Ganache running and then type into the interactive console `web3.eth.sign("test", accounts[0])`. Output will be something like this, note it ends with `01`:
```'0xa3f00fbe4aa835556fa98ce468c4bd0cc91b3fc5c0b1fd1b8945821b310b6f8c614c0943a06fbc76dc134af867fa2e962cd24ea253ec72e2ff53394056bfac3f01'```
(This also works with a new Truffle project, so it's not an artefact of `"@gnosis.pm/util-contracts/src/util/truffleConfig"`, which lies between us and whatever web3 implementation we use.)
- export some private key and do the same for Rinkeby:
 `export MNEMONIC="candy maple cake sugar pudding cream honey rich smooth crumble sweet treat"`
 `npx truffle console --network=rinkeby`
 `web3.eth.sign("test", accounts[0])`
 Output will be this, note it ends with `1b` (27):
```'0x738aa850619814422cac75821422e6b08f5d378485b530ea6d02ebbfef55c050441ca7d66998d970c9e956237a46fce8f588ed0e745588a24781c53258c1d3141b'```

We test with Ganache but we work with `hdwallet-provider` when we aren't using the development network, so it appears we need to take into account both signature versions.

Comment (bh2smith) [reply scripts/utils/internals.js]: This is really good stuff all around. My proposal for implementation then would be that we use 
```js
let v = parseInt(sig.slice(-2), 16) // Will always be 0, 1, 27 or 28
const modifiedRecoveryByte = (v % 27) + 27 + 4 // will always be 31 or 32.
return sig.slice(0. -2) + modifiedRecoveryByte.toString(16)
```
which will always end in "1f" or "20". 

It is my impression that this nasty little detail would more naturally be managed directly in safe transaction services, but I am happy to handle it here for the time being.

Is there some way we could just settle on this temporary solution and get the 3 or 4 other blocked PRs merged before pursuing this any further. Otherwise, I am afraid they will become outdated, lost and possibly accidentally merged before they should (since github doesnt seem to have a way to prevent things frmo being merged in a certain order).

fedgiac: I would be in favor of Felix's more aggressive solution of doing something only if `v===0` or `v===1`, but also this works. There should also be a lot of explanatory comments. Thanks for all the work, by the way, I see it's a very annoying bug.

By the way, I think the Safe transaction service is doing the right thing here, they shouldn't be making guesses to match different ECDSA implementations. If there is a party I'd blame it's possibly Ganache, since it appears to be standard practice to implement `web3.eth.sign` with the ECDSA implementation from the Ethereum yellow paper and [by definition `v` is supposed to be either 27 or 28](https://github.com/ethereum/yellowpaper/blob/7e819ec24cf397a5f0aaf52f00b21702eca78d0a/Paper.tex#L1721).

fleupold: Nice, that was some really good 🕵🏼‍♂️ work, especially finding the place in the yellow paper where this is specified!

Can we file an issue in ganache-cli and linking both the yellow paper as well as that issue in the code here to comment our code?

bh2smith: I will make all of these great changes (mostly to get the issue moving on ganache), but I would like to point out the fact that this method is never used in the ganache testing environment because it is related to a POST to safe transaction services. In fact we even have other hard coded facts within this location that network must be rinkeby or mainnet.

This means we will never ever enter the `if v in {0, 1}` branch of the code.

rmeissner: To be fair the mentioned section in the yellowpaper is `Appendix F. Signing Transactions` and we are not signing a transaction here. Signing messages is not part of the yellow paper (I would love if it is), but therefore you get inconsistent behaviour. Please still try to push the issue on ganache :D, but be aware that `personal_sign` also returns `v` in `{0, 1}`. So saying that it does not comply to the real world scenario (metamask, geth and openethereum) might be more convincing

bh2smith: Yes, This is excellent. I have made almost all the appropriate changes to our code, but still waiting on the issue creation in ganache-cli. I wanted to let @fedgiac have the open-source cred here along with the added bonus of possibly submitting a fix!

That is to say, I have included the if branch (for cases v = 0 or 1) with a comment and link to the yellow paper as well. Will include a link to the ganache issue when it becomes available.

Note that trufflesuite (the maintainer of ganache) requests that we first ask about the issue in [their gitter channel](https://gitter.im/ConsenSys/truffle) before filing it with github.

bh2smith: Ok, at least for today. I have raised the issue in their gitter to see if they would even consider this as an issue. Hopefully someday, this PR and all those depending this will find their place on master.

scripts/utils/internals.js
```diff
@@ -174,15 +168,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     return web3.utils.padLeft(await web3.eth.getStorageAt(safeAddress, fallbackHandlerStorageSlot), 40)
   }
 
+  const getSafeCompatibleSignature = async function (transactionHash, signer) {
+    const sig = await web3.eth.sign(transactionHash, signer)
+    console.log(`Modifying recovery byte (${sig.slice(-2)}) for compatibility with Gnosis Safe Transaction Services`)
```
fleupold, line 173: Is this log statement still needed? It feels like this should be a comment rather than a print statement

scripts/utils/internals.js
```diff
@@ -174,15 +168,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+      // https://github.com/ethereum/yellowpaper/blob/7e819ec24cf397a5f0aaf52f00b21702eca78d0a/Paper.tex#L1721-L1726)
+      // However, ganache-cli: 6.9.1, is currently returning 0 or 1
+      v += 27
+    }
+    // The following signature manipulation is according to
+    // signature standards for Gnosis Safe execTransaction
+    // https://docs.gnosis.io/safe/docs/contracts_signatures/
+    const recoveryByte = v + 4
+    const modifiedSig = sig.slice(0, -2) + recoveryByte.toString(16)
+    return modifiedSig
```
fleupold, line 186: ubernit: no need for the modifiedSig variable.

scripts/utils/internals.js
bh2smith: Right. No longer necessary. I had this here so we might be able to detect if there were other values we didn’t account for. This was essentially my hack for not understanding what could happen.

PR #267, opened 2020-05-19
Move constants to own file

### LINKED ISSUE #223: Use single file for constants
We use many constants in our code, like `bnMaxUint256`, `CALL`, `ADDRESS_0` and more.
They are currently scattered around the repo. Sometimes we import a file just for the constants, and this can be bad because when we import some of the packages we also fetch something from the network, like the exchange. (Related issue #195.)
I propose to start defining our new constants in a file `scripts/utils/constants.js` and eventually move every existing constant there.

fedgiac:
Moves many of the constants defined in this repo into a single constant file.
Note that these constants are used for both scripts and tests, so if we get a constant wrong then the tests may not catch the ensuing errors.

I did not touch constants defined for the Synthetix script.
Note that this PR fixes a small bug that causes orders created with the order creation script to not have the same expiration date as those created with the complete liquidity provision script.
 
Closes #223.

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -15,7 +15,7 @@ const { proceedAnyways } = require("./utils/user_interface_helpers")
 const { toErc20Units } = require("./utils/printing_tools")
 const { sleep } = require("./utils/js_helpers")
 const { verifyBracketsWellFormed } = require("./utils/verify_scripts")(web3, artifacts)
-const { DEFAULT_NUMBER_OF_SAFES } = require("./utils/constants")
+const { DEFAULT_NUM_SAFES } = require("./utils/constants")
 const { default_yargs, checkBracketsForDuplicate } = require("./utils/default_yargs")
 
 const argv = default_yargs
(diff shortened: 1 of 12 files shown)


scripts/utils/constants.js
```diff
@@ -1,5 +1,41 @@
+const BN = require("bn.js")
+
 const ZERO_ADDRESS = "0x" + "0".repeat(40)
 
+// We chose the maximum admissible value minus one in order to distinguish orders from "non-expiring" orders of the web interface
+const DEFAULT_ORDER_EXPIRY = 2 ** 32 - 2
```
bh2smith, line 6: Looks like we are using minus two but the comment say minus 1.

scripts/complete_liquidity_provision.js
```diff
@@ -15,8 +15,9 @@ const { proceedAnyways } = require("./utils/user_interface_helpers")
 const { toErc20Units } = require("./utils/printing_tools")
 const { sleep } = require("./utils/js_helpers")
 const { verifyBracketsWellFormed } = require("./utils/verify_scripts")(web3, artifacts)
-
+const { DEFAULT_NUMBER_OF_SAFES } = require("./utils/constants")
```
bh2smith, line 18: nit: could we call this `DEFAULT_NUM_SAFES`... I have this thing with putting "connector words" into variable names (i.e. `OF`)

PR #269, opened 2020-05-20
JSDOC enforcement.

### LINKED ISSUE #108: Outdated JSDoc
The JSDoc is sometimes outdated, for example that of `buildOrders`.

bh2smith:
This turned out to be quite a challenge.

Still trying to make a typedef file that has all the types so they can be shared among other files while simultaneously importing them without getting other lint warnings. Also notice that one can cheat the linter by using `const z = function(x, y)` instead of `function z(x, y)`

Closes #108


package.json
```diff
@@ -57,4 +58,4 @@
     "truffle": "^5.1.26",
     "truffle-hdwallet-provider": "^1.0.0-web3one.1"
   }
-}
+}
```
fedgiac, line 61: I thought we wanted newlines at the end of the files?
```suggestion
}

```

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+ * @typedef Deposit
+ * @type {object}
+ * @property {number} amount integer denoting amount to be deposited
+ * @property {Address} tokenAddress {@link Address} of token to be deposited
+ * @property {Address} bracketAddress address of bracket into which to deposit
+ */
+
+/**
+ * @typedef Withdrawal
+ *  * Example:
```
fedgiac, line 39:
```suggestion
 * Example:
```

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+ * }
+ * @type {object}
+ * @property {number} amount Integer denoting amount to be deposited
+ * @property {Address} bracketAddress Ethereum address of the bracket from which to withdraw
+ * @property {Address} tokenAddresses List of tokens that the traded wishes to withdraw
+ */
+
+/**
+ * @typedef TokenObject
+ *  * Example:
```
fedgiac, line 53:
```suggestion
 * Example:
```

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+ * @type {object}
+ * @property {string} symbol symbol representing the token
+ * @property {(number|BN)} decimals number of decimals of the token
+ * @property {Address} address address of the token contract on the EVM
+ * @property {object} instance an instance of the token contract
+ */
+
+/**
+ * @typedef Transaction
+ *  * Example:
```
fedgiac, line 69:
```suggestion
 * Example:
```

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+ * Example:
+ * {
+ *   amount: 100,
+ *   tokenAddress: 0x0000000000000000000000000000000000000000,
+ *   bracketAddress: 0x0000000000000000000000000000000000000001
+ * }
+ *
+ * @typedef Deposit
+ * @type {object}
+ * @property {number} amount integer denoting amount to be deposited
```
fedgiac, line 32: This should be a string or a BN.

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @typedef Withdrawal
+ *  * Example:
+ * {
+ *   amount: "100",
+ *   bracketAddress: "0x0000000000000000000000000000000000000000",
+ *   tokenAddress: "0x0000000000000000000000000000000000000000",
+ * }
+ * @type {object}
+ * @property {number} amount Integer denoting amount to be deposited
```
fedgiac, line 46: Same here, string or BN.

scripts/typedef.js
```diff
@@ -0,0 +1,81 @@
...
+ *  {
+ *    operation: CALL,
+ *    to: "0x0000..000",
+ *    value: "10",
+ *    data: "0x00",
+ *  }
+ * @type {object}
+ * @property {number} operation Either CALL or DELEGATECALL
+ * @property {Address} to Ethereum address receiving the transaction
+ * @property {string} value Amount of ETH transferred
```
fedgiac, line 79: Since I commented before that we could use BN and strings for deposit and withdrawal amounts, if you go for both variants then you could also include BN here.

scripts/utils/default_yargs.js
```diff
@@ -15,6 +15,12 @@ const checkNoDuplicate = function (array) {
   return new Set(array).size !== array.length
 }
 
+/**
+ * Ensure brackets (as an array) contains only unique elements.
+ *
+ * @param {any[]} argv script arguments.
```
fedgiac, line 21: Shouldn't it be `object`, since it's not an array?
```suggestion
 * @param {object} argv script arguments.
```

scripts/utils/sign_and_send.js
```diff
@@ -32,8 +37,12 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
 
   /**
    * Signs and sends the transaction to the gnosis-safe UI
-   * @param {Address} masterAddress Address of the master safe owning the brackets
+   *
+   * @param {Address} masterSafe Address of the master safe owning the brackets
    * @param {Transaction} transaction The transaction to be signed and sent
+   * @param {string} network either rinkeby or mainnet
+   * @param {number} [nonce=null] specified transaction index. Will fetch correctg value if not specified.
```
fedgiac, line 44:
```suggestion
   * @param {number} [nonce=null] specified transaction index. Will fetch correct value if not specified.
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -18,67 +28,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
-   * @property {object} instance an instance of the token contract
-   */
-
   /**
    * Returns an instance of the exchange contract
+   *
+   * @param {object} web3 an instance of Web3
+   * @returns {SmartContract} An instance of BatchExchange
    */
   const getExchange = function (web3) {
```
fedgiac, line 37: `web3` should not be here since it's already defined as the input to the module function! I'd say to remove it in this PR.
```suggestion
  /**
   * Returns an instance of the exchange contract
   *
   * @returns {SmartContract} An instance of BatchExchange
   */
  const getExchange = function () {
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -119,10 +77,11 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
   const globalTokenPromisesFromAddress = {}
   /**
-   * Queries EVM for ERC20 token details by address
-   * and returns a list of promises of detailed token information.
+   * Queries EVM for ERC20 token details by address and returns a list of promises of detailed token information.
+   *
    * @param {Address[]} tokenAddresses list of *unique* token addresses whose data is to be fetch from the EVM
-   * @return {Promise<TokenObject>[]} list of detailed/relevant token information
+   * @param {boolean} [debug=false] prints log statements when true
+   * @returns {Promise<TokenObject>[]} list of detailed/relevant token information
```
fedgiac, line 84: This type was actually wrong, since the output is a mapping from addresses to promises of token objects. So, if we want to be formally correct we should use the much less informative `object`.
```suggestion
   * @returns {object} object mapping token addresses to a promise of relevant token information
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -154,11 +113,12 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
-   * Queries EVM for ERC20 token details by token id
-   * and returns a list of detailed token information.
+   * Queries EVM for ERC20 token details by token id and returns a list of detailed token information.
+   *
    * @param {SmartContract} exchange BatchExchange, contract, or any contract implementing `tokenIdToAddressMap`
-   * @param {integer[]} tokenIds list of *unique* token ids whose data is to be fetch from EVM
-   * @return {Promise<TokenObject>[]} list of detailed/relevant token information
+   * @param {number[]} tokenIds list of *unique* token ids whose data is to be fetch from EVM
+   * @param {boolean} [debug=false] prints log statements when true
+   * @returns {Promise<TokenObject>[]} list of detailed/relevant token information
```
fedgiac, line 121: Same issue as before.
```suggestion
   * @returns {object} object mapping token ids to a promise of relevant token information
```

scripts/utils/trading_strategy_helpers.js
```diff
@@ -185,8 +145,10 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
 
   /**
    * Retrieves token information foll all tokens involved in the deposits.
+   *
    * @param {(Deposit|Withdrawal)[]} flux List of {@link Deposit} or {@link Withdrawal}
-   * @return {Promise<TokenObject>[]} list of detailed/relevant token information
+   * @param {boolean} [debug=false] prints log statements when true
+   * @returns {Promise<TokenObject>[]} list of detailed/relevant token information
```
fedgiac, line 151: Yet again, same issue as before.
```suggestion
   * @returns {object} object mapping token addresses to a promise of relevant token information
```

package.json
bh2smith: This was done by yarn or npm. Wasn’t me. It’s the automatic linter

scripts/typedef.js
bh2smith: Is it just string or BN? Can it also be number?

bh2smith:
Thanks @fedgiac for the valuable feedback and thorough review. I have replaced some of your suggestions with {number|string|BN} I hope this is ok!

PR #272, opened 2020-05-22
Remove dependency ethereumjs-util

bh2smith:
It was only used for `keccak256`which is already available in `web3js`

Closes #272 

**Test Plan** Existing unit tests


scripts/utils/internals.js
```diff
@@ -171,6 +168,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @return {Address} Fallback contract of the input Gnosis Safe
    */
   const getFallbackHandler = async function (safeAddress) {
+    const fallbackHandlerStorageSlot = (await web3.utils.keccak256("fallback_manager.handler.address")).toString("hex")
```
nlordell, line 171: nit: the `keccak256` method already returns a string. Also according to the docs it isn't a promise
```suggestion
    const fallbackHandlerStorageSlot = web3.utils.keccak256("fallback_manager.handler.address")
```

Also, this was a module level constant before. Any reason why you moved it? I think this value is fixed and doesn't need to be calculated for each call.

(and to answer your question, I think the `0x` was there before because `ethereumjs-utils` returned a different value than the web3 utils).

nlordell (approved):

scripts/utils/internals.js
bh2smith: It needs to be awaited from web3. I moved it in here because it’s only used here and it’s now in an async function. Does that seem appropriate?

fedgiac (approved):
Nice, one dependency less.

bh2smith:
Would anyone be opposed to merging this into the PK branch first and then it will just be included when that gets merged?

nlordell:
> Would anyone be opposed to merging this into the PK branch first and then it will just be included when that gets merged?

I feel like this is a nice well contained change that should be merged on its own. Can't you just rebase onto master and then merge?

bh2smith:
> I feel like this is a nice well contained change that should be merged on its own. Can't you just rebase onto master and then merge?

Unfortunately I cannot at the #258 contains another removed occurrence of this dependency.

nlordell:
> Unfortunately I cannot at the #258 contains another removed occurrence of this dependency.

My suggestion would be to replace the occurrence as you did here, then you can merge master into #258 once this is merged.

PR #279, opened 2020-05-25
Split withdraw script

### LINKED ISSUE #222: Split withdraw.js into logical components depending on argv
Suggested by Felix in [this comment](https://github.com/gnosis/dex-liquidity-provision/pull/213#discussion_r421104109).

The logic of the withdraw script is markedly different depending on the flags `--requestWithdraw`, `--withdraw`, `--transferFundsToMaster`. The body of the main function could just be used to determine which case is being considered and defer the rest to a function tailored for that case.

fedgiac:
Split the withdraw script into two scripts: `request_withdraw.js` and `withdraw.js`.
All four logical steps in the script have now their own function. They are:
1. requesting a withdrawal
2. withdrawing from the exchange to the brackets
3. transferring ERC20 tokens from the brackets to master
4. simultaneous 2 and 3

Since I had to work on it anyway, I also changed the focus of the documentation from a deposit file to `--brackets` and `--tokenIds`.

PR is relevant to @josojo because it changes the tutorial on how to use the scripts.

Closes #222.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -97,17 +97,19 @@ npx truffle exec scripts/transfer_approve_deposit.js --masterSafe=$MASTER_SAFE -
 
 ### Withdrawing
 
-Funds can be withdrawn using the scripts `request_withdraw.js` and `withdraw.js`.
+Funds can be withdrawn using the scripts `request_withdraw.js` and `claim_withdraw.js`.
 
 To this end, you must specify the brackets you want to withdraw from with `--brackets` and the tokens to be withdrawn using either `--tokens` (which takes a list of addresses) or `--tokenIds` (which takes a list of token IDs).
+To make sure that the transaction does not need more gas than what fits in a block, the amount of tokens times the amount of brackets should indicatively be smaller than 40.
 
 First, a withdraw request must be created for each bracket and token.
-The following command request withdrawing of all DAI and WETH (token ID 7 and 1 respectively) for the brackets at addresses `0x1` and `0x2`:
+The following command request withdrawing of all DAI and WETH (token ID 7 and 1 respectively) for the brackets at addresses `0x0000000000000000000000000000000000000001` and `0x0000000000000000000000000000000000000002`:
 
 ```js
-npx truffle exec scripts/request_withdraw.js --masterSafe=$MASTER_SAFE --brackets=0x1,0x2 --tokenIds=1,7 --network=$NETWORK_NAME
+npx truffle exec scripts/request_withdraw.js --masterSafe=$MASTER_SAFE --brackets=0x0000000000000000000000000000000000000001,0x0000000000000000000000000000000000000002 --tokenIds=1,7 --network=$NETWORK_NAME
 ```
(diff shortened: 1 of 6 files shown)


scripts/README.md
```diff
@@ -97,42 +97,54 @@ npx truffle exec scripts/transfer_approve_deposit.js --masterSafe=$MASTER_SAFE -
...
-npx truffle exec scripts/withdraw.js --withdraw --transferFundsToMaster --masterSafe=$MASTER_SAFE --withdrawalFile="./examples/exampleDepositList.json" --network=$NETWORK_NAME
+npx truffle exec scripts/withdraw.js --masterSafe=$MASTER_SAFE --brackets=0x1,0x2 --tokenIds=1,7 --network=$NETWORK_NAME
 ```
 
-The latter instruction can be split into two independent units, if needed: withdrawing from the exchange to the bracket and transferring funds from the bracket to the master Safe.
+If necessary, for example if some funds are stuck in some of the bracket, the previous command can be split into two independent units: withdrawing from the exchange into the bracket and transferring funds from the bracket into the master Safe.
 
 ```js
-npx truffle exec scripts/withdraw.js --withdraw --masterSafe=$MASTER_SAFE --withdrawalFile="./examples/exampleDepositList.json" --network=$NETWORK_NAME
+npx truffle exec scripts/withdraw.js --withdraw --masterSafe=$MASTER_SAFE --brackets=0x1,0x2 --tokenIds=1,7 --network=$NETWORK_NAME
```
bh2smith, line 125: Would this flag be more descriptive if it were called `claimOnly`? Just a thought. In fact, I feel like the script itself might benefit from being called `claim_withdraw.js`. 

Furthermore, do we really have a realistic use case for ever wanting the funds to only be put back into the brackets? It almost seems like we shouldn't even bother making this a possibility. Why not bind these operations together so that users never even have to think about whats happening in between.

bh2smith: On that very same note, if we are going to keep this "leave funds in limbo" feature, we should probably allow the same for deposits. I am now beginning to think that we should rename the transferApproveDeposit to just "deposit" and try to forget about whats happening in between.

fedgiac:
I agree that case 2 (withdrawing from the exchange to the brackets) has no clear use case, I'd be ok with removing it. What has been used before and I'd be against removing is case 3 (transferring ERC20 tokens from the brackets to master): If you ever ran `request_withdraw` a second time immediately after the first you'd get a similar effect to executing `withdraw`, that is your funds are moved to the brackets, and  you must retrieve them in some non-standard way.
I also like the idea of renaming the script to `claim_withdraw.js`, so it's clear that you can't just withdraw straight away. What about renaming the script like you propose, making it single purpose (withdraw+transferFundsToMaster), and creating yet another script `transfer_funds_to_master.js` to take care of the only remaining clear use case that might be important?

bh2smith:
I absolutely love this @fedgiac! Really good example of the non-idempotency of `request_withdraw`!

PR #283, opened 2020-05-27
Use Node version 12 in Travis

fedgiac:
Since we support the latest version of Node, we should be testing against it in Travis.
The same change could be done in `dex-contracts`, but it should be simple since no changes appear to be required there to switch to Node v12.

Note that it very much looks like installation fails when running `yarn install`, at least on my machine and on Travis. Does it make sense to write a warning somewhere in the readme about this?

Two fixes were required to make Node v12 work in this repo.
1. A test had to change because the name of assertion errors has changed from `'AssertionError [ERR_ASSERTION]'` to `'AssertionError'`, I removed this check to keep our tests compatible with Node v10.
2. Gnosis Safe `v1.1.1-dev.1` did not work with Node v12, so it had to be updated to `v1.1.1-dev.2`. An added bonus is that we can go back to using the npm version, as well as updating when future versions come out!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ Install needed dependencies:
 yarn install
 ```
 
+Note that the installation might be successful even if errors are shown in the console output.
+In case of doubt, running `echo $?` immediately after `yarn install` should return 0 if the installation was successful.
+
 Run Ganache in the background with an increased gas limit (this is needed to deploy Gnosis Protocol):
 
 ```
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "@gnosis.pm/dex-contracts": "^0.3.2",
     "@gnosis.pm/owl-token": "^3.1.0",
-    "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.1",
+    "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.2",
     "@gnosis.pm/solidity-data-structures": "=1.2.4",
     "@gnosis.pm/util-contracts": "^2.0.6",
     "@openzeppelin/contracts": "=2.5.1",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "@gnosis.pm/dex-contracts": "^0.3.2",
     "@gnosis.pm/owl-token": "^3.1.0",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
   "dependencies": {
     "@gnosis.pm/dex-contracts": "^0.3.2",
     "@gnosis.pm/owl-token": "^3.1.0",
-    "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.1",
+    "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.2",
```
bh2smith, line 44: Was this published just for us!?

bh2smith (approved):
Looks good to me, I suppose it wouldn't hurt to mention something in the readme about the installation looking like it may have failed even when it was fine.

package.json
fedgiac: The changes that we need in `safe-contract` will officially be part of future versions, so we will be able to update to newer versions without any issue in the future!

fedgiac:
I believe, based on the messages in PR #161, that Dependabot will pick up future versions of `safe-contracts` without me needing to do anything to reenable it, since I've manually changed the version.

package.json
```diff
@@ -41,7 +41,7 @@
   "dependencies": {
     "@gnosis.pm/dex-contracts": "^0.3.2",
     "@gnosis.pm/owl-token": "^3.1.0",
-    "@gnosis.pm/safe-contracts": "github:gnosis/safe-contracts#v1.1.1-dev.1",
+    "@gnosis.pm/safe-contracts": "v1.1.1-dev.2",
```
josojo, line 44: If we are changing the contract here, will this have an impact on "verifying old setups"?

josojo (approved):
Thanks for trying it out with node v12

package.json
fedgiac: Yes, this is a valid concern, if the contract changes in any non-trivial way then our code is broken. In this case, the changes didn't modify the bytecode.

README.md
```diff
@@ -14,6 +14,9 @@ Install needed dependencies:
 yarn install
 ```
 
+Note that the installation might be successful even if errors are shown in the console output.
+In case of doubt, running `echo $?` immediately after `yarn install` should return 0 if the installation was successful.
```
bh2smith, line 18: Awesome!

bh2smith (approved):
Perf 👌 !

Issue #248, opened 2020-05-13
Liquidity provision for signer: check existence of transaction in web interface


fedgiac:
The current process for a signer to verify the transaction appearing in the web interface involves receive the hash from the transaction creator (plus some more info) and compare it to the output of the scripts with `--verify`.
Safer and quicker would be to compute the hash and use it to retrieve the transaction from the web interface, then tell the user that the transaction is there. (Also checking whether it was already executed.)
This is currently doable using the [Safe Multisig API](https://docs.gnosis.io/safe/docs/services_transactions/): `/api/v1/transactions/{safe-tx-hash}/ GET`, "get transaction by internal SafeTxHash".

PR #285, opened 2020-05-29
New verify logic: Rebuilding and checking for proposed transaction on servers

josojo:
closes: #248 

Initially, it was the idea to display the transaction ID for the proposed account. Currently, this is not available via the API. Hence, people have to check for the correct nonce, in order to identify their transaction.

testplan:

Successfully verifying my transaction:
```
export NETWORK_NAME=rinkeby
export MASTER_SAFE=0xEc9279204C734E335f8b7f7F660dE19968F15b99

npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=0 --quoteTokenId=7 --lowestLimit=0.25 --highestLimit=1 --currentPrice=0.41 --masterSafe=$MASTER_SAFE --depositBaseToken=0.001 --depositQuoteToken=0.0001 --numBrackets=2 --network=$NETWORK_NAME --verify --nonce=54 --brackets=0x8be2011F0A4391874d04A46275F23a480A1Fd355,0x5f605D76024692c784B37B98f78972cC6B1B8Fb3
```
And change the nonce in the upper script and then see that the verification fails.


scripts/complete_liquidity_provision.js
```diff
@@ -182,32 +182,31 @@ module.exports = async (callback) => {
...
       console.log("==> Order placing transaction")
+      await checkTransactionExistenceOnSafeServer(masterSafe, orderTransaction, argv.network, nonce)
     }
-    let nonce = argv.nonce
-    if (nonce === undefined) {
-      nonce = (await masterSafe.nonce()).toNumber()
-    }
-    await signAndSend(masterSafe, orderTransaction, argv.network, nonce, argv.verify)
 
     if (!argv.verify) {
```
fedgiac, line 199: It makes sense to consolidate the two `if (!argv.verify)` into a single one.

scripts/utils/sign_and_send.js
```diff
@@ -7,6 +7,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
   const { getSafeCompatibleSignature, estimateGas } = require("../utils/internals")(web3, artifacts)
   const { ZERO_ADDRESS } = require("./constants")
+  const baseGas = 0
```
fedgiac, line 10: Very much nitpick, but I think that the name is too simple if we want to move this constant to the preamble of the script. I'd consider leaving the constant where it was and duplicate it the next time you use it.

scripts/utils/sign_and_send.js
```diff
@@ -73,7 +67,44 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     console.log("Transaction awaiting execution in the interface", interfaceLink)
   }
 
+  /**
+   * Checks that a transaction was already proposed to the gnosis-safe UI
+   *
+   * @param {Address} masterSafe Address of the master safe owning the brackets
+   * @param {Transaction} transaction The transaction to be signed and sent
```
fedgiac, line 74:
```suggestion
   * @param {Transaction} transaction The transaction whose existence is checked
```

scripts/utils/sign_and_send.js
```diff
@@ -73,7 +67,44 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+   * @param {string} network either rinkeby or mainnet
+   * @param {number} [nonce=null] specified transaction index. Will fetch correct value if not specified.
```
fedgiac, line 76: Nitpick, the index is a different number in the web interface.
```suggestion
   * @param {number} [nonce=null] specified Gnosis Safe transaction nonce. Will fetch correct value if not specified.
```

scripts/utils/sign_and_send.js
```diff
@@ -73,7 +67,44 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+      transaction.to,
+      transaction.value,
+      transaction.data,
+      transaction.operation,
+      safeTxGas,
+      baseGas,
+      0,
+      ZERO_ADDRESS,
+      ZERO_ADDRESS,
+      nonce
```
fedgiac, line 90: I don't think you can use `nonce = null` to estimate the gas, you need to retrieve the nonce first.

scripts/utils/sign_and_send.js
```diff
@@ -73,7 +67,44 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+    const endpoint = `https://safe-transaction.${network}.gnosis.io/api/v1/transactions/${transactionHash}/`
+    await axios.get(endpoint).catch(function (error) {
+      if (error.response.data.detail === "Not found.") {
+        throw new Error("The reproduced transaction was not found. There is something wrong. Contact the proposer.")
+      }
+      throw new Error("Error while talking to the gnosis-interface: " + JSON.stringify(error.response.data))
+    })
+    const interfaceLink = `https://${linkPrefix[network]}gnosis-safe.io/app/#/safes/${masterSafe.address}/transactions`
+    console.log(
+      `The reproduced transaction was is already proposed! You can sign the transaction with nonce ${nonce} here: ${interfaceLink}`
```
fedgiac, line 102:
```suggestion
      `The transaction built from the verification parameters matches a transaction in the web interface. You can sign the transaction with nonce ${nonce} here: ${interfaceLink}`
```

scripts/utils/sign_and_send.js
bh2smith: Interesting catch... was this detected when running the test plan?

scripts/utils/sign_and_send.js
bh2smith: I really like the new suggested wording here!

scripts/utils/sign_and_send.js
```diff
@@ -73,7 +67,44 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+      transaction.data,
+      transaction.operation,
+      safeTxGas,
+      baseGas,
+      0,
+      ZERO_ADDRESS,
+      ZERO_ADDRESS,
+      nonce
+    )
+    console.log(`The reproduced transaction has the gnosis-safe hash ${transactionHash} and nonce ${nonce}`)
```
bh2smith, line 92: What is a "gnosis-safe hash"? Can't it just be a hash? I don't think we need/should explicitly say "reproduced" here since this function is independent of the fact that the hash was reproduced for our use case.

However, I can imagine one might want to see the term reproduced if they are indeed reproducing the transaction. This being said, I can suggest that we simplify this function by removing the use case specific logs and have it return a boolean. Something like 

```py
def transactionExisitsOnSafeServer(...args) -> bool:
```

Then bubble the use case specific logs out into the script where this is called instead. This will make the function more robust for other purposes. 

It would also make me happy to get the word "check" out of the function name.

fedgiac: A "Gnosis Safe hash" is the hash internally used by a Gnosis Safe to identify transactions. I think it's reasonable to remark the difference from the much more common "transaction hash", which is used to identify an Ethereum transaction. 

I really like the idea of having this function return a boolean and moving out the logs.

scripts/utils/sign_and_send.js
fedgiac: Nope, while reading the comments! They say that the nonce must retrieved first, but it's not.

scripts/utils/sign_and_send.js
josojo: Okay, I made the name more descriptive. I really would like to avoid two different variables, as this could cause discrepancies in the transactions later on.

josojo:
Why do you guys prefer it, if the function is returning a bool? Which coding paradigm is speaking for it? If you guys prefer the function to be used for future purposes, then adding more parameters for better printing messages seems for me the better option.

I changed the code, as you both voted for it. But now, the complete_liquidity_provision is much longer and there is a lot of "similar" print statements, which falls into the code duplication category. 

I would prefer to keep the complete_liquidity_provision as simple as possible and shift as much as possible in a well-abstracted manner into other functions.
@bh2smith @fedgiac

scripts/complete_liquidity_provision.js
```diff
@@ -182,32 +186,43 @@ module.exports = async (callback) => {
...
+      }
+      console.log("==> Funds transferring transaction")
+      if (await transactionExistsOnSafeServer(masterSafe, bundledFundingTransaction, argv.network, nonce + 1)) {
+        console.log(
+          `The funds transferring transaction built from the verification parameters matches the transaction in the interface! You can sign the transaction with nonce ${nonce} here: ${interfaceLink}`
+        )
+      } else {
+        console.log(
+          "Error: The funds transferring transaction built from the verification parameters does not match any transaction in the interface!"
+        )
```
bh2smith, line 224: Although this might look like code duplication, notice that the error messages are different. Of course we could come up with a more generic message here. My point was more about the fact the we need to remove the term "reconstructed" transaction since the function can be used for other purposes (i.e. like just verifying if the same transaction already exists when it has not been reconstructed). 

I would be ok with the old implementation if these ideas are considered.

fedgiac:
I don't think that the script is simpler if we keep the logs inside the functions, if anything we hide the complexity, since the function we call is not tested anyway. On the other hand, with these changes we could test `transactionExistsOnSafeServer` with a JS test if we ever felt the need to.
I don't have a strong opinion on this however, I was also fine with how it was before.

bh2smith:
>  I was also fine with how it was before.

My only major concern with how it was before was that the wording contained something like "reproduced" or "reconstructed" transaction that made it less generic than it necessarily had to be (because I can think of a use case where the exact same logic could be used on transaction data that wasn't necessarily "reconstructed", but simply just a retry of the already existing transaction). I am sorry if this caused more trouble than intended.

josojo:
I prefer to make the error as obvious as possible. Hence, I am throwing it. This has the additional benefit that the exit code of the script can be checked from other scripts as well.

But you are right that one might wish to differentiate between unexpected errors and "possible script outcomes". I am not sure here...

PR #286, opened 2020-05-29
Parallelize checks for complete liquidity provision

### LINKED ISSUE #277: Make complete liquidity provision faster on startup
We collected many checks at the start of the complete liquidity provision script, all of which are executed in series. All these checks and some more instructions could be executed in parallel instead, achieving a better running time.

fedgiac:
Makes complete liquidity provision faster by parallelizing things with async/await.
I intentionally avoided `Promise.all` for readability and efficiency.

Running times to the first check, assuming brackets are not given.
- before this PR: 19±1s
- after this PR: 13±1s
Considering that Truffle itself takes 6±1s to initialize, the running time of the logic in this script was reduced by a factor of 2.

If brackets are given the script should be even more efficient in comparison.

Closes #277.

The script can be tried with `npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=210 --highestLimit=230 --currentPrice=220 --depositBaseToken=0.01 --depositQuoteToken=1 --numBrackets=20 --network=rinkeby  --masterSafe=FILLME`, assuming master has USDC and WETH.
A follow-up test is to try the same command with the extra option ` --brackets=`.

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -117,9 +117,9 @@ module.exports = async (callback) => {
     const { instance: quoteToken, decimals: quoteTokenDecimals } = quoteTokenData
     const depositBaseToken = toErc20Units(argv.depositBaseToken, baseTokenDecimals)
     const depositQuoteToken = toErc20Units(argv.depositQuoteToken, quoteTokenDecimals)
-    const isBaseTokenBalanceSufficientPromise = checkSufficiencyOfBalance(baseToken, argv.masterSafe, depositBaseToken)
-    const isQuoteTokenBalanceSufficientPromise = checkSufficiencyOfBalance(quoteToken, argv.masterSafe, depositQuoteToken)
-    const priceCheckPromise = isPriceReasonable(baseTokenData, quoteTokenData, argv.currentPrice)
+    const hasSufficientBaseTokenPromise = checkSufficiencyOfBalance(baseToken, argv.masterSafe, depositBaseToken)
+    const hasSufficientQuoteTokenPromise = checkSufficiencyOfBalance(quoteToken, argv.masterSafe, depositQuoteToken)
+    const isPriceCloseToDexagsPromise = isPriceReasonable(baseTokenData, quoteTokenData, argv.currentPrice)
 
     const signer = await signerPromise
     console.log("Using account:", signer)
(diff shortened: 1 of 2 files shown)


scripts/complete_liquidity_provision.js
```diff
@@ -81,42 +82,62 @@ const argv = default_yargs
...
-    if (!(await checkSufficiencyOfBalance(quoteToken, masterSafe.address, depositQuoteToken))) {
-      callback(`Error: MasterSafe ${masterSafe.address} has insufficient balance for quote token ${quoteToken.address}`)
+    if (!(await isQuoteTokenBalanceSufficientPromise)) {
+      callback(`Error: MasterSafe ${argv.masterSafe} has insufficient balance for quote token ${quoteToken.address}`)
     }
+
     // check price against dex.ag's API
-    const priceCheck = await isPriceReasonable(baseTokenData, quoteTokenData, argv.currentPrice)
-    if (!priceCheck) {
+    if (!(await priceCheckPromise)) {
```
bh2smith, line 140: nit: although I am generally not a fan of having entire sentences for variable names, I feel like "priceCheck" has been neglected here. 

Also, if you are looking for something equivalent but slightly shorter for the "is" promises:

```
isXXXTokenBalanceSufficientPromise
hasSufficientXXXToken
```

scripts/complete_liquidity_provision.js
```diff
@@ -131,28 +152,21 @@ module.exports = async (callback) => {
...
-      await verifyBracketsWellFormed(masterSafe.address, bracketAddresses, null, null, true)
-      // Detect if provided brackets have existing orders.
-      const existingOrders = await Promise.all(
-        bracketAddresses.map(async (safeAddr) => {
-          return hasExistingOrders(safeAddr, exchange)
-        })
-      )
-      const dirtyBrackets = bracketAddresses.filter((_, i) => existingOrders[i] == true)
+      await assertBracketsWellFormedPromise
+      const dirtyBrackets = await bracketsWithExistingOrdersPromise
```
bh2smith, line 159: I'm happy to see this name stuck around!

scripts/complete_liquidity_provision.js
```diff
@@ -131,28 +152,21 @@ module.exports = async (callback) => {
...
       if (
-        existingOrders.some((t) => t) &&
+        dirtyBrackets.length !== 0 &&
```
bh2smith, line 161: And we found a place to use it twice! This is where I very much wish we were using python so we could write 

```py
if any(dirtyBrackets):
```

scripts/complete_liquidity_provision.js
```diff
@@ -131,28 +152,21 @@ module.exports = async (callback) => {
...
         !(await proceedAnyways(`The following brackets have existing orders:\n  ${dirtyBrackets.join()}\n`))
       ) {
         callback("Error: Existing order verification failed.")
       }
     } else {
       assert(!argv.verify, "Trading Brackets need to be provided via --brackets when verifying a transaction")
       console.log(`==> Deploying ${argv.numBrackets} trading brackets`)
-      bracketAddresses = await deployFleetOfSafes(masterSafe.address, argv.numBrackets)
+      bracketAddresses = await deployFleetOfSafes(argv.masterSafe, argv.numBrackets)
```
bh2smith, line 169: ~~Nice change here. Do we happen to know if there was any performance boost with regards to this change?~~

Never mind, I see what you did down below (didn't need to resolve the master safe till way later.

bh2smith (approved):
This is very nice. I like what you have done here! Maybe just include a comment about the "promise initialization phase"

scripts/complete_liquidity_provision.js
fedgiac: Yes, this is the reason I did it. I don't actually save any time in practice by moving the resolution to the end of the file: at some point early on I need to resolve `masterOwnersPromise` anyway, which in turn requires resolving `masterSafePromise`. It might still help if we'll ever change how we resolve `masterOwnersPromise`.

PR #292, opened 2020-06-03
Simplify installation steps

fedgiac:
Creates more yarn scripts to simplify setting up of the project and adds building artifacts to Travis.
Might be interesting to amend the tutorial with this new procedure, @josojo. (I could also do it instead.)

To test the new commands, check that [Travis is working as expected](https://travis-ci.com/github/gnosis/dex-liquidity-provision/builds/169430145) and try to reinstall this repo from scratch:
```
git clone https://github.com/gnosis/dex-liquidity-provision.git
cd dex-liquidity-provision
yarn install
yarn build
```
and then try to run any of the scripts in Rinkeby.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -8,12 +8,11 @@ Use at your own risk!
 
 The scripts require the following software installed: [git](https://git-scm.com/), [yarn](https://yarnpkg.com/) and [node](https://nodejs.org/en/).
 
-The Gnosis Protocol contracts must be compiled and the deployment addresses must be injected into built contract-artifacts.
-To do so, run:
 
+Install needed dependencies and build needed artifact:
 ```
-yarn compile
-yarn run networks-inject
+yarn install
+yarn builD
 ```
 
 Create a gnosis-safe wallet [here-mainnet](https://gnosis-safe.io) or [here-rinkeby](https://rinkeby.gnosis-safe.io). This wallet will be called your Master Safe in the following. It is used to bundle the transactions and setup the bracket-traders.
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -8,11 +8,11 @@ Use at your own risk!
 
 The scripts require the following software installed: [git](https://git-scm.com/), [yarn](https://yarnpkg.com/) and [node](https://nodejs.org/en/).
 
-
 Install needed dependencies and build needed artifact:
+
 ```
 yarn install
-yarn builD
+yarn build
 ```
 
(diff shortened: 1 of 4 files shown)


scripts/README.md
```diff
@@ -8,12 +8,11 @@ Use at your own risk!
...
 
-The Gnosis Protocol contracts must be compiled and the deployment addresses must be injected into built contract-artifacts.
-To do so, run:
 
+Install needed dependencies and build needed artifact:
 ```
-yarn compile
-yarn run networks-inject
+yarn install
+yarn builD
```
josojo, line 15: Here is an upper letter hiding

fedgiac:
Manual execution of Ganache: I thought `yarn testnet` helps because 1. we tell developers the options we expect without having to change the readme every time 2. we wouldn't need to remember all necessary Ganache configurations every time we need to switch projects.
I see how it might be overkill, since we don't make many configuration changes in all our projects.

Yarn prepack: dex-contracts defines `prepack` to be [an alias](https://github.com/gnosis/dex-contracts/blob/master/package.json#L28) to `build`, and build also inject the networks. Are you proposing to also add the alias `prepack` to this repo? I personally don't understand what "prepack" is supposed to mean, so I'm not sure how to implement it.

Updating tutorial: the release idea seems good to me. We could collect breaking changes in a txt file in this repo and remove the file when releasing a new version, which implies updating the tutorial. If you like this, I could already create the file.

josojo:
> Updating tutorial: the release idea seems good to me. We could collect breaking changes in a txt file in this repo and remove the file when releasing a new version, which implies updating the tutorial. If you like this, I could already create the file.

Is the breaking changes file necessary? We could also tag and name it in upcoming PRs. And then once a new release is done, we go through the tagged Prs and collect all breaking changes. I think that was also the group's favorite approach for dex-contracts...

bh2smith:
Thanks @fedgiac for your explanation here. I wasn't aware of the aliasing that was going on so this is all clear and, I agree, unnecessary. 

As for the tutorial updating, I think the easiest way to go here (and possibly what @josojo is suggesting), is that we simply tag this repo at a certain point - like right after this is merged - with a version number and then instruct the user to checkout the project at the specific tag. This way, it is clear that the documentation should work (as described) for the checkout version.

fedgiac:
I was thinking to make it even simpler: we could tag breaking changes with `requires tutorial change` and once the tutorial is updated we remove all tags. This could be done in a release cycle, but also independently of it.

PR #293, opened 2020-06-03
Fix bytecode check

### LINKED ISSUE #217: [Bug] verifyBracketsWellFormed not working as expected.
A few problems when trying to run bracket verification:

```
- Verify that masterCopy of brackets is the known masterCopy
Error: Returned values aren't valid, did it run Out of Gas?
    at getMasterCopy (/Users/admin/Projects/gnosis/dex-liquidity-provision/scripts/utils/internals.js:170:17)
```

```
- Verify proxy bytecode
AssertionError [ERR_ASSERTION]: Bad bytecode for bracket 0xfA4a18c2218945bC018BF94D093BCa66c88D3c40
```

Note that I am using the following safe setup on rinkeby:
```
masterSafe: 
    0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c
brackets: 
    0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636
    0xfA4a18c2218945bC018BF94D093BCa66c88D3c40
```

and that these brackets were created manually. It is possible that the second error is because the safes were created manually, but if this is the case I would expect the error message to be a bit more specific.

fedgiac:
Reintroduces proxy bytecode verification. See discussion in issue #217.

Unrelated to this PR, but note that the verification script could be made much faster with the same simple tricks used in PR #286.

Closes #217.

Test plan: unit test for bytecode changes was restored.
On mainnet, this should be successful:
```
npx truffle exec scripts/verify_correctness.js  --network=mainnet --brackets=0x7c46953a4f0b404e21ebc0d69883adaef5aaa5c5 --masterSafe=0x5CC3a4C846aB8b0B78a729b8B1a50e2E0BB66740
```
and this should fail:
```
npx truffle exec scripts/verify_correctness.js  --network=mainnet --brackets=0xaafd977483df35bc90c27abbf96e6f7df5143013 --masterSafe=0x5CC3a4C846aB8b0B78a729b8B1a50e2E0BB66740
```

### DIFF
--- a/scripts/utils/verify_scripts.js
+++ b/scripts/utils/verify_scripts.js
@@ -27,6 +27,19 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     const master = await getSafe(masterAddress)
     const brackets = await Promise.all(bracketAddresses.map((bracketAddress) => getSafe(bracketAddress)))
 
+    // TODO - enable this verification according to discussion in
+    // https://github.com/gnosis/dex-liquidity-provision/issues/217
+    // log("- Verify proxy bytecode")
+    // await Promise.all(
+    //   bracketAddresses.map(async (bracketAddress) => {
+    //     assert.equal(
+    //       await web3.eth.getCode(bracketAddress),
+    //       GnosisSafeProxy.deployedBytecode,
+    //       `Bytecode at bracket ${bracketAddress} does not agree with that GnosisSafeProxy v1.1.1`
+    //     )
+    //   })
+    // )
+
     if (!masterOwners || !masterThreshold) log("Warning: master safe owner verification skipped")
     else {
       log("- Verify owners of masterSafe")
@@ -43,18 +56,6 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
         "Master owners are different than expected"
       )
     }
-    // TODO - enable this verification according to discussion in
-    // https://github.com/gnosis/dex-liquidity-provision/issues/217
-    // log("- Verify proxy bytecode")
-    // await Promise.all(
-    //   bracketAddresses.map(async (bracketAddress) => {
-    //     assert.equal(
-    //       await web3.eth.getCode(bracketAddress),
-    //       GnosisSafeProxy.deployedBytecode,
-    //       `Bytecode at bracket ${bracketAddress} does not agree with that GnosisSafeProxy v1.1.1`
-    //     )
-    //   })
-    // )
 
     log("- Verify that brackets are owned solely by masterSafe")
     await Promise.all(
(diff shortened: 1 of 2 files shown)


test/verification_script.js
```diff
@@ -163,18 +163,18 @@ contract("Verification checks", function (accounts) {
...
+  describe("Brackets' deployed bytecode coincides with that of a Gnosis Safe proxy", async () => {
+    it("throws if bytecode differs", async () => {
+      const masterSafe = await GnosisSafe.at(await deploySafe(gnosisSafeMasterCopy, proxyFactory, [safeOwner], 1))
+      const evilProxy = await EvilGnosisSafeProxy.new(GnosisSafe.address)
+      const evilSafe = await GnosisSafe.at(evilProxy.address)
+      await evilSafe.setup([masterSafe.address], "1", ZERO_ADDRESS, "0x", ZERO_ADDRESS, ZERO_ADDRESS, "0", ZERO_ADDRESS)
+      await assert.rejects(verifyCorrectSetup([evilProxy.address], masterSafe.address), {
+        message: `Bytecode at bracket ${evilProxy.address} does not agree with that of a Gnosis Safe Proxy v1.1.1`,
+      })
+      // TODO - this test only checks against bad proxy, but not against a bad safe.
```
fedgiac, line 175: This comment was already there. I'd say we don't need to do further checks, since there is no fundamental reason why this check should not pass with a bad proxy but pass with a "bad Safe". Is it fine if I delete this TODO?

scripts/utils/verify_scripts.js
```diff
@@ -11,8 +11,9 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const { getDexagPrice, checkNoProfitableOffer } = require("./price_utils")
 
   const GnosisSafe = artifacts.require("GnosisSafe.sol")
-  // const GnosisSafeProxy = artifacts.require("GnosisSafeProxy.sol")
+  const GnosisSafeProxyFactory = artifacts.require("GnosisSafeProxyFactory.sol")
   const gnosisSafeMasterCopy = GnosisSafe.deployed()
+  const expectedBytecodePromise = GnosisSafeProxyFactory.deployed().then((proxyFactory) => proxyFactory.proxyRuntimeCode())
```
josojo, line 16: A nice trick to get the byteCode with the right libs.

fedgiac:
It should work for all brackets ever created by the scripts, since we've always used the same proxy factory!

The bytecode is not hardcoded: a nice feature of the Gnosis Safe proxy factory is that you can ask it the standard code of a deployed proxy, so I'm just comparing the actual code of a bracket to the standard bytecode of something created by the proxy factory. This means that we are trusting the code of the proxy factory, but that should be fine since the code is publicly available and anyway the proxy factory is what we use to create brackets in the first place.

PR #300, opened 2020-06-08
[Synthetix] Place orders only when likely match

### LINKED ISSUE #282: [Synthetix Bot] Improve Gas Consumption.
Based on the [Gas Cost Analysis](https://gnosisinc.slack.com/archives/CC3C5SMB5/p1590398784016900) for the calendar week 21 of 2020.

We have considered out 4 potential avenues for improvement with the following being the most promising:

Rather than placing orders in every batch, this bot service we check the onchain price oracle along with the existing orderbook and only place orders in the batch if it is expected to be matched. We note that, the bot would not be capable of considering orders placed in the last minute or two of the current batch. However, these orders would be would picked up in the next if they are still valid.

As a rough estimate, this is expected to substantially reduce the gas costs by placing orders sparsely rather than in every batch. Essentially, the bots previous actions of placing orders in every batch was literally the worst case of our new proposal (i.e. when there is a matching order in every batch). A potential attack vector then would be that there are potentially matching "dust orders" in every batch, so we should incorporate an economic viability model as well. 

This will involve making calls to our price estimator API: Here is an example link to show what the requests would return:

https://dex-price-estimator.gnosis.io//api/v1/markets/11-4/estimated-buy-amount/100000000?atoms=true

bh2smith:
Closes #282. 

This PR improves the synthetix script to only place order when there is a likely match according to the query from price estimation service. 

**Test Plan:**

Run

```sh
truffle exec scripts/synthetix/facilitate_trade.js --network mainnet
```
and observe the logs:

```
Using network 'mainnet'.

Using account 0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1
Oracle sETH Price (in sUSD) 243.3257626212688802
Exchange buy  sETH price (in sUSD) 246.6144973941891
Exchange sell sETH price (in sUSD) 233.72582778494302
Not placing buy  sETH order, our rate of 242.60 is too low  for exchange.
Not placing sell sETH order, our rate of 244.06 is too high for exchange.
```

On Rinkeby (i.e. where there is no s-liquidity), 

```sh
truffle exec scripts/synthetix/facilitate_trade.js --network rinkeby
```
with log output:

```
Using network 'rinkeby'.

Using account 0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1
Oracle sETH Price (in sUSD) 243.19490611887486801
Exchange buy  sETH price (in sUSD) Infinity
Exchange sell sETH price (in sUSD) 0
Not placing buy  sETH order, our rate of 241.98 is too low  for exchange.
Not placing sell sETH order, our rate of 244.41 is too high for exchange.
```

**Update Test Plan** @fleupold  mentioned it would be nice to have an example in which there is actually a matching order:

After placing an order to sell 15 sUSD for sETH at a price of 250.xx (much more than the current rate) and running this script we find:

Screenshot 2020-06-10 at 11 28 58 AM

```
➜  dex-liquidity-provision git:(282/synth_gas_savings) truffle exec scripts/synthetix/facilitate_trade.js --network mainnet
Using network 'mainnet'.

Using account 0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1
Oracle sETH Price (in sUSD) 242.74183486202928576
Exchange sell sETH price (in sUSD) 251.75399416750443
Exchange buy  sETH price (in sUSD) 249.74803241942786
Not placing buy  sETH order, our rate of 242.01 is too low  for exchange.
Placing an order to sell sETH at 242.8632057794603
Using current "standard" gas price scaled by 1: 36000000001
```

Unfortunately, we cannot test that this bot will actually place the orders until there is a match (and rinkeby does not contain any liquidity.

### DIFF
--- a/scripts/synthetix/facilitate_trade.js
+++ b/scripts/synthetix/facilitate_trade.js
@@ -65,6 +65,12 @@ const estimatePrice = async function (buyTokenId, sellTokenId, sellAmount, netwo
 
 const MIN_SELL_USD = 10
 
+/* All prices refered to hear are for sETH in sUSD.
+ * The term "our" referres to prices we are buying and selling for while
+ * the term "their" refers to the buy and sell prices offered by the exchange.
+ * For example, if ourBuyPrice = 100 and theirSellPrice = 90, this means
+ * we are willing to pay 100 sUSD for 1 ETH and the exchange is offering 1 ETH for 90 sUSD
+ */
 module.exports = async (callback) => {
   try {
     const networkId = await web3.eth.net.getId()
--- a/scripts/synthetix/facilitate_trade.js
+++ b/scripts/synthetix/facilitate_trade.js
@@ -2,10 +2,11 @@ const { SynthetixJs } = require("synthetix-js")
 const ethers = require("ethers")
 const fetch = require("node-fetch")
 const { getUnlimitedOrderAmounts } = require("@gnosis.pm/dex-contracts")
+const { default_yargs } = require("../utils/default_yargs")
 
 const { getExchange } = require("../utils/trading_strategy_helpers")(web3, artifacts)
 
-const argv = require("../utils/default_yargs")
+const argv = default_yargs
   .option("gasPrice", {
     type: "string",
     describe: "Gas price to be used for order submission",
@@ -50,6 +51,30 @@ const gasStationURL = {
   4: "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/",
 }
 
+const toWEI = function (value, decimals) {
+  const h = Math.floor(decimals / 2)
+  const a = new BN(10).pow(new BN(h))
+  const b = new BN(10).pow(new BN(decimals - h))
+
+  return new BN(value * a).mul(new BN(b))
+}
+
+const BN = require("bn.js")
+// Could also use linkPrefix here.
+const estimationURLPrexix = {
+  1: "https://dex-price-estimator.gnosis.io//api/v1/",
+  4: "https://dex-price-estimator.rinkeby.gnosis.io//api/v1/",
+}
+
+const priceEstimation = async function (buyTokenId, sellTokenId, amount, networkId) {
+  const searchCriteria = `markets/${buyTokenId}-${sellTokenId}/estimated-buy-amount/${amount}?atoms=true`
+  const estimationData = await (await fetch(estimationURLPrexix[networkId] + searchCriteria)).json()
+
+  return estimationData.buyAmountInBase / estimationData.sellAmountInQuote
+}
+
+const MIN_SELL_USD = 10
+
 module.exports = async (callback) => {
   try {
     const networkId = await web3.eth.net.getId()
@@ -73,44 +98,75 @@ module.exports = async (callback) => {
     // Note that sUSD always has a price of 1 within synthetix protocol.
     const exchangeRate = await snxjs.ExchangeRates.rateForCurrency(sETHKey)
     const formatedRate = snxjs.utils.formatEther(exchangeRate)
-    console.log("sETH Price", snxjs.utils.formatEther(exchangeRate))
+    console.log("Oracle sETH Price (in sUSD)", formatedRate)
+
+    const exchangeBuyETHPrice = await priceEstimation(
+      sETH.exchangeId,
+      sUSD.exchangeId,
+      toWEI(MIN_SELL_USD, sUSD.decimals),
+      networkId
+    )
+    console.log("Exchange buy  sETH price (in sUSD)", 1 / exchangeBuyETHPrice)
+
+    const minSellETH = toWEI(MIN_SELL_USD / formatedRate, sETH.decimals)
+    const exchangeSellETHPrice = await priceEstimation(sUSD.exchangeId, sETH.exchangeId, minSellETH, networkId)
+    console.log("Exchange sell sETH price (in sUSD)", exchangeSellETHPrice)
 
     // Using synthetix's fees, and formatting their return values with their tools, plus parseFloat.
     const sETHTosUSDFee = parseFloat(snxjs.utils.formatEther(await snxjs.Exchanger.feeRateForExchange(sETHKey, sUSDKey)))
     const sUSDTosETHFee = parseFloat(snxjs.utils.formatEther(await snxjs.Exchanger.feeRateForExchange(sUSDKey, sETHKey)))
 
-    // Compute buy-sell amounts based on unlimited orders with rates from above.
-    const { base: sellSUSDAmount, quote: buyETHAmount } = getUnlimitedOrderAmounts(
-      1 / (formatedRate * (1 - sUSDTosETHFee)),
-      sETH.decimals,
-      sUSD.decimals
-    )
-    const { base: sellETHAmount, quote: buySUSDAmount } = getUnlimitedOrderAmounts(
-      formatedRate * (1 + sETHTosUSDFee),
-      sUSD.decimals,
-      sETH.decimals
-    )
+    // Initialize all arrays needed for potential order placement.
+    const [buyAmounts, sellAmounts, buyTokens, sellTokens] = [[], [], [], []]
 
-    const buyAmounts = [buyETHAmount, buySUSDAmount]
-    const sellAmounts = [sellSUSDAmount, sellETHAmount]
+    // Compute buy-sell amounts based on unlimited orders with rates from above when the price is right.
+    const ourBuyETHRate = formatedRate * (1 - sUSDTosETHFee)
+    if (ourBuyETHRate < exchangeBuyETHPrice) {
+      console.log(`Placing an order to buy sETH at ${ourBuyETHRate}`)
+      const { base: sellSUSDAmount, quote: buyETHAmount } = getUnlimitedOrderAmounts(
+        1 / ourBuyETHRate,
+        sETH.decimals,
+        sUSD.decimals
+      )
+      buyAmounts.push(buyETHAmount)
+      sellAmounts.push(sellSUSDAmount)
+      buyTokens.push(sETH)
+      sellTokens.push(sUSD)
+    } else {
+      console.log(`Not placing buy  sETH order, our rate of ${ourBuyETHRate.toFixed(2)} is too low  for exchange.`)
+    }
 
-    // Fetch auction index and declare validity interval for orders.
-    // Note that order validity interval is inclusive on both sides.
-    const batchId = (await exchange.getCurrentBatchId.call()).toNumber()
-    const validFroms = Array(2).fill(batchId)
-    const validTos = Array(2).fill(batchId)
+    const ourSellETHRate = formatedRate * (1 + sETHTosUSDFee)
+    if (ourSellETHRate < exchangeSellETHPrice) {
+      console.log(`Placing an order to sell sETH at ${ourSellETHRate}`)
+      const { base: sellETHAmount, quote: buySUSDAmount } = getUnlimitedOrderAmounts(
+        formatedRate * (1 + sETHTosUSDFee),
+        sUSD.decimals,
+        sETH.decimals
+      )
+      buyAmounts.push(buySUSDAmount)
+      sellAmounts.push(sellETHAmount)
+      buyTokens.push(sUSD)
+      sellTokens.push(sETH)
+    } else {
+      console.log(`Not placing sell sETH order, our rate of ${ourSellETHRate.toFixed(2)} is too high for exchange.`)
+    }
 
-    // Avoid querying exchange by tokenAddress for fixed tokenId
-    const buyTokens = [sETH, sUSD].map((token) => token.exchangeId)
-    const sellTokens = [sUSD, sETH].map((token) => token.exchangeId)
+    if (buyAmounts.length > 0) {
+      // Fetch auction index and declare validity interval for orders.
+      // Note that order validity interval is inclusive on both sides.
+      const batchId = (await exchange.getCurrentBatchId.call()).toNumber()
+      const validFroms = Array(buyAmounts.length).fill(batchId)
+      const validTos = Array(buyAmounts.length).fill(batchId)
 
-    const gasPrices = await (await fetch(gasStationURL[networkId])).json()
-    const scaledGasPrice = parseInt(gasPrices[argv.gasPrice] * argv.gasPriceScale)
-    console.log(`Using current "${argv.gasPrice}" gas price scaled by ${argv.gasPriceScale}: ${scaledGasPrice}`)
-    await exchange.placeValidFromOrders(buyTokens, sellTokens, validFroms, validTos, buyAmounts, sellAmounts, {
-      from: account,
-      gasPrice: scaledGasPrice,
-    })
+      const gasPrices = await (await fetch(gasStationURL[networkId])).json()
+      const scaledGasPrice = parseInt(gasPrices[argv.gasPrice] * argv.gasPriceScale)
+      console.log(`Using current "${argv.gasPrice}" gas price scaled by ${argv.gasPriceScale}: ${scaledGasPrice}`)
+      await exchange.placeValidFromOrders(buyTokens, sellTokens, validFroms, validTos, buyAmounts, sellAmounts, {
+        from: account,
+        gasPrice: scaledGasPrice,
+      })
+    }
     callback()
   } catch (error) {
     callback(error)
--- a/scripts/synthetix/facilitate_trade.js
+++ b/scripts/synthetix/facilitate_trade.js
@@ -56,8 +56,8 @@ const estimationURLPrexix = {
   4: "https://dex-price-estimator.rinkeby.gnosis.io//api/v1/",
 }
 
-const estimatePrice = async function (buyTokenId, sellTokenId, amount, networkId) {
-  const searchCriteria = `markets/${buyTokenId}-${sellTokenId}/estimated-buy-amount/${amount}?atoms=true`
+const estimatePrice = async function (buyTokenId, sellTokenId, sellAmount, networkId) {
+  const searchCriteria = `markets/${buyTokenId}-${sellTokenId}/estimated-buy-amount/${sellAmount}?atoms=true`
   const estimationData = await (await fetch(estimationURLPrexix[networkId] + searchCriteria)).json()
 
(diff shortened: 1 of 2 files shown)


scripts/synthetix/facilitate_trade.js
```diff
@@ -50,6 +51,30 @@ const gasStationURL = {
   4: "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/",
 }
 
+const toWEI = function (value, decimals) {
+  const h = Math.floor(decimals / 2)
+  const a = new BN(10).pow(new BN(h))
+  const b = new BN(10).pow(new BN(decimals - h))
+
+  return new BN(value * a).mul(new BN(b))
+}
```
fedgiac, line 61: Why not using `toErc20Units` here? (You'd have to convert float to string with no more decimals than `decimals` first, or better create a `floatToErc20Unit` that does that.)
I really don't like this function, I believe there are things that could go wrong here because of float shenanigans.
If `decimals=6` and `value=10/260`, then the output is different than expected by more than 1%. This is because the BN * float operation in line 59 truncates to the nearest integer.
I believe the reason why `h` exists is to juggle the amounts in `value*a` so that for common input values the result is an integer smaller than `2^52` but not so small that you lose too much precision in the truncation. It feels quite fragile.
Also, because of rounding issues I wouldn't be really confident that if `h = Math.floor(decimals / 2)` then `Math.floor(h) + Math.floor(decimals-h)` is the same integer as `decimals`, as this code requires.

scripts/synthetix/facilitate_trade.js
```diff
@@ -50,6 +51,30 @@ const gasStationURL = {
...
+
+const BN = require("bn.js")
+// Could also use linkPrefix here.
+const estimationURLPrexix = {
+  1: "https://dex-price-estimator.gnosis.io//api/v1/",
+  4: "https://dex-price-estimator.rinkeby.gnosis.io//api/v1/",
```
fedgiac, line 65: Is the double slash in the links intended? It looks strange.

scripts/synthetix/facilitate_trade.js
```diff
@@ -73,44 +98,75 @@ module.exports = async (callback) => {
...
-    await exchange.placeValidFromOrders(buyTokens, sellTokens, validFroms, validTos, buyAmounts, sellAmounts, {
-      from: account,
-      gasPrice: scaledGasPrice,
-    })
+    // Initialize all arrays needed for potential order placement.
+    const [buyAmounts, sellAmounts, buyTokens, sellTokens] = [[], [], [], []]
+
+    // Compute buy-sell amounts based on unlimited orders with rates from above when the price is right.
+    const ourBuyETHRate = formatedRate * (1 - sUSDTosETHFee)
+    if (ourBuyETHRate < exchangeBuyETHPrice) {
```
fedgiac, line 123: Shouldn't you be comparing the buy amount of the exchange with our *sell* rate instead? "They are willing to buy at a price higher than we are willing to sell" is what I'd expect here. Same below.

scripts/synthetix/facilitate_trade.js
bh2smith: Yes, I would be happy to use our already existing functionality... I used this because I didn't realize we already had something. Will find it and make use of it asap!

scripts/synthetix/facilitate_trade.js
bh2smith: It sure does look strange.. but that is the link... Even if I remove the double slash it redirects me to the double slash.

scripts/synthetix/facilitate_trade.js
bh2smith: I found this very confusing at the time of writing as well. If the exchange is offering a price lower than what we are willing to pay, then we place an order.

 I must admit that I made the  inequalities based on the fact that there were no orders being placed at the time of coding. I also wrote the logs in a way that would make me confident in the inequalities as they are:

```
Oracle sETH Price (in sUSD) 243.3257626212688802
Exchange buy  sETH price (in sUSD) 246.6144973941891
Exchange sell sETH price (in sUSD) 233.72582778494302
Not placing buy  sETH order, our rate of 242.60 is too low  for exchange.
Not placing sell sETH order, our rate of 244.06 is too high for exchange.
```

I will double check this, but at the time of writing I was pretty sure I had the inequalities the correct way around.

bh2smith: I do agree this inequality looks completely backwards though.

bh2smith: Oh I see what you mean. We should be comparing our buy price with their sell price and vice versa.

bh2smith: Ok, so I haven't exactly got everything worked out here yet, but I would like to clarify that exchange price values here are the prices at which a user to realize. This can be seen by the fact that `exchangeBuyPrice > exchangeSellPrice` when there are no overlapping orders. That is, if you are buying ETH you can expect to pay `exchangeBuyPrice` and vice versa. So, although the values might appear confusing at first, or possibly the semantics are unclear, I think the values being compared make sense... Whether the inequalities are backwards is a different story I need to look into closer.

bh2smith: Turns out the comparison variables were correctly named, but this particular line in the code was missing a multiplicative inversion...

scripts/synthetix/facilitate_trade.js
```diff
@@ -73,44 +88,89 @@ module.exports = async (callback) => {
...
-    await exchange.placeValidFromOrders(buyTokens, sellTokens, validFroms, validTos, buyAmounts, sellAmounts, {
-      from: account,
-      gasPrice: scaledGasPrice,
-    })
+    // Initialize order array.
+    const orders = []
+
+    // Compute buy-sell amounts based on unlimited orders with rates from above when the price is right.
+    const ourBuyETHRate = formatedRate * (1 - sUSDTosETHFee)
+    if (ourBuyETHRate > 1 / exchangeBuyETHPrice) {
```
fleupold, line 114: When I first reviewed this I was also mostly focussing on the print statements which convinced me the inequality was correct. The fact that we are inverting the value in the print statement and have to remember to do it here as well is a bit confusing. Maybe we can make it clearer by naming the variable `exchangeBuyETHPriceInEth` so that the unit of measurement is clearer.

In natural language, I'd agree with @fedgiac that we want to place an order if "snxBuyEthRateInSusd" is larger than the "gpSellEthRateInSusd" that is the synthetix exchange is willing to buy eth at a price higher than GP is willing to sell it for (meaning we have a match). However since we are placing sell orders we would have to place a sell sUSD order.

So maybe the logic should be `if snxSellSusdRateInEth < gpBuySusdRateInEth` (SNXchange is willing to sell for less than GP is willing to buy for), place a sell SUSD order.

bh2smith: I would be happy and willing to change the internal variable names, but I would like to keep the print statements in a human readable prices (i.e. the price of ETH in USD so that the user has something to relate to). I hope that I am never able to interpret the price of USD in ETH unless they eventually become one-to-one! At the time of writing this, the price of USD in ETH is at 0.004112518506

scripts/synthetix/facilitate_trade.js
```diff
@@ -73,44 +88,89 @@ module.exports = async (callback) => {
     // Note that sUSD always has a price of 1 within synthetix protocol.
     const exchangeRate = await snxjs.ExchangeRates.rateForCurrency(sETHKey)
     const formatedRate = snxjs.utils.formatEther(exchangeRate)
-    console.log("sETH Price", snxjs.utils.formatEther(exchangeRate))
+    console.log("Oracle sETH Price (in sUSD)", formatedRate)
+
+    const exchangeBuyETHPrice = await estimatePrice(
```
fedgiac, line 93: The name of the variable sounds to me as "the price at which the exchange *buys* ETH", but this is more "the price at which the exchange *lets you buy* ETH". What about `exchangePriceWhenBuyingUsdForEth`? Not nice, I admit, but I hope clearer. Same thing for the other exchange price. It should also be clearer that the price is in ETH, so 1/what you are normally used to.

scripts/synthetix/facilitate_trade.js
```diff
@@ -73,44 +88,89 @@ module.exports = async (callback) => {
...
+
+    // Compute buy-sell amounts based on unlimited orders with rates from above when the price is right.
+    const ourBuyETHRate = formatedRate * (1 - sUSDTosETHFee)
+    if (ourBuyETHRate > 1 / exchangeBuyETHPrice) {
+      // If we are willing to pay more than the exchange quotes as buy price.
+      console.log(`Placing an order to buy sETH at ${ourBuyETHRate}`)
+      const { base: sellSUSDAmount, quote: buyETHAmount } = getUnlimitedOrderAmounts(
+        1 / ourBuyETHRate,
+        sETH.decimals,
+        sUSD.decimals
```
fedgiac, line 120: Not an issue of this PR, but I wonder if it makes sense to have the input to this function be an object, so you have to specify `quoteTokenDecimal: sETH.decimals` and so on. You'd see immediately if there had been mismatches.

scripts/synthetix/facilitate_trade.js
```diff
@@ -73,44 +88,89 @@ module.exports = async (callback) => {
...
+    } else {
+      console.log(`Not placing buy  sETH order, our rate of ${ourBuyETHRate.toFixed(2)} is too low  for exchange.`)
+    }
+
+    const ourSellETHRate = formatedRate * (1 + sETHTosUSDFee)
+    if (ourSellETHRate < exchangeSellETHPrice) {
+      // If we are willing to sell at a price less than exchange quote.
+      console.log(`Placing an order to sell sETH at ${ourSellETHRate}`)
+      const { base: sellETHAmount, quote: buySUSDAmount } = getUnlimitedOrderAmounts(
+        formatedRate * (1 + sETHTosUSDFee),
```
fedgiac, line 137: Very nitpick, but why not using `ourSellETHRate` here, similar to the previous call to `getUnlimitedOrderAmounts`?
```suggestion
        ourSellETHRate,
```

scripts/utils/printing_tools.js
```diff
@@ -31,6 +31,18 @@ const toErc20Units = function (amount, decimals) {
...
+/**
+ * A generalized version of "toWei" for tokens with an arbitrary amount of decimals.
+ * If the decimal representation has more decimals than the maximum amount possible, then the extra decimals are truncated.
+ *
+ * @param {number} amount Float representation for the amount of some ERC20 token
+ * @param {(number|string|BN)} decimals Maximum number of decimals of the token
+ * @returns {BN} number of token units corresponding to the input amount
+ */
+const floatToErc20Units = function (amount, decimals) {
+  return toErc20Units(amount.toString(), decimals)
```
fedgiac, line 43: I think this function will fail for `amount=10^21` because as a string it's `1e+21`, which is not supported by `toErc20Units`. `amount=10^20` would still work, however. 
It would be good to always print the float with decimal notation, I've no idea how to do that in JS however.

fedgiac: Oh, and `toErc20Units` fails if the number of decimals is more than what you would expect from the token. So, if you try to do `toErc20Units("1.111", 2)` it will fail. What I would do here in e.g. Python would be to tell my parser to print exactly `decimals` digits.

scripts/synthetix/facilitate_trade.js
```diff
@@ -50,6 +51,20 @@ const gasStationURL = {
...
+const estimationURLPrexix = {
+  1: "https://dex-price-estimator.gnosis.io//api/v1/",
+  4: "https://dex-price-estimator.rinkeby.gnosis.io//api/v1/",
+}
+
+const estimatePrice = async function (buyTokenId, sellTokenId, amount, networkId) {
+  const searchCriteria = `markets/${buyTokenId}-${sellTokenId}/estimated-buy-amount/${amount}?atoms=true`
+  const estimationData = await (await fetch(estimationURLPrexix[networkId] + searchCriteria)).json()
+
+  return estimationData.buyAmountInBase / estimationData.sellAmountInQuote
```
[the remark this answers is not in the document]
bh2smith: I think the expected output when specifiec sell amount cant be filled is precisely what we are seeing in the case of rinkeby test plan (above). i.e. infinity and zero.

You can try these out yourself buy adding zeros to the following query:


https://dex-price-estimator.gnosis.io//api/v1/markets/8-9/estimated-buy-amount/10000000000000000000?atoms=true

> If so, it might make sense to check that estimationData.sellAmountInQuote === amount for our purposes.

In any case, yes we can throw in this assertion but I believe it will always be true.

scripts/synthetix/facilitate_trade.js
bh2smith: This is certainly something I would like to tackle and it has been mentioned (in passing) during the TS migration of `dex-contracts`. Funny that this actually used to be the case, but I had requested it be removed since only the decimals attribute of the token object was ever used. 

In reality, from an object oriented programming perspective `Token` should be a class and this method should be implemented on that class, then it would be `token.getUnlimitedOrderAmounts(price, otherToken)`

scripts/utils/printing_tools.js
bh2smith: >  I think this function will fail for amount=10^21

I do not expect amounts to be of this size, but will be happy to put an assertion.

> Oh, and toErc20Units fails if the number of decimals is more than what you would expect from the token.

We also will not experience these issues in this script since the decimals are fixed at 18. Perhaps I should have left the `toWEI` implementation that I already had to keep this script entirely self contained.

It seems to me that you have suggested I use your awesome helper method and then pointed out a bunch of problems with it? Would you agree that I revert the use of this helper method in favour of the self-contained `toWEI`, put more assertions or include some comments about how the edge cases are unexpected to occur in this particular script?

scripts/synthetix/facilitate_trade.js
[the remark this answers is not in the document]
fedgiac: I see! Then I wouldn't change it, at most a comment explaining this.

scripts/utils/printing_tools.js
fedgiac: We could do that for now, but I'd definitely open an issue about it so that we'll eventually have a reliable way to go from token units to decimal representation and vice versa without relying on some hack. We could extend `toErc20Units` so that it accepts float. Feel free to assign me to it.

scripts/synthetix/facilitate_trade.js
```diff
@@ -50,6 +51,30 @@ const gasStationURL = {
   4: "https://safe-relay.rinkeby.gnosis.io/api/v1/gas-station/",
 }
 
+const toWEI = function (value, decimals) {
```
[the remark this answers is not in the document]
bh2smith: I have created an issue in our repo for the absurd implementation provided by web3js #302 and considering that we have hard coded the token into this script will temporarily use an ever so slightly generalized implementation that makes use of web3js for this. That is, I have introduced a small mapping at the top of the file

```js
const unitMapping = {
  18: "ether",
}
```

and make calls to the amount formatting as follows:

```js
web3.utils.toWEI(MIN_SELL_USD, unitMapping[sUSD.decimals])
```

I hope this is sufficient for now while we work on a more appropriate solution.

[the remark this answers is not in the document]
bh2smith: Never mind the last comment, this monkey patch actually wind up being much uglier than expected. I will persist in using the utility function we already have here in the project and offload the necessary repairs to the issue linked above.

scripts/synthetix/facilitate_trade.js
```diff
@@ -51,6 +51,26 @@ const gasStationURL = {
...
+  const estimationData = await (await fetch(estimationURLPrexix[networkId] + searchCriteria)).json()
+
+  return estimationData.buyAmountInBase / estimationData.sellAmountInQuote
+}
+
+const MIN_SELL_USD = 10
+
+/* All prices refered to hear are for sETH in sUSD.
+ * The term "our" referres to prices we are buying and selling for while
+ * the term "their" refers to the buy and sell prices offered by the exchange.
```
fleupold, line 70: Can you be explicit about what is Gnosis Protocol and what is the SNX Exchange?

scripts/synthetix/facilitate_trade.js
```diff
@@ -51,6 +51,26 @@ const gasStationURL = {
...
 }
 
+const estimationURLPrexix = {
+  1: "https://dex-price-estimator.gnosis.io//api/v1/",
+  4: "https://dex-price-estimator.rinkeby.gnosis.io//api/v1/",
+}
+
+const estimatePrice = async function (buyTokenId, sellTokenId, sellAmount, networkId) {
+  const searchCriteria = `markets/${buyTokenId}-${sellTokenId}/estimated-buy-amount/${sellAmount}?atoms=true`
+  const estimationData = await (await fetch(estimationURLPrexix[networkId] + searchCriteria)).json()
```
fleupold, line 61: nit: shouldn't need to await on json() as it's likekly synchronous.

scripts/synthetix/facilitate_trade.js
```diff
@@ -51,6 +51,26 @@ const gasStationURL = {
...
+
+  return estimationData.buyAmountInBase / estimationData.sellAmountInQuote
+}
+
+const MIN_SELL_USD = 10
+
+/* All prices refered to hear are for sETH in sUSD.
```
fleupold, line 68: nit: here. And also this is not the case for the inverted ones right?

scripts/synthetix/facilitate_trade.js
bh2smith: and fix the typo in the word "refers". Yes!

scripts/synthetix/facilitate_trade.js
bh2smith: I needs to be there. I have dealt with this very same comment in the past.

fleupold: You are right, it's needed https://github.com/node-fetch/node-fetch/blob/fa627f4b0cc60570037c2fc8db88b1771ae185b4/example.js#L14

I assume the first promise returns as soon as there is the first bit of response data but json() requires the full response to be there so it is a separate promise. Clever, but a bit non-standard.

bh2smith: Good stuff! I only hope I can remember how to find this link again later when it comes up again.

PR #306, opened 2020-06-11
Verify trading account has sufficent sellToken

### LINKED ISSUE #304: [Synthetix] Place orders when sufficient balance is available
Currently the script facilitating syntheitx trades  places orders purely on the price comparisons, but does not take its own balance into consideration. This would be a simple way to save some gas (i.e. not placing orders to sell tokens it does not have).

bh2smith:
This change simply checks the bots balance right before constructing the order to place. Note that the balance checks are only made if an order is about to be placed which saves some EVM calls. Also, the logs will be more coherent (stating that the order is about to be placed, but the balance was insufficient if this is the case while not mentioning the balance issue if it wasn't going to try and place an order anyway).

Closes #304

### DIFF
--- a/scripts/synthetix/facilitate_trade.js
+++ b/scripts/synthetix/facilitate_trade.js
@@ -97,17 +97,13 @@ module.exports = async (callback) => {
     const formatedRate = snxjs.utils.formatEther(exchangeRate)
     console.log("Oracle sETH Price (in sUSD)", formatedRate)
 
-    const theirSellPriceInverted = await estimatePrice(
-      sETH.exchangeId,
-      sUSD.exchangeId,
-      floatToErc20Units(MIN_SELL_USD, sUSD.decimals),
-      networkId
-    )
+    const minSellsUSD = floatToErc20Units(MIN_SELL_USD, sUSD.decimals)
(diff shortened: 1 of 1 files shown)


scripts/synthetix/facilitate_trade.js
```diff
@@ -121,33 +117,47 @@ module.exports = async (callback) => {
...
+        )
+        orders.push({
+          buyToken: sETH.exchangeId,
+          sellToken: sUSD.exchangeId,
+          buyAmount: buyETHAmount,
+          sellAmount: sellSUSDAmount,
+        })
+      } else {
+        console.log(`Warning: Insufficient sUSD (${sUSDBalance} < ${minSellsUSD}) for order placement.`)
+      }
```
fedgiac, line 136: I'd consider moving line 120 inside the if and remove "but verifying sUSD balance first", unless you really want to stress that this check is being done in the log. Same change in the if below.

bh2smith: I want the user to know that an preset was being attempted. But that the balance failed. If the first log is moved inside of the balance verification the user might not know what happened.

PR #309, opened 2020-06-13
Safe Airdrop; distribute funds to arbitrarily many receivers

bh2smith:
For the upcoming **Gnosis Protocol** incentive program pay out, these changes implement a convenient script which can parse a JSON "Transfer File" and distribute tokens to arbitrarily (bounded by block gas limit) many accounts in a single Ethereum transaction.

**Note that** this functionality is now tested and we have found that exactly 218 transfers can be made in a single transaction while using strictly below 4M Gas.

**Questions:**

- Should we hard code maximum number of transfers (i.e. ~200) and if so, should we give the user the ability to proceed anyways? 
- Same question with the balance verification? Should we allow the user to proceed anyway if the balance is determined to be insufficient? I would say NO to this.
- One important test here is "Ensures no partial transfers happen on execution failure". However, the only way to test this was to allow the possibility to run the function in "unsafe" mode. Its like introducing the possibility for failure. Do we really want this?

**TODO:**
- [x] Accept CSV format: This conversion can be found in a separate repo - https://github.com/anxolin/challenge-payment/pull/1
- [x] allow `dry-run` functionality that confirms pending transaction with equivalent data already awaiting execution 
- [x] Add script usage examples in README.

**Test Plan**

To run the bundled transfer propsal
```sh
npx truffle exec scripts/airdrop.js --masterSafe 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --transferFile examples/sampleTransferFile.json --network rinkeby
```

Should emit logs like:
```
Using network 'rinkeby'.

Preparing transaction data...
Transferring 271828 USDC to 0x0000000000000000000000000000000000000001 from 0xd939...4c
Transferring 314159 USDC to 0x0000000000000000000000000000000000000002 from 0xd939...4c
Transferring 10000 USDC to 0x0000000000000000000000000000000000000003 from 0xd939...4c
Transferring 10000 USDC to 0x0000000000000000000000000000000000000004 from 0xd939...4c
Transfer bundle contains 4 elements and sends 1 distinct tokens.
Ensuring sufficient USDC balance for this transfer...
Balance verification passed.
Are you sure you want to send this transaction to the EVM? [yN] y
Signing and posting multi-send transaction 0x0f3a1faa1f83bb288bc1d7d703daf908ff912561cb2fd5b2eb3e117e61676d1b from proposer account 0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1
Transaction awaiting execution in the interface https://rinkeby.gnosis-safe.io/app/#/safes/0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c/transactions
```

To do a "verification" run simply add the argument `--verify` and observe the difference in the last two lines of the logs.

```sh
npx truffle exec scripts/airdrop.js --masterSafe 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --transferFile examples/sampleTransferFile.json --network rinkeby --verify
```

```
Using network 'rinkeby'.

Preparing transaction data...
Transferring 271828 USDC to 0x0000000000000000000000000000000000000001 from 0xd939...4c
Transferring 314159 USDC to 0x0000000000000000000000000000000000000002 from 0xd939...4c
Transferring 10000 USDC to 0x0000000000000000000000000000000000000003 from 0xd939...4c
Transferring 10000 USDC to 0x0000000000000000000000000000000000000004 from 0xd939...4c
Transfer bundle contains 4 elements and sends 1 distinct tokens.
Ensuring sufficient USDC balance for this transfer...
Balance verification passed.
Verifying transaction
The transaction matches a transaction in the interface! You can sign the transaction with nonce 144 here: https://rinkeby.gnosis-safe.io/app/#/safes/0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c/transactions
```

To see the proceed anyway prompt, try this:

```sh
npx truffle exec scripts/airdrop.js --fundAccount 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --transferFile examples/twoHundredOneTransfers.json --network rinkeby
```


josojo:
Very cool that you are coding such a bundled multi-transfer so quickly down. I think you are on the right way, only tests are missing.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -360,6 +361,66 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+        log(`Ensuring sufficient ${token.symbol} balance for this transfer...`)
+        if (masterBalance.lt(cumulativeAmounts.get(tokenAddress))) {
+          throw new Error(
+            `Master Safe has insufficient ${token.symbol} balance (${masterBalance.toString()} < ${cumulativeAmounts
+              .get(tokenAddress)
+              .toString()})`
+          )
+        }
+      }
+    }
```
fedgiac, line 419: Here we await the balances of the master Safe one token after the other. You could replace the for loop with a `Promise.all`: 
```suggestion
    if (!unsafe) {
      // Ensure sufficient funds.
      await Promise.all(
        uniqueTokens.map(async (tokenAddress) => {
          const token = await tokenInfo[tokenAddress]
          const masterBalance = await token.instance.balanceOf(masterAddress)
          log(`Ensuring sufficient ${token.symbol} balance for this transfer...`)
          if (masterBalance.lt(cumulativeAmounts.get(tokenAddress))) {
            throw new Error(
              `Master Safe has insufficient ${token.symbol} balance (${masterBalance.toString()} < ${cumulativeAmounts
                .get(tokenAddress)
                .toString()})`
            )
          }
        })
      )
    }
```
For extra efficiency (~1s) you could retrieve the master balances during the "await" before this loop.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -360,6 +361,66 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+
+    const uniqueTokens = uniqueItems(transferList.map((t) => t.tokenAddress))
+    const tokenInfo = fetchTokenInfoForFlux(transferList)
+
+    // Will be used to make sufficient balance assertion before transfer.
+    const cumulativeAmounts = new Map(uniqueTokens.map((address) => [address, new BN(0)]))
+    const transactions = await Promise.all(
+      transferList.map(async (transfer) => {
+        const token = await tokenInfo[transfer.tokenAddress]
+        let unitAmount = new BN(transfer.amount)
```
fedgiac, line 387: This will fail for realistic amounts (assuming `useWei=true`) with your current JSON format.
To see what I mean, add this to `sampleTransferFile.json`:
```
  {
    "amount": 1000000000000000000000,
    "tokenAddress": "0xc778417e063141139fce010982780140aa0cd5ab",
    "receiver": "0x0000000000000000000000000000000000000005"
  }
```
Note that the amount is just 1000 for a token with 18 decimals.
That's because there are no quotes around the amount value, so it's read as a number, and `BN` rightfully fails to parse float amounts larger than `2^52`.
Note that just converting this amount to string and passing it to the same function might not be enough. Because of float precision this number could change when made into a string, and a user who deposited the exact amount in his master Safe would get an "unfixable" insufficient balance error when running the scripts even if the balance is enough.
Not sure what's best here. Since the JSON is our user interface, I'd say the script should fail if `typeof(transfer.amount)` is `number` with some helpful error like "wrap your amount in quotes" and only accept strings.

fedgiac (commented):
Nice PR! I like how many already existing components could be reused.

The logic looks good. Two minor comments:
1. You might want to point out in the readme that the max amount of transfers you can do in one transaction depends on the token. Some tokens are much cheaper than others to move around.
2. I'd move the sample transaction file to the `examples` folder for consistency. (Or get rid of the examples folder, but then you have to tag this PR as tutorial breaking I believe, as well as change the readme.

scripts/utils/trading_strategy_helpers.js
fedgiac: I'd add a "large number" test to catch this in the future.

bh2smith: Good to point this out! I like the `typeof` suggestion and will go with that.

bh2smith:
Thanks for your careful review @fedgiac!

> point out ... that the max amount of transfers depends on the token. Some tokens are much cheaper than others to move around.

Yes, I would also probably like to considering putting a hard limit within the script with a "proceedAnyways" what do you think about that.


> move the sample transaction file to the examples folder for consistency.

Sure thing... just gotta find this examples folder.

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+    demandOption: true,
+  })
+  .option("verify", {
+    type: "boolean",
+    default: false,
+    describe: "Do not actually send transactions, just simulate their submission",
+  })
+  .option("useWei", {
+    type: "string",
+    describe: "file name (and path) to the list transfers",
```
alfetopito, line 25: Description does not match parameter.

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+    const masterSafe = await GnosisSafe.at(argv.masterSafe)
+
+    const transfers = JSON.parse(await fs.readFile(argv.transferFile, "utf8"))
+
+    console.log("Preparing transaction data...")
+    const transaction = await buildTransferDataFromList(masterSafe.address, transfers, argv.useWei, false, true)
+
+    if (!argv.verify) {
+      const answer = await promptUser("Are you sure you want to send this transaction to the EVM? [yN] ")
+      if (answer == "y" || answer.toLowerCase() == "yes") {
```
alfetopito, line 41: I would also consider lowercasing the answer when comparing against `y`

```suggestion
      if (answer.toLowerCase() == "y" || answer.toLowerCase() == "yes") {
```


Of if you want to be fancy, use a switch :)

```ts
switch (answer.toLowerCase()) {
  case `y`:
  case `yes`:
  case `yeah`:
  case `yup`:
  case `yep`:
  case `hell yeah`:
  case `sure`:
  case `why not, eh?`:
    ...
  default:
    //no
}

File: scripts/typedef.js [hunk 5]
```diff
@@ -34,6 +34,21 @@
...
+ * Example:
+ * {
+ *   amount: 100,
+ *   tokenAddress: 0x0000000000000000000000000000000000000000,
+ *   toAddress: 0x0000000000000000000000000000000000000001
+ * }
+ *
+ * @typedef Transfer
+ * @type {object}
+ * @property {number|string|BN} amount integer denoting amount to be transfered
```
alfetopito, line 47: Only integers? It does not match the example where you have decimals.

scripts/airdrop.js
bh2smith: I like the suggestion, but it actually makes me want to restrict it now to only lower case `y` because that is what is suggested in the prompt (i.e. `[yN]`)

scripts/typedef.js
bh2smith:
```suggestion
 * @property {number|string|BN} amount value denoting amount to be transferred
```

scripts/airdrop.js
bh2smith: Good eye! Thanks for catching this.

scripts/utils/trading_strategy_helpers.js
anxolin: Yep, Agree with @fedgiac. I think in general, it's better to use strings, specially for the mentioned reason, and also for the error of working with floating point instead of `BN` or `BigNumber`

Note that `BigNumber` can be used to handle decimal numbers, `BN` for big integers. Both allow to create instances of them from a `string`

Example of one issue with floats that we don't get if we use `BigNumber` for example:
image

scripts/README.md
```diff
@@ -62,6 +60,25 @@ The fleet size should be smaller than or equal to 20, in order to ensure that th
 Please document the displayed bracket-trader addresses. They are required for future withdrawals.
 They can also be retrieved from the created transactions. However, since this is a manual process, it is quite cumbersome to extract them right now.
 
+### Safe Airdrop
+
+Create your own transferFile (or use our sample `scripts/data/sampleTransferFile.json`)
```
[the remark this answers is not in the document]
anxolin: Same comment as this for the the reference to this README: `See `scripts/README.md` for details.`

I think it should be a link.

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+    type: "string",
+    describe: "file name (and path) to the list transfers",
+    demandOption: true,
+  })
+  .option("verify", {
+    type: "boolean",
+    default: false,
+    describe: "Do not actually send transactions, just simulate their submission",
+  })
+  .option("useWei", {
```
anxolin: I traditionally call this "wei", although I've been told wei is only for ETH, so it's atoms
That's the naming convention in the price estimator for example

Not really important anyways, feel free to not change. Anyone in the space understands :)

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+    type: "boolean",
+    describe: "boolen flag used to indicate if the amounts provided are already in atoms",
```
anxolin: typo in `boolen`, actually. The type says already boolean. I think we can remove the type from the description "Flag used to ..."

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+    demandOption: false,
+  }).argv
+
+module.exports = async (callback) => {
+  try {
+    const GnosisSafe = artifacts.require("GnosisSafe")
+    const masterSafe = await GnosisSafe.at(argv.masterSafe)
+
+    const transfers = JSON.parse(await fs.readFile(argv.transferFile, "utf8"))
```
anxolin, line 29: For reading a JSON you can just do a `require`

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+
+    console.log("Preparing transaction data...")
+    const transaction = await buildTransferDataFromList(masterSafe.address, transfers, argv.useWei, false, true)
```
anxolin: ubernit: This comment doesn't belong to this PR. But just want to take the opportunity to mention something, since I see it in many parts of the codebase. I believe this way of passing parameters is considered a bad practice.

The reason is that, as a reader/reviewer, I can guess what the first 3 parameters are, but for the last 2 I would need to jump to buildTransferDataFromList implementation:
`buildTransferDataFromList(masterSafe.address, transfers, argv.useWei, false, true)` 

Further more, if I by mistake invert the order `false, true` (instead of `true`, `false``), cause I though the order was different.

For any non trivial function signature (more than one param), we could use named parameters.
```js
buildTransferDataFromList({ 
  masterAddress: masterSafe.address, 
 transferList: transfers, 
 useWei: argv.useWei, 
 unsafe: false,
 debug: true
})
```

Actually, it would allow to define default values for `unsafe` and `debug` so the call is simpler.

scripts/airdrop.js
```diff
@@ -0,0 +1,54 @@
...
+        await signAndSend(masterSafe, transaction, argv.network)
+      }
+    } else {
+      console.log("Verifying transaction")
+      await transactionExistsOnSafeServer(masterSafe, transaction, argv.network, (await masterSafe.nonce()).toNumber())
+    }
+
+    callback()
+  } catch (error) {
+    console.log(error.response)
```
anxolin, line 46: `console.error` instead of  console.log`

why printing `error.response` instead of `error`?

scripts/utils/trading_strategy_helpers.js
bh2smith: I have decided to remove useWEI all together because it was not worth the extra complexity.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -360,6 +361,67 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+              `Master Safe has insufficient ${token.symbol} balance (${masterBalance.toString()} < ${cumulativeAmounts
+                .get(tokenAddress)
+                .toString()})`
+            )
+          }
+        })
+      )
+      log("Balance verification passed")
+    }
+    return buildBundledTransaction(transactions)
```
anxolin, line 418: Great function here the `buildBundledTransaction`

Just an idea, if you are interested for future versions :) 
`execTransactionData` does 2 things:
- Prepare the multisend call, with all the encoded smaller calls
- Send the transaction

If we just remove the part of sending the transaction (or move it to somewhere else), we could for example print the details of the transaction. This would allow anyone to execute this transaction WITHOUT any private key. Just signing it with his Ledger or Gnosis Safe in MyEtherWallet

anxolin (approved):
Amazing! Surprised how fast you put something like this together. Also nice tests.
I would suggest calling this `multisend` instead of `airdrop`

My comments are just nitpicks. The most important thing to address I would say is using strings for the values.
This script will make 80 people happy soon 🎉

Related to your questions:
> Should we hard code maximum number of transfers (i.e. ~200) and if so, should we give the user the ability to proceed anyways?

I think so. Any known limit should be added. Future versions couldprocesss the payments in batches, but I think it's already a great start.
I would allow to set it by param, because it can depend on the token. We could just have a safe limit by default.

> Same question with the balance verification? Should we allow the user to proceed anyway if the balance is determined to be insufficient? I would say NO to this.

Agree

> One important test here is "Ensures no partial transfers happen on execution failure". However, the only way to test this was to allow the possibility to run the function in "unsafe" mode. Its like introducing the possibility for failure. Do we really want this?

As a feature not. Functionally I don't any reason to skip the validation. If it helps you doing your test, looks fine too me.

examples/sampleTransferFile.json
```diff
@@ -0,0 +1,22 @@
+[
+    {
+        "amount": 0.271828,
```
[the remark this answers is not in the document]
bh2smith: Already taken care of. Strings FTW!

scripts/README.md
```diff
@@ -62,6 +60,25 @@ The fleet size should be smaller than or equal to 20, in order to ensure that th
...
+With a master safe that you own execute:
+
+```sh
+npx truffle exec scripts/airdrop.js --masterSafe 0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --transferFile scripts/data/sampleTransferFile.json --network rinkeby
```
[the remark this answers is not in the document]
bh2smith: ~~Master Safe is the universal term for this particular project (which was originally only for bracket strategies - aka CMM). When we migrate this "multisend" functionality into its own standalone repo, then I would change the name here.~~

Actually. I think this would be a good place to use the term `fundAccount`.

scripts/airdrop.js
bh2smith: Show me!

scripts/utils/trading_strategy_helpers.js
bh2smith: We do not want to mix send transaction with this other thing because we still need a way to prompt the user after we have acquired all the transaction details. "This is what you are about to sign would are you sure?"

scripts/README.md
```diff
@@ -62,6 +60,25 @@ The fleet size should be smaller than or equal to 20, in order to ensure that th
 Please document the displayed bracket-trader addresses. They are required for future withdrawals.
 They can also be retrieved from the created transactions. However, since this is a manual process, it is quite cumbersome to extract them right now.
 
+### Safe Airdrop
+
+Create your own transferFile (or use our sample `scripts/data/sampleTransferFile.json`)
+With a master safe that you own execute:
```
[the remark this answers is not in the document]
bh2smith: > For the next step you will need the address of your Gnosis Safe, if you don't have one yet, create it in https://gnosis-safe.io/app

Great suggestion, this is already at the top of the document in the "Prerequisites section".

scripts/airdrop.js
fedgiac: I wouldn't do this: a big difference between `require` and `fs.readFile` is how it interprets relative path strings. The former makes the path relative to the path of the script, while the latter makes the path relative to the working directory of the user when executing the script. SInce this path is inputted by the user, it should be relative to the private working path.

scripts/airdrop.js
fedgiac: In hindsight, I'd like to have most of the functions of this projects handled like that.

scripts/airdrop.js
bh2smith: I have removed useWEI as it is too much unnecessary maintenance.

scripts/airdrop.js
bh2smith: I agree completely about this, but would like to save these changes for a separate PR (i.e. refactoring).

scripts/airdrop.js
bh2smith: I don't know. I can change this.

scripts/airdrop.js
bh2smith: In that case, I am going to leave this the way it is, since it works and there is no real problem here.

scripts/airdrop.js
bh2smith: > In hindsight, I'd like to have most of the functions of this projects handled like that.

Once this becomes a typescript project we wont need to do any of this. Essentially what is being proposed is like a hack in JS to make unsafe code ever so slightly safer than it is. If we are going to start doing this, we should just migrate to typescript where we aren't trying to trick our IDE.

anxolin: Even for typescript projects is a good pattern.

Yes, no need to address it. It's just someone I wanted to mention, since I haven't been reading that many PRs from this repo, and I just thought it was good to mention it.

anxolin:
I think it's not required for this GNO challenge, cause for me is more than good enough.

But small feedbacks are:

image

`1.` Show the amounts in units, not atoms.

`2.`Amazing if we can show a summary of the transfer per toke:
* Original: `Transfer bundle contains 125 elements and sends 1 distinct tokens.` 
* Proposal:
```
Contains 125 transfers with 1 distincts tokens:
   * 1000 GNO to 125 addresses
```

`3.` This is not needed AT ALL, but I know you enjoy learning about this libraries, so in case u don't know it https://www.npmjs.com/package/chalk would allow you to print the content in colours, so the transfers and amounts stand more. And for example, highlight more the message `Error: The transaction does not match any transaction in the interface!`

scripts/utils/trading_strategy_helpers.js
anxolin: I mean this for running it without the need of a PK, not sure if I explained myself very well. 

It would be a flag, that would, instead of sending the transaction, would print the `to, data, gas` so anyone can sign that MANUALLY in myetherwallet.com

This would remove the need of a `proposer` account.
I don't mean to include it in this PR. Just consider for the project, since it can make sense also for the CMM.

Happy to discuss it if personally too

bh2smith:
Thanks @anxolin for your thoughts.

> 1. Show the amounts in units, not atoms.

While I agree this might be nice, the point of this log message is to display to the user that the number of decimals has been taken into account and it is meant to give the user a sudden urge of hesitation and force them to double check that everything is as expected. Showing the units alone does not give me any confidence that the script is being executed correctly. I hope this is a reasonable assumption.


> 2.Amazing if we can show a summary of the transfer per token:
> Original: Transfer bundle contains 125 elements and sends 1 distinct tokens.
> Proposal: Contains 125 transfers with 1 distincts tokens:
>      * 1000 GNO to 125 addresses

I like this a lot and will try and incorporate this suggestion now.

As for your point number 3. I would create a nice to have issue with regards to this nice to have feature.

scripts/utils/trading_strategy_helpers.js
fedgiac: As it works right now `buildBundledTransaction` is very generic, in the sense that it's not the part of the code that requires the proposer. Anything can be bundled together and there is no need for a private key.
Note also that the functions `execTransaction` and `execTransactionData` are only used for testing. (We keep them in the main project in case we wanted to bypass the web interface in the future.) The actual transaction that is submitted to the blockchain is currently not built by us but by the Multisig web interface. So, it works like this:
1. we build a transaction that we would like to have executed on our master Safe (`fundAccount` here), "execute" means "call the Safe contract function `execTransaction`"
2. we tell the web interface "put up this transaction to be executed". In order to do this, we need a signature of at least one user, this is where the need of a proposer account come from
3. the web interface collects user signatures and builds the actual transaction that will be submitted to the blockchain from our proposed transaction and the signatures
4. someone submits the web interface's transaction to the blockchain from the web interface.

Our script is directly involved only in step 1. So, I don't really know what myetherwallet.com does, but I believe the functions here are already fairly general so they might already be working as you wish they would be.
Anyway, happy to have more eyes looking at this project, also happy to discuss if you want more details/have more ideas!

.env
```diff
@@ -0,0 +1,4 @@
+INFURA_KEY=9408f47dedf04716a03ef994182cf150
+PK=0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
+FUND_ACCOUNT=0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c
+TRANSFER_FILE=examples/sampleTransferFile.json
```
fedgiac, line 4: To other reviewers: this is intended and the key doesn't hold funds.

scripts/README.md
```diff
@@ -62,6 +66,37 @@ The fleet size should be smaller than or equal to 20, in order to ensure that th
...
+export FUND_ACCOUNT=<your gnosis safe>
+export TRANSFER_FILE=<path to your transfer file>
+```
+
+Alternatively, there is a sample [.env](.env) file that is not tracked by the project where you can paste these values and `source .env`
+
+With all configuration in place, we are ready to run the script.
+
+```sh
+npx truffle exec scripts/airdrop.js --fundAccount $FUND_ACCOUNT --transferFile $TRANSFER_FILE --network rinkeby
```
fedgiac, line 86: Very nitpicky, consistency with the other examples in the readme:
```suggestion
npx truffle exec scripts/airdrop.js --fundAccount=$FUND_ACCOUNT --transferFile=$TRANSFER_FILE --network=$NETWORK_NAME
```

fedgiac (approved):
Very nice PR!

scripts/README.md
bh2smith: Yes, this is fine, although I am not a fan of the `=` but will accept.

anxolin:
> it is meant to give the user a sudden urge of hesitation and force them to double check that everything is as expected

😂 "sudden urge"

I don't fully agree with the assumption. I think the user should have confidence in the script and it should display the numbers in human readable way. I think you have that concern, but a user would assume the decimals taken into account.

Having said that, are part of the nice to have, it doesn't bother me at all. And can always be revisited in the future if someone else thinks like me.

bh2smith:
Well lucky for you I actually incorporated your suggestion before the merge anyway.

PR #313, opened 2020-06-15
[Synthetix] Use BN comparison operators.

bh2smith:
Can't believe this was missed and super surprised this was actually working in some cases!

These values are of type `BN` so we need to use the correct comparison operators!


fleupold (approved):
Great find! 
I feel like I have also made this mistake before and this was one of the great examples of what typescript would have prevented us from doing wrong.

josojo (approved):
Yes, another pro for typescript

PR #324, opened 2020-06-23
Improve fetching token info for non-standard ERC20

### LINKED ISSUE #226: fetchTokenInfoFromExchange for non-standard tokens
Anxo pointed out that fetchTokenInfoFromExchange should be adapted soo that it also works, if the tokens don't implement the ERC20 protocol, as for example, they don't define the symbol interface.

fetchTokenInfoFromExchange should consider this...

fedgiac:
Closes #226.

Some tokens don't define their symbol or number of decimals, or have an output type that is not that of the functions as defined by `ERC20Detailed`.
This PR handles these exceptions as follows:
- check if token is in whitelist, if so return values from list.
- perform network request, if successful return result.
- if request fails for any reason, return a more helpful error message.

Note that the script still fails if the symbol or number of decimals is not known after a network request. This is because decimals are needed by the script, and a missing symbol could be telling the user that at the chosen address there is no ERC20 token.

### Test plan

I added tests for the newly introduced function.

I also tried out the command that appears in the linked issue (using WETH to make the tokens fail the price check), but only after removing the checks that the proposed is an owner. Then I ran the script with the default Ganache address and saw that the execution was successful until the WETH balance check.
```
# remove ownership check from complete_liquidity_provision!
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=23 --quoteTokenId=1 --lowestLimit=320 --highestLimit=720 --currentPrice=503 --masterSafe=0x5CC3a4C846aB8b0B78a729b8B1a50e2E0BB66740 --depositBaseToken=156.25 --depositQuoteToken=65000 --numBrackets=14 --network=mainnet
```
Expected result:
```
Using network 'mainnet'.

Using account: 0x627306090abaB3A6e1400e9345bC60c78a8BEf57
==> Performing safety checks
         chosen price: 503 WETH bought for 1 MKR
         dex.ag price: 2.08282529334201945798 WETH bought for 1 MKR
Error: MasterSafe 0x5CC3a4C846aB8b0B78a729b8B1a50e2E0BB66740 has insufficient balance for quote token 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2
Truffle v5.1.31 (core: 5.1.31)
Node v12.16.3
```

### DIFF
--- a/scripts/utils/hardcoded_tokens.json
+++ b/scripts/utils/hardcoded_tokens.json
@@ -1,0 +1,15 @@
+{
+  "1": {
+    "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359": {
+      "name": "Sai Stablecoin v1.0",
+      "symbol": "SAI",
+      "decimals": 18
+    },
+    "0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2": {
+      "name": "Maker",
+      "symbol": "MKR",
+      "decimals": 18
+    }
+  },
+  "4": {} 
+}
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -20,6 +20,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const { uniqueItems } = require("./js_helpers")
   const { DEFAULT_ORDER_EXPIRY, CALL } = require("./constants")
 
+  const ERC20 = artifacts.require("ERC20Detailed")
   const BatchExchange = Contract(require("@gnosis.pm/dex-contracts/build/contracts/BatchExchange"))
   const GnosisSafe = artifacts.require("GnosisSafe")
   const FleetFactory = artifacts.require("FleetFactory")
@@ -28,6 +29,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const exchangePromise = BatchExchange.deployed()
   const gnosisSafeMasterCopyPromise = GnosisSafe.deployed()
   const fleetFactoryPromise = FleetFactory.deployed()
+  const hardcodedTokensByNetwork = require("./hardcoded_tokens")
 
   /**
    * Returns an instance of the exchange contract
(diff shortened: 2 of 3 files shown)


scripts/utils/hardcoded_tokens.json
```diff
@@ -0,0 +1,15 @@
+{
+  "1": {
+    "0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359": {
+      "name": "Sai Stablecoin v1.0",
+      "symbol": "SAI",
+      "decimals": 18
+    },
+    "0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2": {
```
josojo, line 8: I checked both entries on etherscan. They are correct

scripts/utils/trading_strategy_helpers.js
```diff
@@ -75,6 +77,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     return orders != null
   }
 
+  /**
+   * Returns the symbol of the ERC20 token in input.
+   * Returns the velue from a list if present, otherwise executes a contract call.
```
josojo, line 82: value

scripts/utils/trading_strategy_helpers.js
```diff
@@ -75,6 +77,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+   *
+   * @param {string} method the name of the methods that (can be "symbol", "decimals", or "name")
```
josojo, line 84: Would it be wise to use here an enum?

(maybe not worth the effort as the whole project is not typed)

josojo (approved):
I love the very clean unit tests!

scripts/utils/hardcoded_tokens.json
fedgiac: Thanks for checking!

scripts/utils/trading_strategy_helpers.js
fedgiac: Never heard about enum, seems like it could be a good idea.

fedgiac: (After discussing privately, we decided to merge without enums.)

test/fetch_token_info.js
```diff
@@ -0,0 +1,26 @@
+const TestToken = artifacts.require("DetailedMintableToken")
+
+const { tokenDetail } = require("../scripts/utils/trading_strategy_helpers")(web3, artifacts)
+
+contract("FetchTokenInfo", () => {
+  describe("tokenDetail", () => {
+    it("relies on network for tokens outside list", async () => {
```
bh2smith, line 7: What is the meaning of "list" here?

scripts/utils/trading_strategy_helpers.js
```diff
@@ -75,6 +77,33 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+   * @param {SmartContract} tokenInstance instance of the token at the previous address
+   * @returns {string|number} the result of calling the selected method
+   */
+  const tokenDetail = async function (method, tokenInstance) {
+    let detail
+    const tokenAddress = tokenInstance.address
+    const networkId = tokenInstance.constructor.network_id
+    if (hardcodedTokensByNetwork[networkId] && hardcodedTokensByNetwork[networkId][tokenAddress]) {
+      return hardcodedTokensByNetwork[networkId][tokenAddress][method]
+    }
```
bh2smith, line 94: this project also contains a few other hard coded tokens by network:

https://github.com/gnosis/dex-liquidity-provision/blob/02e942f65b1a72671d2da455ab0fa2e9c655e9d1/scripts/synthetix/facilitate_trade.js#L23-L47

Not sure if it might be a good idea to have all these items in the same place. It is beginning to feel like it seems the synthetix scripts are entirely independent from the rest of this project and should perhaps be moved into their own repo.

PR #332, opened 2020-06-26
Replace dex.ag exchange with 1inch

### LINKED ISSUE #328: Integrate with 1inch
Actually 1inch is much better than dex.agg from a perspective of token coverage.
> OWL, DMG, COMP, UMA etc are all covered by 1inch.

We need this to support users putting in the right price and prevent confusions/errors, which we had during the DMM token sale with the front-end.

fedgiac:
Closes #328.

Basic implementation of the 1inch API, which is used to replace dex.ag to make the price ckeck available for more tokens. Note that the price we compute from 1inch is however less reliable from that of dex.ag because 1inch returns an exact selling/buying price for a specific amount, while we want a price that gives a general overview of the value of the token like dex.ag does.

### Test plan
See the price check failing with this command deploying brackets for the pair OWL/USDC at an unrealistic price of 100 and see that the suggested price is right (a bit more than 1 OWL per USDC). (Note the two tokens have different decimals.)
```
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=4 --quoteTokenId=0 --lowestLimit=210 --highestLimit=230 --currentPrice=100 --depositBaseToken=0.01 --depositQuoteToken=1 --numBrackets=20  --masterSafe=0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --network=rinkeby
```
```
<snip>
         chosen price: 100 OWL bought for 1 USDC
         dex.ag price: 1.0660178208889928 OWL bought for 1 USDC
```

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -5,6 +5,13 @@ const BN = require("bn.js")
 const { decodeOrders, getUnitPrice } = require("@gnosis.pm/dex-contracts")
 const { Fraction } = require("@gnosis.pm/dex-contracts")
 
+const { TEN } = require("./constants")
+const { toErc20Units } = require("./printing_tools")
+
+/**
+ * @typedef {import('../typedef.js').TokenObject} TokenObject
+ */
+
 const checkCorrectnessOfDeposits = async (
   currentPrice,
   bracketAddress,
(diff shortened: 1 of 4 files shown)


scripts/utils/price_utils.js
```diff
@@ -5,6 +5,13 @@ const BN = require("bn.js")
 const { decodeOrders, getUnitPrice } = require("@gnosis.pm/dex-contracts")
 const { Fraction } = require("@gnosis.pm/dex-contracts")
 
+const { TEN } = require("./constants")
+const { toErc20Units } = require("./printing_tools")
+
+/**
```
josojo, line 11: Is this comment needed?

scripts/utils/price_utils.js
```diff
@@ -118,19 +125,78 @@ const getDexagPrice = async function (tokenBought, tokenSold, globalPriceStorage
...
+    globalPriceStorage[baseToken + "-" + quoteToken] = price
+  }
+  return price
+}
+
 const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
-  const dexagPrice = await getDexagPrice(quoteTokenData.symbol, baseTokenData.symbol)
-  if (dexagPrice === undefined) {
-    console.log("Warning: could not perform price check against dex.ag.")
+  const onlinePrice = await getDexagPrice(quoteTokenData.symbol, baseTokenData.symbol)
```
josojo, line 188: Are we still using dexag?

scripts/utils/price_utils.js
```diff
@@ -118,19 +125,78 @@ const getDexagPrice = async function (tokenBought, tokenSold, globalPriceStorage
...
+    return globalPriceStorage[baseToken + "-" + quoteToken]
+  }
+  if (globalPriceStorage !== null && quoteToken + "-" + baseToken in globalPriceStorage) {
+    return 1.0 / globalPriceStorage[quoteToken + "-" + baseToken]
+  }
+  // use ETH instead of WETH for a more reliable price
+  baseToken = baseToken == "WETH" ? "ETH" : baseToken
+  quoteToken = quoteToken == "WETH" ? "ETH" : quoteToken
+  // 1inch returns only buy and sell prices for specific
+  const quoteTokenAmount = toErc20Units(TEN.pow(new BN(quoteToken.decimals)), quoteToken.decimals)
```
josojo, line 157: Should it be: toErc20Units(TEN, quoteToken.decimals)? I am confused right now.

Should we maybe use 1, instead of 10, just because selling 10BTC could already influence the price quite a bit?

scripts/utils/price_utils.js
```diff
@@ -118,19 +125,78 @@ const getDexagPrice = async function (tokenBought, tokenSold, globalPriceStorage
...
+ * one of the two.
+ *
+ * See https://1inch.exchange/#/api for API documentation.
+ *
+ * @param {TokenObject} baseToken base token for the price
+ * @param {TokenObject} quoteToken quote token for the price
+ * @param {object} [globalPriceStorage=null] object linking token pairs to prices
+ * @returns {number|undefined} desired price if available, undefined otherwise
+ */
+const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorage = null) {
```
josojo, line 145: This function looks good, but it is never called?

josojo (commented):
Thanks for doing it so quickly!

scripts/utils/price_utils.js
fedgiac: I checked, having this comment removes a linting warning. When checking, I noticed we have a lot of linting warning already, mostly because of the JSDoc checking (see [here](https://travis-ci.com/github/gnosis/dex-liquidity-provision/builds/173123954#L217)). Maybe @bh2smith knows what's going on with the warnings?

scripts/utils/price_utils.js
fedgiac: Whoops! That's absolutely a mistake, very good you spotted it!

scripts/utils/price_utils.js
fedgiac: We were, this is why all tests I did were working!

josojo (approved):
Looking good!


I would also delete the function `getDexagPrice`

josojo:
Dex.ag is still in the logs, and it appears in several places. Anyways, I am merging it right now, since there is this looming risk that users will not be protected with dex.ag from inserting the price incorrectly for many tokens.

scripts/utils/price_utils.js
bh2smith: Yea. I imposed forced JSDocs in the entire project. You might recall reviewing and approving it in #269

scripts/utils/price_utils.js
bh2smith: Probably should be deleted then.

scripts/utils/price_utils.js
bh2smith: This is absolutely needed in order to import the TokenObject type if it is being used here. All this JSdoc crap will go away when we move to TS.

PR #335, opened 2020-06-29
Add function to retrieve output of quickest promise

fedgiac:
Introduces the function `returnFirstNotErroring`. Given an array of promises as input, it resolves to the output of the input promise that resolves first (ignoring errors in other promises). If all promises fail, it rejects with an error.

This function is to be used when retrieving prices from aggregators like 1inch or dex.ag: Instead of choosing a single aggregator, we will be able to request prices from any aggregator and return the price of the fastest. Errors, which correspond to failing to retrieve prices from an exchange (for example because the token is not listed), can be safely ignored as long as at least one aggregator returns a price.

This function behaves similarly to the JS method `Promise.any`, which is however currently [not supported](https://kangax.github.io/compat-table/esnext/) by any version of Node.js. It is supposed to be replaced by `Promise.any` as soon as it becomes available.

The new tests should cover most use cases of this function.

### DIFF
--- a/scripts/utils/js_helpers.js
+++ b/scripts/utils/js_helpers.js
@@ -19,44 +19,6 @@ const sleep = function (milliseconds) {
 }
 
 /**
- * Helper function for returnFirstNotErroring. Returns a promise that
- * resolves to the value of the earliest input promise to resolve, or
- * rejects with a combined error if both promises reject.
- *
- * @param {Promise} combinedPromises promise combining other promises
- * with the same logic as in this function
- * @param {Promise} appendedPromise promise to add to combinedPromises
- * @returns {Promise} timeout
- */
-const appendPromise = (combinedPromises, appendedPromise) =>
-  new Promise((resolve, reject) => {
-    let combinedIsCaught = false
-    let combinedErrors
-    let appendedIsCaught = false
-    let appendedError
-    appendedPromise
-      .then((result) => resolve(result))
-      .catch((error) => {
-        appendedError = error.message || error
-        if (combinedIsCaught) {
-          reject(new Error(appendedError + " && " + combinedErrors))
-        } else {
-          appendedIsCaught = true
-        }
-      })
-    combinedPromises
-      .then((result) => resolve(result))
-      .catch((error) => {
-        combinedErrors = error.message || error
-        if (appendedIsCaught) {
-          reject(new Error(appendedError + " && " + combinedErrors))
-        } else {
-          combinedIsCaught = true
-        }
-      })
-  })
-
-/**
  * Given an array of promises, returns a promise resolving to the
  * output of the first resolving promise of the array. If any promise
  * has been rejected in the meantime, this failure is ignored. If all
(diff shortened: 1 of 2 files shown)


scripts/utils/js_helpers.js
```diff
@@ -18,7 +18,87 @@ const sleep = function (milliseconds) {
...
+ *    message is not available) in the order they appear in the input
+ *    array. (Promise.any returns an AggregateError.)
+ * 2. Even if all input promises are already resolved, the output is
+ *    a pending promise
+ *
+ * @param {Promise[]} promiseArray array of promises (or values)
+ * @returns {Promise} promise returning the value returned by the
+ * input promise that is the quickest to resolve
+ */
+const returnFirstNotErroring = function (promiseArray) {
```
nlordell, line 85: I think this can be implemented by just creating a single `new Promise(async (resolve, reject) => { ... })` and then adding a `.then` for each of the promises. You might also be able to use `await Promise.all` to create an `AggregateError` if all the promises fail.

fedgiac: Done! How is the result?
I believe also `AggregateError` is not supported right now, this is why I stitched together the error messages like this and returned a simple error.

scripts/utils/js_helpers.js
```diff
@@ -18,7 +18,57 @@ const sleep = function (milliseconds) {
...
+const returnFirstNotErroring = (promiseArray) =>
+  new Promise((resolve, reject) => {
+    let resolved = false
+    const errorPromises = []
+    for (const promise of promiseArray) {
+      errorPromises.push(
+        Promise.resolve(promise).then(
+          (result) => {
+            resolved = true
+            resolve(result)
```
nlordell, line 53: Does this need to be in a `if (!resolved) { resolve(result) }`? I'm not sure how promises behave when they are resolved multiple times.

scripts/utils/js_helpers.js
```diff
@@ -18,7 +18,57 @@ const sleep = function (milliseconds) {
...
+          (error) => error
+        )
+      )
+    }
+    Promise.all(errorPromises).then((errorArray) => {
+      if (!resolved) {
+        let errorString = ""
+        errorArray.forEach((error) => {
+          errorString += " && " + String(error.message || error)
+        })
```
nlordell, line 64: nit: I think using `map` and `join` is a little less verbose:
```suggestion
        let errorString = errorArray.map(error => String(error.message || error)).join(" && ")
```

nlordell (approved):
I much prefer it this way! I also find it easier to understand :smile:

scripts/utils/js_helpers.js
nlordell: TIL, you can `resolve` and `reject` as much as you want, only the first call matters. Maybe the flag isn't needed at all?

bh2smith (approved):
This seems pretty cool. But are we sure we want the first price returned? What if the fastest price is the worst? Anyway, looks really good. Too bad this logic couldn't live somewhere more appropriate (i.e. outside of this project).

fedgiac:
The user gives the price explicitly anyway, the retrieved price is only used to double check that the price of the user is not unrealistic. As long as the price is in the same ballpark as the correct one, we should be fine. Also, 1inch price can take more than ten seconds to be returned, which is more time than the actual execution of the script without this check.

PR #339, opened 2020-06-30
[README] Move Token Distribution Instructions to main README

bh2smith:
The Token Distributor guide was sitting right in the middle of the CMM scripts. This moves it out at least temporarily before it is migrated into its own repo.


josojo (approved):
thanks for the clean up

fedgiac (approved):
Looks good!

I just noticed that there is no information about the actual liquidity scripts on the main readme, maybe we should add that in another PR.

PR #340, opened 2020-06-30
View script

fedgiac:
Related to issue #31.

Adds a basic script returning the balance of all brackets (either all brackets owned by a master Safe or all input brackets). Zero balances are ignored.

Known limitations:
1. script does not return the bracket balances of tokens for which the bracket has no open orders.
2. the result is incorrect in brackets that have an order or a withdraw request being settled in the same batch.
3. if used with many brackets, the script fails because of Infura rate limiting.

Test that the script works:
```
npx truffle exec scripts/viewer.js --masterSafe=0x5CC3a4C846aB8b0B78a729b8B1a50e2E0BB66740 --network=mainnet
```

Note that DAI were deployed incorrectly today:
```
npx truffle exec scripts/viewer.js --brackets=0x85C6e1d780c1d55a3d7d363fc162EDc48c540475 --network=mainnet
```

I was able to run the script with ~160 brackets, running time was 1m14s.
Trying to run it with about 300 brackets failed because of Infura rate limits.


scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+  return fullOptionalString
+}
+
+module.exports = async (callback) => {
+  try {
+    const exchangePromise = getExchange(web3)
+    let bracketAddresses
+    if (argv.masterSafe) {
+      bracketAddresses = await getDeployedBrackets(argv.masterSafe)
```
bh2smith, line 49:
```suggestion
      bracketAddresses = await getDeployedBrackets(argv.masterSafe)
      console.log(`Found ${bracketAddresses.length} brackets`)
```

scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+        tradedTokenIds = uniqueItems(tradedTokenIds)
+        const tokenInfoPromises = fetchTokenInfoFromExchange(exchange, tradedTokenIds)
+        return {
+          bracketAddress,
+          tokenIds: tradedTokenIds,
+          tokenInfoPromises,
+        }
+      })
+    )
+
```
bh2smith, line 74:
```suggestion
    console.log("Recovering token balances per bracket...")
```

scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+      console.log(
+        `Balance of ${balanceItem.bracketAddress} for token ${tokenData.symbol}: ${fromErc20Units(
+          balanceItem.totalBalance,
+          tokenData.decimals
+        )}${optionalString}`
+      )
+    }
+
+    console.log()
+    console.log("Total funds:")
```
bh2smith, line 129:
```suggestion
    console.log("/nTotal funds:")
```

scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+      const optionalString = buildOptionalString({ tokenData, ...balanceItem })
+      console.log(
+        `Balance of ${balanceItem.bracketAddress} for token ${tokenData.symbol}: ${fromErc20Units(
+          balanceItem.totalBalance,
+          tokenData.decimals
+        )}${optionalString}`
+      )
+    }
+
+    console.log("/nTotal funds:")
```
Voronchuk, line 129: It doesn't render as intended in my Mac OS terminal, see screen:
Screenshot 2020-07-01 at 12 17 27

fedgiac: Should be fixed now, let me know if this is still an issue for you.

bh2smith:
Just a general comment for now. I think the filename should be `balance_viewer` since there are other things that can be "viewed".

bh2smith:
Just a general comment for now. I think the filename should be `balance_viewer` since there are other things that can be "viewed". Also, I think we should include a point on how to use it in the README.

fedgiac:
I intentionally didn't add this script to the readme because @josojo mentioned there is already some better way to visualize the balances with Dune. I don't know how it works however. Also, I don't know if it shows all the extra information printed by this script (claimable amount and amount directly stored in bracket), which has already come quite handy.

bh2smith:
> I intentionally didn't add this script to the readme because @josojo mentioned there is already some better way to visualize the balances with Dune.

It is fine without the readme (I suppose), but we already have user(s) who are interested in acquiring this information directly from this repo using this script. I really don't think it would hurt to include something here (since it is here now anyway).

scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+    let bracketAddresses
+    if (argv.masterSafe) {
+      bracketAddresses = await getDeployedBrackets(argv.masterSafe)
+      console.log(`Found ${bracketAddresses.length} brackets`)
+    } else {
+      if (!argv.brackets) {
+        callback("One of --masterSafe or --brackets=... is required.")
+      }
+      bracketAddresses = argv.brackets
+    }
```
bh2smith, line 56: So if I provide both master safe and brackets arguments the brackets are ignored and `masterSafe` overrides it.

scripts/viewer.js
```diff
@@ -0,0 +1,150 @@
...
+      if (balanceItem.totalBalance.isZero()) {
+        continue
+      }
+      totalBalanceSum[balanceItem.tokenId] = (totalBalanceSum[balanceItem.tokenId] || new BN(0)).add(balanceItem.totalBalance)
+      requestedForWithdrawSum[balanceItem.tokenId] = (requestedForWithdrawSum[balanceItem.tokenId] || new BN(0)).add(
+        balanceItem.requestedForWithdraw
+      )
+      storedInBracketSum[balanceItem.tokenId] = (storedInBracketSum[balanceItem.tokenId] || new BN(0)).add(
+        balanceItem.storedInBracket
+      )
```
bh2smith, line 118: Maybe just a nit, but given that JS doesn't have default dict like in python: you may want to consider using the following more common technique for making your accumulators:

https://github.com/gnosis/dex-liquidity-provision/blob/154b7c5a838c049d0d7a316f1b19decc642e9081/scripts/utils/trading_strategy_helpers.js#L449-L457

See definition and implementation of `cummulativeAmounts` as a Map initialized to zeroes. Of course this requires knowing all the unique keys before hand.

bh2smith (approved):
Looks pretty good. I find the string formatting function under documented and confusing without having any comments. Makes the top of this script unapproachable in terms of reading the code.

scripts/README.md
```diff
@@ -66,6 +66,22 @@ The fleet size should be smaller than or equal to 20, in order to ensure that th
 Please document the displayed bracket-trader addresses. They are required for future withdrawals.
 They can also be retrieved from the created transactions. However, since this is a manual process, it is quite cumbersome to extract them right now.
 
+### Monitoring deployed funds
```
bh2smith, line 69:
```suggestion
### Monitoring Deployed Funds
```

scripts/viewer.js
fedgiac: I tried to make this work but there are two issues with this approach:
1. you can't specify a default for `Map.get`
2. because of 1, you need to know all keys of the map in advance to default them to zero, leading to quite some extra code.

The advantage of `Map` is that you don't risk to overwrite some preexisting prototype keys, but in this case all keys are integers, so there is no risk of a conflict.

So, overall I'd keep the current implementation to avoid the extra code.

scripts/README.md
fedgiac: This could be turned into an issue, since there is no consistency in the title capitalization of the readme.

scripts/README.md
```diff
@@ -51,7 +51,7 @@ Making the requests to the gnosis-interfaces does not cost any gas. However, sig
 
 Here is an example script invocation:
 
-```js
```
bh2smith, line 54: does this not highlight syntax propoerly or what is the reason for removing it?

bh2smith: Ideally all of these commands should be tagged with `sh` to highlight as shell command.

fedgiac: Well, it's not js code, so I removed that. I wanted to replace is with something representing the terminal, but then I'd have had to choose between `sh`, `bash`, and `zsh`.
Let's see:

fedgiac: JS:
```js
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --depositBaseToken=0.1 --depositQuoteToken=10 --numBrackets=10 --network=$NETWORK_NAME
```
Bash:
```bash
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --depositBaseToken=0.1 --depositQuoteToken=10 --numBrackets=10 --network=$NETWORK_NAME
```
Zsh:
```zsh
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --depositBaseToken=0.1 --depositQuoteToken=10 --numBrackets=10 --network=$NETWORK_NAME
```
Sh:
```sh
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=4 --lowestLimit=150 --highestLimit=200 --currentPrice=175 --masterSafe=$MASTER_SAFE --depositBaseToken=0.1 --depositQuoteToken=10 --numBrackets=10 --network=$NETWORK_NAME
```

PR #347, opened 2020-07-01
Error on missing Infura key

### LINKED ISSUE #326: Missing Infura key in documentation
Since the default Infura key in this project has been made invalid, users must specify their own Infura key. This extra step should be mentioned in the documentation. Error:
```
Error: There was a timeout while attempting to connect to the network.
       Check to see that your provider is valid.
       If you have a slow internet connection, try configuring a longer timeout in your Truffle config. Use the networks[networkName].networkCheckTimeout property to do this.
    at Timeout.setTimeout (/path/dex-liquidity-provision/node_modules/truffle/build/webpack:/packages/provider/index.js:56:1)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
```

fedgiac:
Fixes #326.

Makes all scripts fail if no Infura key is specified.

Note that the current default Infura key is invalid, and an user trying the scripts without a key would see only a network error.

## Test plan
```
unset INFURA_KEY # and comment out INFURA_KEY from .env if present 
npx truffle exec scripts/claim_withdraw.js --masterSafe=0x0
```
Expected output:
```
$ npx truffle exec scripts/claim_withdraw.js --masterSafe=0x0
> Warning: possible unsupported (undocumented in help) command line option: --masterSafe
Error: no Infura key found! You can create a new project key after registering an account at https://infura.io/
```

Note that the extra yargs doesn't break the help:
```
npx truffle exec scripts/claim_withdraw.js --help --network=mainnet
```
(It also works without `--network=mainnet`, but there needs to be Ganache running.)

### DIFF
--- a/truffle-config.js
+++ b/truffle-config.js
@@ -1,5 +1,9 @@
 const truffleConfig = require("@gnosis.pm/util-contracts/src/util/truffleConfig")
 
+const argv = require("yargs").help(false).version(false).option("network", {
+  type: "string",
+}).argv
+
 const DEFAULT_GAS_LIMIT = 6e6
 const DEFAULT_GAS_PRICE_GWEI = 5
 const DEFAULT_MNEMONIC = "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat"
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Nice I really like this. We should be aware, however, that this truffle-config is essentially being used across all projects within gnosis so we are no longer using the common version.

Issue #348, opened 2020-07-01
TransferApproveDeposit - Partially Executed


bh2smith:
This transaction has filed to complete: Successfully only sending funds into the bracket and possibly also approving the exchange for the deposit amount.

https://etherscan.io/tx/0xa538f4a0afa57118c14100ac538b0dbb2c094efddb4393aa48ea3e9cb637aa6b/advanced#internal

It seems the transaction ran out of gas at this point and now the funds are in the bracket.

fedgiac:
Same issue described in #350.

PR #349, opened 2020-07-01
Deposit from bracket

bh2smith:
We have discovered a partially filled transaction #348 and this script is used to move the funds from a sub safe into the exchange.

Test Plan:

From the existing masterSafe, first transfer some funds into a bracket that is owned by this safe. Then create a `DEPOSIT_FILE` with the following JSON format

```
[
    {
        "amount": "100",
        "tokenAddress": "0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea",
        "bracketAddress": "0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636"
    },
    {
        "amount": "100",
        "tokenAddress": "0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea",
        "bracketAddress": "0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636"
    }
]
```

```sh
npx truffle exec scripts/deposit.js --masterSafe=0xd9395aeE9141a3Efeb6d16057c8f67fBE296734c --depositFile=$DEPOSIT_FILE --network=rinkeby
```

here is a link to a successful transaction resulting from this test plan

https://rinkeby.etherscan.io/tx/0x9ad237eaa61cddcead609026a975d16d5346c917aa340bec7080dc2df9cedbc1

### DIFF
--- a/test/scripts/deposit.js
+++ b/test/scripts/deposit.js
@@ -31,7 +31,7 @@ contract("Deposit scripts", function (accounts) {
       const decimals = tokenData.decimals || 18
       // Adding token to exchange is not actually necessary since one can deposit and withdraw any token.
       const { id, token } = await addCustomMintableTokenToExchange(exchange, symbol, decimals, accounts[0])
-      tokenInfo.push({ symbol: symbol, decimals: decimals, id: id, token: token, address: token.address })
+      tokenInfo.push({ symbol, decimals, id, token, address: token.address })
       for (const bracket of bracketAddresses) {
         await token.mint(bracket, amount)
         deposits.push({
--- a/test/scripts/deposit.js
+++ b/test/scripts/deposit.js
@@ -67,7 +67,7 @@ contract("Deposit scripts", function (accounts) {
         const bracketBalance = (await token.balanceOf(bracketAddress)).toString()
         const exchangeBalance = (await exchange.getBalance(bracketAddress, tokenAddress)).toString()
         assert.equal(bracketBalance, "0", "Bracket balance should be zero")
-        assert.equal(exchangeBalance, amount, "Exchange balance should should contain amount")
+        assert.equal(exchangeBalance, amount, "Exchange balance should contain amount")
       }
     })
   })
--- a/test/scripts/deposit.js
+++ b/test/scripts/deposit.js
@@ -1,0 +1,74 @@
+const Contract = require("@truffle/contract")
+
+const BatchExchange = Contract(require("@gnosis.pm/dex-contracts/build/contracts/BatchExchange"))
+const GnosisSafe = artifacts.require("GnosisSafe")
+const ProxyFactory = artifacts.require("GnosisSafeProxyFactory")
+
+const { addCustomMintableTokenToExchange, deploySafe } = require("../test_utils")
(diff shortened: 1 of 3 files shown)


test/scripts/deposit.js
```diff
@@ -0,0 +1,74 @@
...
+    const masterSafe = await GnosisSafe.at(await deploySafe(gnosisSafeMasterCopy, proxyFactory, [safeOwner], 1))
+    const bracketAddresses = await deployFleetOfSafes(masterSafe.address, numberOfBrackets)
+    const tokenInfo = []
+    const deposits = []
+    for (const { tokenData = {}, amount } of amounts) {
+      const symbol = tokenData.symbol || "TEST"
+      const decimals = tokenData.decimals || 18
+      // Adding token to exchange is not actually necessary since one can deposit and withdraw any token.
+      const { id, token } = await addCustomMintableTokenToExchange(exchange, symbol, decimals, accounts[0])
+      tokenInfo.push({ symbol: symbol, decimals: decimals, id: id, token: token, address: token.address })
```
fedgiac, line 34: This should also work:
```suggestion
      tokenInfo.push({ symbol, decimals, id, token, address: token.address })
```
But feel free to leave it as it was.

test/scripts/deposit.js
```diff
@@ -0,0 +1,74 @@
...
+  })
+
+  const setup = async function (numberOfBrackets, amounts) {
+    const masterSafe = await GnosisSafe.at(await deploySafe(gnosisSafeMasterCopy, proxyFactory, [safeOwner], 1))
+    const bracketAddresses = await deployFleetOfSafes(masterSafe.address, numberOfBrackets)
+    const tokenInfo = []
+    const deposits = []
+    for (const { tokenData = {}, amount } of amounts) {
+      const symbol = tokenData.symbol || "TEST"
+      const decimals = tokenData.decimals || 18
```
fedgiac, line 31: We only accept a deposit file, so I'd say that at least the token decimals should not be optional and the test should fail if no decimals are provided.

test/scripts/deposit.js
```diff
@@ -0,0 +1,74 @@
...
+        assert.equal(exchangeBalance, "0", "Exchange balance should should be zero")
+        assert.equal(pendingExchangeBalance, amount, "Exchange pending balance should contain amount")
+      }
+      await waitForNSeconds(301)
+      // Balance assertions after batch is closed
+      for (const { amount, tokenAddress, bracketAddress } of deposits) {
+        const bracketBalance = (await token.balanceOf(bracketAddress)).toString()
+        const exchangeBalance = (await exchange.getBalance(bracketAddress, tokenAddress)).toString()
+        assert.equal(bracketBalance, "0", "Bracket balance should be zero")
+        assert.equal(exchangeBalance, amount, "Exchange balance should should contain amount")
```
fedgiac, line 70:
```suggestion
        assert.equal(exchangeBalance, amount, "Exchange balance should contain amount")
```

fedgiac (approved):
Looks good.

test/scripts/deposit.js
bh2smith: I like it.

test/scripts/deposit.js
bh2smith: Technically, we only use WEI here anyWEI so decimals don't matter at all. I would like to leave it for now, because I plan to use decimals in a future change to deposits.

PR #360, opened 2020-07-08
Fix USDC decimals in verification script

fedgiac:
Fixes bug introduced in #332.
This bug involves the verification script and had two effects:
- All orders were considered "negligible" because of their low USD value, so all profitability check failure would have been ignored.
- Under some uncommon circumstances, it would make the profitability check to fail incorrectly.

Not sure how to properly test for this issue, since it's caused by the interaction between 1inch and our standard token format.


josojo (approved):
Thanks for finding and fixing it!

bh2smith (commented):
Why is this hard coded in the first place? Seems like a recipe for disaster.

fedgiac:
Very good point, I opened an issue referencing this PR.

PR #361, opened 2020-07-08
Do not retrieve prices online if not needed in the test

fedgiac:
Some tests fetch prices online even when not needed. This PR reduces the amount of network requests in tests. The running time of the tests goes down by about 30 seconds each.

### Test plan
Compare running times of the [current master Travis build](https://travis-ci.com/github/gnosis/dex-liquidity-provision/builds/174594521) with the Travis build of this PR for the tests:
1. throws if orders do not buy and sell the same tokens in a loop
2. throws if orders of one bracket are not profitable

### DIFF
--- a/test/test_utils.js
+++ b/test/test_utils.js
@@ -53,9 +53,36 @@ const createTokenAndGetData = async function (symbol, decimals) {
   return { address: token.address, tokenData: tokenData }
 }
 
+/**
+ * Creates a price storage with dummy prices for the input tokens.
+ * Used to make less network requests during testing.
+ *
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
This looks fine for now, but I would suggest at some point that we build a proper "priceSource" class that is "mockable" for tests but can be used to "normalize" multiple different price sources. This will be handy at some point if and when we use multiple different price sources like uniswap, chainlink, 1inch, dex-ag etc... so that our project can expect to receive the same formatted information all the time.

PR #370, opened 2020-07-17
Remove hardcoded dex.ag from log output

fedgiac:
The log output of our script explicitly references to dex.ag as our price source. This PR changes this.

The output of the external price feed now includes information on the exchange from which the price was retrieved. This is what is then used for the log output.

I also removed some leftover code in the 1inch API that comes from the implementation of the dex.ag API. This code was supposed to change "WETH" to "ETH", but
1. this is not needed for 1inch, and
2. as currently implemented, it has no effect.


josojo (approved):
Looks really good. Thanks

PR #377, opened 2020-07-24
Airdrop from CSV

bh2smith:
This PR is intended to generalize the airdrop process by allowing both JSON and CSV file extensions. However, we notice that the current CSV->JSON parsing tool is not very general at the moment because it reads the column title `gno_estimation` as amount and uses the hard-coded `GNO_ADDRESS` for the `tokenAddress` field. Furthermore, the existing leader boards for the incentive program are not consistent in the naming of `receiver` (i.e we see both of `trader` and `owner` appearing in column headers). 

For example, here we see "trader" as receiver
https://explore.duneanalytics.com/dashboard/gnosis-protocol-challenge-v3---17th-july-14th-aug

and here we see "owner"
https://explore.duneanalytics.com/queries/5106#10063


scripts/airdrop.js
```diff
@@ -1,4 +1,6 @@
 const fs = require("fs").promises
+const csv = require("csv-parser")
```
nlordell, line 2: Does this require a new package in `package.json` and `yarn.lock`?

scripts/airdrop.js
```diff
@@ -1,4 +1,6 @@
...
+const BigNumber = require("bignumber.js")
```
nlordell, line 3: I think this package is also missing an explicit dependency in `package.json`. The reason why this may be important is that if you don't add an explicit dependency, and one of your other dependencies stop using this package internally, then this will stop working out of nowhere.

scripts/airdrop.js
```diff
@@ -21,12 +23,56 @@ const argv = default_yargs
     describe: "Do not actually send transactions, just simulate their submission",
   }).argv
 
+const getFileExtension = function (filename) {
```
nlordell, line 26: This method is part of the NodeJS standard packages: https://nodejs.org/dist/latest-v12.x/docs/api/path.html#path_path_extname_path

scripts/airdrop.js
```diff
@@ -21,12 +23,56 @@ const argv = default_yargs
...
+}
+
+const parseCsvFile = async function (filePath) {
+  return new Promise((resolve, reject) => {
+    const results = []
+    fs.createReadStream(filePath, { encoding: "utf-8" })
+      .pipe(csv())
+      .on("data", (data) => results.push(data))
+      .on("end", () => resolve(results))
+      .on("error", (error) => reject(error))
```
nlordell, line 38: So PRO.

scripts/airdrop.js
```diff
@@ -21,12 +23,56 @@ const argv = default_yargs
...
+  })
+}
+// Note that this function is very specific to the incentive program itself
```
nlordell, line 41: uber-nit:
```suggestion
}

// Note that this function is very specific to the incentive program itself
```

scripts/airdrop.js
```diff
@@ -21,12 +23,56 @@ const argv = default_yargs
...
+
+    const payments = results.map(toPayment).filter((payment) => !payment.amount.isZero())
+    return payments.map(({ amount, receiver, tokenAddress }) => ({
+      amount: amount.toString(10),
+      receiver,
+      tokenAddress,
+    }))
+  } else if (ext == "json") {
+    return JSON.parse(await fs.readFile(filename, "utf8"))
+  }
```
nlordell, line 67: nit: I think you should handle the extra case where its neither kind of file
```suggestion
  } else {
    throw new Error(`unsupported file type .${ext}`)
  }
```

scripts/airdrop.js
```diff
@@ -21,12 +23,56 @@ const argv = default_yargs
...
+  if (ext === "csv") {
+    const results = await parseCsvFile(filename)
+
+    const payments = results.map(toPayment).filter((payment) => !payment.amount.isZero())
+    return payments.map(({ amount, receiver, tokenAddress }) => ({
+      amount: amount.toString(10),
+      receiver,
+      tokenAddress,
+    }))
+  } else if (ext == "json") {
```
nlordell, line 65: uber-nit:
```suggestion
  } else if (ext === "json") {
```

nlordell (approved):
Looks good, just a few nits.

scripts/airdrop.js
bh2smith: It does. I am still experimenting with this and probably should have marked it as a draft.

scripts/airdrop.js
bh2smith: I can't tell the difference here.

nlordell: The extra newline 😄

scripts/airdrop.js
fedgiac: I remember we had a bug in this repo for this very reason: one of our file was importing `bn.js` but we didn't explicitly mark it as a dependency in `package.json`. The result was that because of that we were using an old version of `bn.js` imported by another package which had a subtle but critical bug (bad input parsing on `BN` creation) that was quite hard to debug.

scripts/airdrop.js
```diff
@@ -63,7 +57,7 @@ const parseTransferFile = async function (filename) {
       receiver,
       tokenAddress,
     }))
-  } else if (ext === "json") {
+  } else if (ext === ".json") {
     return JSON.parse(await fs.readFile(filename, "utf8"))
   } else {
     throw new Error(`unsupported file type .${ext}`)
```
nlordell, line 63: Just a little issue with my previous suggestion since `path.extname` already includes the `.`

```suggestion
    throw new Error(`unsupported file type ${ext}`)
```

scripts/airdrop.js
```diff
@@ -21,13 +24,55 @@ const argv = default_yargs
...
+  }
+}
+
+const parseTransferFile = async function (filename) {
+  const ext = path.extname(filename).toLowerCase()
+
+  if (ext === ".csv") {
+    const results = await parseCsvFile(filename)
+
+    const payments = results.map(toPayment).filter((payment) => !payment.amount.isZero())
```
josojo, line 54: Should we do this filtering outside of the parsing, so that it will be applied to both formats?

scripts/airdrop.js
```diff
@@ -21,13 +24,55 @@ const argv = default_yargs
...
+    return JSON.parse(await fs.readFile(filename, "utf8"))
+  } else {
+    throw new Error(`unsupported file type ${ext}`)
+  }
+}
+
 module.exports = async (callback) => {
   try {
+    const transfers = await parseTransferFile(argv.transferFile)
+    console.log(transfers)
```
josojo, line 70: Is this console.log intended?

josojo (approved):
Cool! Thanks. I think we are ready for Monday!

scripts/airdrop.js
```diff
@@ -21,13 +24,55 @@ const argv = default_yargs
     describe: "Do not actually send transactions, just simulate their submission",
   }).argv
 
+const parseCsvFile = async function (filePath) {
```
anxolin, line 27: I think this belongs more in a utils file, right?
Is not related to airdrop and can be used with any CSV

anxolin (approved):

scripts/airdrop.js
bh2smith: No! Whoops!

scripts/airdrop.js
bh2smith: Yes, I suppose so.

scripts/airdrop.js
bh2smith: Yes! I will fix this.

PR #381, opened 2020-07-27
[Readme] Include Airdrop transfer file standard

bh2smith:
Just including file format for csv and json and rounding up


josojo (approved):
Thanks!

Issue #383, opened 2020-07-28
OrderSellValueUSD has reversed base and quote tokens


bh2smith:
The following function returns inverted price values because the first two parameters passed into the function are reversed. 

https://github.com/gnosis/dex-liquidity-provision/blob/0abecafb87036219bea36388d4f9a26e8dbbc2b2/scripts/utils/price_utils.js#L247-L258

See the doc string definition of `getOneInchPrice`

https://github.com/gnosis/dex-liquidity-provision/blob/0abecafb87036219bea36388d4f9a26e8dbbc2b2/scripts/utils/price_utils.js#L140-L145

I suspect base and quote as terms are not well understood. We would benefit from having a unit test for expectations regarding the return values of this function.

This failed unit test demonstrates the bug quite clearly:

```js
  describe("orderSellValueInUSD()", async () => {
    it("Ensures function returns expected values", async () => {
      const ETHtokenId = (await addCustomMintableTokenToExchange(exchange, "ETH", 18, accounts[0])).id
      const tokenInfo = fetchTokenInfoFromExchange(exchange, [ETHtokenId])
      const globalPriceStorage = {}
      globalPriceStorage["ETH-USDC"] = { price: 250.0 }

      const order = {
        sellTokenBalance: "1000000000000000000",
        sellToken: ETHtokenId,
      }
      const value = await orderSellValueInUSD(order, tokenInfo, globalPriceStorage)

      assert.equal(value.toString(10), "250")
    })
  })
```

PR #386, opened 2020-07-28
[Claim Withdraw] Exclude Dust amounts 

### LINKED ISSUE #345: [Withdraw] Improve logging and exclude dust amounts.
There is some missing information in withdraw logging that would be usefully to know (i.e. number of brackets). We may also wish to exclude those with insignificant balance from the transaction, since it costs more in gas than the balances are worth. That is, exclude dust orders.

Furthermore, we this script should also display a cumulative sum of balances.

bh2smith:
This script update excludes withdraw claims valuing less than 1 USD (i.e. those costing more in transaction fees than the value itself).

There a couple new unit tests showing that the script works as desired.

Closes #345

### DIFF
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -70,16 +70,21 @@ module.exports = function (web3, artifacts) {
         tokenBracketPairs.map(([bracketAddress, tokenData]) => amountFunction(bracketAddress, tokenData, exchange))
       )
       withdrawals = []
-      maxWithdrawableAmounts.forEach(async (amount, index) => {
-        const tokenData = tokenBracketPairs[index][1]
-        const usdValue = await amountUSDValue(amount, tokenData, globalPriceStorage)
-        if (usdValue.gte(ONE))
+      for (const [index, amount] of maxWithdrawableAmounts.entries()) {
+        const token = tokenBracketPairs[index][1]
+        const bracketAddress = tokenBracketPairs[index][0]
+        const usdValue = await amountUSDValue(amount, token, globalPriceStorage)
+        if (usdValue.gte(ONE)) {
           withdrawals.push({
-            bracketAddress: tokenBracketPairs[index][0],
-            tokenAddress: tokenData.address,
-            amount: amount,
+            bracketAddress,
+            tokenAddress: token.address,
+            amount,
+            usdValue,
           })
-      })
+        } else {
+          log(`Skipping request for ${token.symbol} on bracket ${bracketAddress} since USD value < 1`)
+        }
+      }
     }
 
     return {
@@ -116,7 +121,7 @@ module.exports = function (web3, artifacts) {
 
     return transactionPromise
   }
-  const prepareWithdraw = async function (argv, printOutput = false) {
+  const prepareWithdraw = async function (argv, printOutput = false, globalPriceStorage = {}) {
     const log = printOutput ? (...a) => console.log(...a) : () => {}
 
     assertGoodArguments(argv)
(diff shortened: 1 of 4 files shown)


scripts/wrapper/withdraw.js
```diff
@@ -64,18 +65,26 @@ module.exports = function (web3, artifacts) {
...
+        const usdValue = await amountUSDValue(amount, token, globalPriceStorage)
+        if (usdValue.gte(ONE)) {
           withdrawals.push({
-            bracketAddress: tokenBracketPairs[index][0],
-            tokenAddress: tokenBracketPairs[index][1].address,
-            amount: amount,
+            bracketAddress,
+            tokenAddress: token.address,
+            amount,
+            usdValue,
```
josojo, line 82: Why do we store this value? Will it ever be used?

test/scripts/withdraw.js
```diff
@@ -326,11 +333,61 @@ contract("Withdraw script", function (accounts) {
...
+      const argv1 = {
+        masterSafe: masterSafe.address,
+        brackets: bracketAddresses,
+        tokens: [tokenInfo[0].address, tokenInfo[1].address],
+      }
+      const transaction1 = await prepareRequestWithdraw(argv1)
+      await execTransaction(masterSafe, safeOwner, transaction1)
+      await waitForNSeconds(301)
+
+      const argv2 = {
```
josojo, line 362: This one is not needed, as it is the same as argv2, right?

test/scripts/withdraw.js
```diff
@@ -326,11 +333,61 @@ contract("Withdraw script", function (accounts) {
...
+        brackets: bracketAddresses,
+        tokens: [tokenInfo[0].address, tokenInfo[1].address],
+      }
+      const transaction2 = await prepareWithdraw(argv2, false, globalPriceStorage)
+      await execTransaction(masterSafe, safeOwner, transaction2)
+
+      const usdcAddress = tokenInfo[0].address
+      for (const { amount, tokenAddress, bracketAddress } of deposits) {
+        const bracketBalance = (await (await ERC20.at(tokenAddress)).balanceOf(bracketAddress)).toString()
+        const pending = (await exchange.getPendingWithdraw(bracketAddress, tokenAddress))[0].toString()
```
josojo, line 373: pending could be renamed to pendingWithdraw for more readablity

test/scripts/withdraw.js
```diff
@@ -326,11 +333,61 @@ contract("Withdraw script", function (accounts) {
...
+        if (tokenAddress === usdcAddress) {
+          assert.equal(bracketBalance, "0", "Not expecting to claim USDC balance")
```
josojo, line 375: Should we add another assert here? assert: pending > 0? 

in order to confirm that it was not withdrawn?

josojo (approved):
Thanks a lot Ben,

just minors...

scripts/wrapper/withdraw.js
bh2smith: This was something I meant to delete although I thought it might be used in this way at some point.

test/scripts/withdraw.js
bh2smith: Yes, there are many places in this test file that makes redundant use of argv2. Thanks for pointing it out. I will change all of them

josojo:
@bh2smith 

There is one error in the PR and Ben, it's not your fault at all!
```
await getOneinchPrice(baseToken={ symbol: "WETH", decimals: 18 }, quoteToken= { symbol: "USDC", decimals: 6 }, globalPriceStorage)
```
returns 1/300 instead of 300 and hence all the calculation is incorrect. 

I think we need to change the function `getOneinchPrice` to return the right price and unfortunately, this has also implications on all other places, where the price function is used.

Can you double-check whether my statement is true?

bh2smith:
@josojo - I believe you are referring to the bug report I filed yesterday #383 (I believe I already shared this with you). The calculation in this PR is correct (as seen by the unit tests). I was going to go and fix this bug next.

PR #387, opened 2020-07-28
[Synthetix] Increase Min Trade Size

bh2smith:
This PR is a HOTFIX. Since the bot is now attempting to fill an order worth about 35 USD and it is not being matched. It appears solvers are not picking up orders with the current 10 USD min trade size. It would be much more likely that they would include a min trade of 100. Should we make it more?

The new docker image related to this change has already been pushed.

### DIFF
--- a/scripts/synthetix/facilitate_trade.js
+++ b/scripts/synthetix/facilitate_trade.js
@@ -52,7 +52,7 @@ const estimatePrice = async function (buyTokenId, sellTokenId, sellAmount, netwo
   return estimationData.buyAmountInBase / estimationData.sellAmountInQuote
 }
 
-const MIN_SELL_USD = 10
+const MIN_SELL_USD = 100
 
 /* All prices, except those explicitly named with "inverted" refer to the price of sETH in sUSD.
  * The term "our" is with regards to prices we, the synthetix bot, are buying and selling for while


nlordell (approved):
I think long-term the solver will use the current gas estimate to decide whether a trade is profitable. I think at 40GWei the trade needs to be around 1150$ for it to be profitable for the solver. Currently the plan is to subsidize this, but maybe the amount should be a little more (like 200-500)?

josojo:
I think we should only trade "with ourselves" if the volume is bigger than our minimal economical viable order size without subsidies. That means we should only trade orders around the size of ~1500 USD, if we don't want to make it dependent on the current gas situation.

bh2smith:
All these suggestions are great, but it is still a fact that this bot only has about 800$ worth of funding of each token. This is merely a hot fix to ensure the bot is not submitting a bunch of unnecessary transactions and essentially burning gas for no reason.

There is a separate task and branch (`snx_economic_viability`) related to the pure economic viability aspect of this bot's order submission. I can make the min trade size a bit more here, but I wouldn't get carried away (especially not with more funds than it even has). Would 250 seem more appropriate?

nlordell:
> Would 250 seem more appropriate?

I would go higher. If you have 800 USD of funds to trade, maybe 400 USD (so half of what you can trade?)

bh2smith:
We talked about this a bit more in the office today and determined that the solver gives a 10x subsidy factor so that we could actually be using a value of 150. Shall I change it to this and push a new image?

josojo:
> We talked about this a bit more in the office today and determined that the solver gives a 10x subsidy factor so that we could actually be using a value of 150. Shall I change it to this and push a new image?

I would not be in favor of 150, as we might decrease the subsidy soon and it is not worth if we are generating this little trading volume for the high gas costs.
But it's just a personal opinion. Maybe @fleupold has a different preference.

bh2smith:
Ok, given that this is a hot fix, everything is running as intended, there is no consensus on a value and there is a draft PR with the next, more important, improvement I am going to merge this and it can be forgotten.

PR #389, opened 2020-07-29
fix and test orderSellValueInUSD

### LINKED ISSUE #383: OrderSellValueUSD has reversed base and quote tokens

bh2smith:
Closes #383 

We simply swap the bast and quote token as indicated in the issue. Also, introducing a unit test to make sure these kind things don't happen again.

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -132,7 +132,7 @@ const getDexagPrice = async function (tokenBought, tokenSold, globalPriceStorage
  *
  * Note that the output price is not very accurate, especially for tokens with
  * very high value and very little liquidity. This is due to the fact that 1inch
- * returns a sell and a by price for a given amount
+ * returns a sell and a buy price for a given amount
  *
  * See https://1inch.exchange/#/api for API documentation.
  *
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -246,8 +246,8 @@ const checkNoProfitableOffer = async (order, exchange, tokenInfo, globalPriceSto
 
 const orderSellValueInUSD = async (order, tokenInfo, globalPriceStorage = null) => {
   const currentMarketPriceSlice = await getOneinchPrice(
-    { symbol: "USDC", decimals: 6 },
     await tokenInfo[order.sellToken],
+    { symbol: "USDC", decimals: 6 },
     globalPriceStorage
   )
   const currentMarketPrice = currentMarketPriceSlice.price
@@ -268,6 +268,7 @@ const amountUSDValue = async function (amount, tokenInfo, globalPriceStorage = n
 
 module.exports = {
   amountUSDValue,
+  orderSellValueInUSD,
   isPriceReasonable,
   areBoundsReasonable,
   checkCorrectnessOfDeposits,
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -216,7 +216,7 @@ const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedP
   return true
 }
 
-const checkNoProfitableOffer = async (order, tokenInfo, globalPriceStorage = null) => {
+const checkNoProfitableOffer = async (order, exchange, tokenInfo, globalPriceStorage = null) => {
   const currentMarketPriceSlice = await getOneinchPrice(
     await tokenInfo[order.buyToken],
     await tokenInfo[order.sellToken],
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -199,7 +199,7 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
 }
 
 const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
-  const onlinePriceSlice = await getOneinchPrice(quoteTokenData, baseTokenData)
+  const onlinePriceSlice = await getOneinchPrice(baseTokenData, quoteTokenData)
   const onlinePrice = onlinePriceSlice.price
   if (onlinePrice === undefined) {
     console.log("Warning: could not perform price check against price aggregator.")
@@ -219,8 +219,8 @@ const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedP
 
 const checkNoProfitableOffer = async (order, exchange, tokenInfo, globalPriceStorage = null) => {
   const currentMarketPriceSlice = await getOneinchPrice(
-    await tokenInfo[order.buyToken],
     await tokenInfo[order.sellToken],
+    await tokenInfo[order.buyToken],
     globalPriceStorage
   )
   const currentMarketPrice = currentMarketPriceSlice.price
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -164,8 +164,8 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
   return priceFeedSlice
 }
 
-const isPriceReasonable = async (buyToken, sellToken, price, acceptedPriceDeviationInPercentage = 2) => {
-  const onlinePriceSlice = await getOneinchPrice(buyToken, sellToken)
+const isPriceReasonable = async (baseToken, quoteToken, price, acceptedPriceDeviationInPercentage = 2) => {
+  const onlinePriceSlice = await getOneinchPrice(baseToken, quoteToken)
   const onlinePrice = onlinePriceSlice.price
   if (onlinePrice === undefined) {
     console.log("Warning: could not perform price check against price aggregator.")
@@ -174,8 +174,8 @@ const isPriceReasonable = async (buyToken, sellToken, price, acceptedPriceDeviat
     console.log(
       `Warning: the chosen price differs by more than ${acceptedPriceDeviationInPercentage} percent from the price found on ${onlinePriceSlice.source}.`
     )
-    console.log(`    chosen price: ${price} ${buyToken.symbol} bought for 1 ${sellToken.symbol}`)
-    console.log(`    ${onlinePriceSlice.source} price: ${onlinePrice} ${buyToken.symbol} bought for 1 ${sellToken.symbol}`)
+    console.log(`    chosen price: ${price} ${baseToken.symbol} bought for 1 ${quoteToken.symbol}`)
+    console.log(`    ${onlinePriceSlice.source} price: ${onlinePrice} ${baseToken.symbol} bought for 1 ${quoteToken.symbol}`)
     return false
   }
   return true
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -197,8 +197,8 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
   return priceFeedSlice
 }
 
-const isPriceReasonable = async (baseToken, quoteToken, price, acceptedPriceDeviationInPercentage = 2) => {
-  const onlinePriceSlice = await getOneinchPrice(baseToken, quoteToken)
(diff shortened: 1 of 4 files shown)


test/price_utils.js
```diff
@@ -170,4 +175,39 @@ contract("PriceOracle", function (accounts) {
       assert(GEMvalue.eq(new BN(3)))
     })
   })
+  describe("orderSellValueInUSD()", async () => {
+    it.only("Ensures function returns expected values", async () => {
```
bh2smith, line 179:
```suggestion
    it("Ensures function returns expected values", async () => {
```

josojo:
I think we need to touch all places where the pricecheck has been used in the past. E.g. `isPriceReasonable` etc. 

I know it will be quite some work, but we need to do it...

bh2smith:
> I think we need to touch all places where the pricecheck has been used in the past. E.g. isPriceReasonable etc.

I don't see why. Don't we have unit tests for this kinda thing?

josojo:
The unit tests do not cover calls to external parties. For all the unit tests, we just hand over the globalpricestorage object.

If it was correct in the past like that:
```
const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
  const onlinePriceSlice = await getOneinchPrice(quoteTokenData, baseTokenData)
```
Then I think we need to make it like that now:
```
const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
  const onlinePriceSlice = await getOneinchPrice(baseTokenData, quoteTokenData)
```
we need to find all these places..

bh2smith:
@josojo - I fixed two places where it was clear that these needed to be changed, the only last place where it is unclear what prices are expected is here:

https://github.com/gnosis/dex-liquidity-provision/blob/6ae00ae9b4a33c11ce2185f0a343cbeeb4fe24ff/scripts/utils/verify_scripts.js#L132-L136

However, I suspect they don't need to be changed. Perhaps you have more information about this particular spot.

scripts/utils/price_utils.js
```diff
@@ -198,8 +164,8 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
   return priceFeedSlice
 }
 
-const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
-  const onlinePriceSlice = await getOneinchPrice(quoteTokenData, baseTokenData)
+const isPriceReasonable = async (buyToken, sellToken, price, acceptedPriceDeviationInPercentage = 2) => {
```
josojo, line 167: I really like base and quote here. Base and quote describe much better the expected price than buy and sell token.

scripts/utils/price_utils.js
```diff
@@ -92,48 +92,14 @@ const areBoundsReasonable = function (currentPrice, lowestLimit, highestLimit) {
   return currentPriceWithinBounds && boundsCloseTocurrentPrice
 }
 
-// returns undefined if the price was not available
```
josojo, line 95: Please insert it again. This is a non-related change.

scripts/utils/price_utils.js
```diff
@@ -208,16 +174,14 @@ const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedP
...
+    console.log(`    ${onlinePriceSlice.source} price: ${onlinePrice} ${buyToken.symbol} bought for 1 ${sellToken.symbol}`)
     return false
   }
   return true
 }
 
-const checkNoProfitableOffer = async (order, exchange, tokenInfo, globalPriceStorage = null) => {
+const checkNoProfitableOffer = async (order, tokenInfo, globalPriceStorage = null) => {
   const currentMarketPriceSlice = await getOneinchPrice(
     await tokenInfo[order.buyToken],
```
josojo, line 186: Why doesn't it have to be changed here?

josojo (commented):
I will review it later after lunch more in detail

scripts/utils/price_utils.js
bh2smith: Because exchange was not being used. This is just an improvement (removing an unused variable)

scripts/utils/price_utils.js
bh2smith: I have put them back.

scripts/utils/price_utils.js
bh2smith: I will revert this change as well and make a new PR then

scripts/utils/price_utils.js
```diff
@@ -132,8 +132,7 @@ const getDexagPrice = async function (tokenBought, tokenSold, globalPriceStorage
  *
  * Note that the output price is not very accurate, especially for tokens with
  * very high value and very little liquidity. This is due to the fact that 1inch
- * returns a sell and a by price for a given amount, unlike dex.ag, and this function returns only
- * one of the two.
+ * returns a sell and a by price for a given amount
```
bh2smith, line 135:
```suggestion
 * returns a sell and a buy price for a given amount
```

scripts/utils/price_utils.js
josojo: thanks for removing the "exchange" thing out of this PR. But actually, I was talking about why the order of quote and base token is not changed here.

josojo (commented):
> @josojo - I fixed two places where it was clear that these needed to be changed, the only last place where it is unclear what prices are expected is here:
> 
> https://github.com/gnosis/dex-liquidity-provision/blob/6ae00ae9b4a33c11ce2185f0a343cbeeb4fe24ff/scripts/utils/verify_scripts.js#L132-L136
> 
> However, I suspect they don't need to be changed. Perhaps you have more information about this particular spot.

Here, we are just fetch all prices upfront, so that we don't need to do it again in `checkNoProfitableOffer`. I think it does not matter whether we are getting the price or inverse price here, as the funciton` checkNoProfitableOffer` will pick the right one.

scripts/utils/price_utils.js
josojo: I looked into it and I am quite convinced that it needs to be changed.

test/price_utils.js
```diff
@@ -170,4 +175,39 @@ contract("PriceOracle", function (accounts) {
...
+          sellToken: GEMtokenId,
+          expectedValue: "9", // Note that orderSellValueInUSD returns truncated integers so 9.9 becomes 9!
+        },
+      ]
+      for (const order of orders) {
+        const value = await orderSellValueInUSD(order, tokenInfo, globalPriceStorage)
+        assert.equal(value.toString(10), order.expectedValue)
+      }
+    })
+  })
```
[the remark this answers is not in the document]
bh2smith: I just realized that we shouldn’t actually need to register the tokens on the exchange in order to test this function. As long as the token data is provided.

josojo:
I think that the base and quote tokens still have to be switched here?
```
const checkNoProfitableOffer = async (order, tokenInfo, globalPriceStorage = null) => {
  const currentMarketPriceSlice = await getOneinchPrice( <Switch it>)
```
right?

bh2smith:
I can check, but I don’t think so.

bh2smith:
@josojo  - This was already changed perhaps in a previous PR.

```js
const checkNoProfitableOffer = async (order, tokenInfo, globalPriceStorage = null) => {
  const currentMarketPriceSlice = await getOneinchPrice(
    await tokenInfo[order.buyToken],
    await tokenInfo[order.sellToken],
    globalPriceStorage
  )
```

nlordell:
Looking at the `getOneinchPrice` function with:

```js
const baseToken = "DAI";
const quoteToken = "ETH";
```

This would make an HTTP request to:
`https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=ETH&toTokenSymbol=DAI&amount=100000000000000000000`

```
{
	toTokenAmount	"30761722252948536614616",
	fromTokenAmount	"100000000000000000000"
}
```

The price would then be computed to: `toTokenAmount / fromTokenAmount` which would give a result of `307.62`. However:

> https://www.investopedia.com/terms/b/basecurrency.asp 
> When provided with an exchange rate, currency pairs indicate how much of the quote currency is needed to buy one unit of the provided base currency.

That would mean that this price would indicate that 307 quote token (ETH) is required to buy 1 base token (DAI). So I think the issue isn't that the calls to `getOneinchPrice` method need be swapped, but it should be swapped internally. Specifically:
```js
  const url =
    "https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=" +
    quoteToken.symbol +
    "&toTokenSymbol=" +
    baseToken.symbol +
    "&amount=" +
    quoteTokenAmount.toString()
```

should become:
```js
  const url =
    "https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=" +
    baseToken.symbol +
    "&toTokenSymbol=" +
    quoteToken.symbol +
    "&amount=" +
    quoteTokenAmount.toString()
```

OR 

```js
price = amountReceived / amountUsed / decimalCorrection
```

should become

```js
price = (amountUsed / amountReceived) * decimalCorrection
```

Depending if you want ask prices or bid prices.

> @jososo I really like base and quote here. Base and quote describe much better the expected price than buy and sell token.

While this is true about the unit of the price, it doesn't really specify wither the price is an ask or bid price.

bh2smith:
Thanks for uncovering this deeper issue @nlordell. from my understanding of base-quote with regards to buy and sell (or From and To in 1Inch terminology), In the scenario where I want to by `ETH` with `DAI` 
- `DAI` is From token and `ETH` is To token because I am going from `DAI` to `ETH`
- `ETH` is baseToken and DAI is the quoteToken

So, the URL having quote as from and base as to would be correct. This would then mean that we should go with the second proposed option that 

```js
price = (amountUsed / amountReceived) * decimalCorrection
```

I will give this a try in a new branch since reverting all these changes will become rather messy and see how it goes.

nlordell:
> So, the URL having quote as from and base as to would be correct. This would then mean that we should go with the second proposed option that

That really depends if you want your price to be an ask order or a bid order. So for the currency pair `ETH-DAI`:
- The ask price would use the API call with `fromToken=ETH&toToken=DAI` and compute the price as `toTokenAmount / fromTokenAmount` (this is my "option 1")
- The bid price would use the API call with `fromToken=DAI&toToken=ETH` and compute the price as `fromTokenAmount / toTokenAmount` (this is my "option 2")

So both options are correct, they just compute different things, and depending what the price estimate is used for one is probably better than the other.

Alternatively you could use the **middle** of the spread and average the bid and ask prices.

bh2smith:
Closing in favor or #395

josojo:
>So I think the issue isn't that the calls to getOneinchPrice method need be swapped, but it should be swapped internally.

Yes, we need to do both. Change it internally, in order to get the function `getOneInchPrice` return the right value and swap it externally (in the function call) to get the correct - same behavior as before - behavior.

Personally, I don't care whether we are using the ask or bid-price

PR #390, opened 2020-07-29
[Easy] More expresive and correct method names

bh2smith:
changes `buildRequestWithdraw` to  `buildWithdrawRequest` (because that's what it is) and `buildWithdraw` to `buildWithdrawClaim`. I think we should also move all "transaction builders" into their own file call `transaction_builders.js` but will save this for another time since a lot of import statements need to be moved as well.


josojo (approved):
thanks

PR #391, opened 2020-07-29
[Synthetix] Economic Viability of Order Placement.

bh2smith:
The goal of this PR is to evaluate cost of order placement before hand and compare with the expected "surplus" as defined by the trading fees as a percentage of the magnitude of the order placement (i.e. sell amount or buy amount)

https://github.com/gnosis/dex-liquidity-provision/blob/d9435ab6ddcc384e332293c8a71dc92ae2d552bd/scripts/synthetix/facilitate_trade.js#L93-L96

Currently, the code includes a gas estimation before order placement, which is to be multiplied by the `gasPrice` to determine the amount of Ether. There is also already a ether price determined by the script which can be used to estimate the transaction costs. It remains to compare with the order size and determine whether the order placement is economically viable.

It will be important to note that the fees for orders selling `sUSD` are different from those selling `sETH`


bh2smith:
Archiving repo.

PR #395, opened 2020-07-30
Fix OneInch Swapped base-quote (i.e. inverted prices)

### LINKED ISSUE #383: OrderSellValueUSD has reversed base and quote tokens

bh2smith:
Closes #383

Turned out to be nice and clean after all!

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -131,6 +131,7 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
     throw new Error("Invalid token input for retrieving price from aggregator.")
   }
   const quoteTokenAmount = toErc20Units("1", quoteToken.decimals)
+  console.log(`Requesting ${baseToken.symbol}-${quoteToken.symbol} price from external source: 1Inch`)
   const url =
     "https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=" +
     quoteToken.symbol +
@@ -165,8 +166,14 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
   return priceFeedSlice
 }
 
-const isPriceReasonable = async (baseTokenData, quoteTokenData, price, acceptedPriceDeviationInPercentage = 2) => {
-  const onlinePriceSlice = await getOneinchPrice(baseTokenData, quoteTokenData)
+const isPriceReasonable = async (
+  baseTokenData,
+  quoteTokenData,
+  price,
+  acceptedPriceDeviationInPercentage = 2,
+  globalPriceStorage = {}
+) => {
+  const onlinePriceSlice = await getOneinchPrice(baseTokenData, quoteTokenData, globalPriceStorage)
   const onlinePrice = onlinePriceSlice.price
   if (onlinePrice === undefined) {
     console.log("Warning: could not perform price check against price aggregator.")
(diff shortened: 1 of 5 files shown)


josojo:
Why don't we need to change it over here?
```
const checkNoProfitableOffer = async (order, tokenInfo, globalPriceStorage = null) => {
  const currentMarketPriceSlice = await getOneinchPrice(
    await tokenInfo[order.buyToken],
    await tokenInfo[order.sellToken],
    globalPriceStorage
  )
```

bh2smith:
@josojo - If it needs to be changed maybe it would be good to include some comments about what this is attempting to evaluate. At this point I am unable to determine what this measurement is intending to do. Are there no unit tests for this function?

scripts/utils/price_utils.js
```diff
@@ -131,6 +131,7 @@ const getOneinchPrice = async function (baseToken, quoteToken, globalPriceStorag
     throw new Error("Invalid token input for retrieving price from aggregator.")
   }
   const quoteTokenAmount = toErc20Units("1", quoteToken.decimals)
+  console.log(`Requesting ${baseToken.symbol}-${quoteToken.symbol} price from external source: 1Inch`)
```
josojo, line 134: Maybe we want to remove this?

josojo (approved):
Thanks Ben, especially for going this extra step to speed up all our tests! great.
Splendid

scripts/utils/price_utils.js
bh2smith: You don’t want to know when an external price source is queried? I thought it was good to know. Especially for that user whose script failed after 750 seconds. It was likely due to this request. Then he would have known.

josojo: I am indifferent, but I think that it spams the inexperienced user with too much information. They won't be able to make this mentioned connection
A cleaner way to keep it,  might be to enable it in debug mode.

bh2smith: OK I will remove it.

bh2smith: Mostly because this method does have a debug optional parameter.

PR #410, opened 2020-08-06
Remove unused dependency @gnosis.pm/mock-contract

fedgiac:
I was surprised that the major update of `@gnosis.pm/mock-contract` in PR #407 didn't cause any failure and checked where this package is used. It seems it's not used anywhere in the JS code and I don't see any NPM script that would need it.

Let me know if you know why this package should stay, otherwise you can approve this PR.


josojo (approved):

fedgiac:
Sorry Alex, I force pushed because the automerge reintroduced the package I wanted to remove. Re-requesting review just to be safe, then I merge.

josojo (approved):

PR #419, opened 2020-08-11
Npm package dependency - remove redundant config file

fedgiac:
Suggestion for PR #409.

It removes a redundant config file. Not sure how much the extra fine `inject_networks.sh` improves the situation compared to the redundant config file.
(Alex, please double check that everything works fine with a Mac.)


josojo:
In agreement with federico, I am closing it.

PR #434, opened 2020-08-20
FleetFactoryDeterministic smart contract change

andre-meyer:
In order to allow us to precalculate the safe addresses that will be deployed, this PR changes the FleetFactory (in a new file) to use `createProxyFromNonce` instead of `createProxy`

The salt that is used is the salt-passed abi encoded with the current index hashed.

Compare the FleetFactory with this contract (all that was changed is the call to `proxyFromNonce` and the salt)
https://github.com/gnosis/dex-liquidity-provision/blob/master/contracts/FleetFactory.sol


contracts/FleetFactoryDeterministic.sol
```diff
@@ -0,0 +1,48 @@
...
+          template,
+          "",
+          uint256(
+            keccak256(
+              abi.encodePacked(saltNonce, i)
+            )
+          )
+        )
+      );
+      fleet[i] = proxy;
```
rmeissner, line 32: maybe check that `proxy != address(0)` as the deployment failed in that case (should normally not happen ... most likely case is that a contract already exists at that address)

andre-meyer: But 5 lines later it does the check? https://github.com/gnosis/safe-contracts/blob/development/contracts/proxies/GnosisSafeProxyFactory.sol#L53

Either way, I added it here too

contracts/FleetFactoryDeterministic.sol
```diff
@@ -0,0 +1,49 @@
...
+contract FleetFactoryDeterministic {
+  GnosisSafeProxyFactory public proxyFactory;
+
+  event FleetDeployed(address indexed owner, address[] fleet);
+
+  constructor(GnosisSafeProxyFactory _proxyFactory) public {
+    proxyFactory = _proxyFactory;
+  }
+
+  function deployFleetWithNonce(address owner, uint256 size, address template, uint256 saltNonce) external {
```
josojo, line 15: Should this function return an array of addresses[] with the all the fleets, in order to get the fleets very easily?

andre-meyer:
I will deploy this version now for an initial test - there shouldn't be any more changes to the smart contract necessary, and it's only a "helper" contract. I'll deploy and add the artifact to the `networks.json`. I will write more tests for this at a later point.

andre-meyer:
```
   Deploying 'FleetFactoryDeterministic'
   -------------------------------------
   > transaction hash:    0xcaa53e63fa6e2f85ed641cfc0550fe37b731bfa9f0d64911f523851b762de37d
   > Blocks: 0            Seconds: 0
   > contract address:    0xc4c4Ac527FF49A146017aee57Cd2c9D2379EDac2
   > block number:        745
   > block timestamp:     1598881822
   > account:             0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1
   > balance:             98.13800972
   > gas used:            285984 (0x45d20)
   > gas price:           5 gwei
   > value sent:          0 ETH
   > total cost:          0.00142992 ETH
```

On Mainnet with roughly 240 GWEI right now, it would be 0.0689221 ETH for $29.91219

josojo:
@fedgiac Could you review this PR? While I also see the strange gas cost difference during deployment, the smart contract deployed seems to be correct. Hence I used Andres smart contract deployment and put the address into the networks.json

fedgiac:
I made PR #455 to merge the new deployment addresses into here.
While I don't see any failures from Etherscan, I'm under the understanding that there were unexpected errors when interacting with the deployed contract. The deployment method was also not the standard one use in this repo but ad hoc for this contract, so it's not easy to replicate what happened.
Note that the [gas used in the deployment](https://dashboard.tenderly.co/tx/main/0xd837747066c20a7162cdd72d950ddba6e4f3e6f22b2ffce9fe97a0404482a98c/gas-usage) is not even enough to [pay for the `CREATE` opcode](https://github.com/djrtwo/evm-opcode-gas-costs/blob/master/opcode-gas-costs_EIP-150_revision-1e18248_2017-04-12.csv) after discounting the initial gas. I think this is a bug in Etherscan contract verification.

fedgiac (commented):
I like that the addresses are computed offline instead of relying on web3 calls.
The changes look fine, there are still some small details I'd like to see changed or discussed. I implemented most of the changes I propose in #456.

PR #435, opened 2020-08-21
[do-not-merge] List old brackets

fedgiac:
This PR creates a modified version of the script `get_deployed_brackets` that works with brackets created using `GnosisSafeProxyFactory` directly.

This is more of a test bench than production code, but could be used to get information on the old brackets for Dune.

I also notice that the brackets obtained by this script are more than those available in our list of old brackets, and in general the existing list of old brackets contains many inconsistencies.

@josojo, this PR might be of interest to you, it's a first attempt to use batch requests.
I fear that batch requests are not going to solve our issues with `Error: Invalid JSON RPC response: {}` however, I encountered this error many times and inconsistently when trying out this script.

Test plan
```
npx truffle exec scripts/get_old_deployed_brackets.js --masterSafe=0x583788f490a278DB2a8d6D7226264b4985f75678 --outputFile=test.csv --network=mainnet
```
Sample output:
```
<snip>
The following addresses have been deployed from your MASTER SAFE: <snip>
Records written to file.
Number of owned brackets: 426.
Number of errors when retrieving proxy owner info: 225 out of a total of 3224.
Tokens traded by the brackets: [
   1,  7, 15,  9, 16,  4,
   8, 18, 11, 13,  3, 10,
  19,  2,  0
]
```

Note that the error count that is printed on screen when searching for the owned brackets does not necessarily mean that there is an error in the retrieved information, since some Safe proxies do not use the right templates or their owners cannot be retrieved for other legitimate reasons.

### DIFF
--- a/scripts/utils/batch_request.js
+++ b/scripts/utils/batch_request.js
@@ -1,0 +1,32 @@
+// https://github.com/ethereum/web3.js/issues/1446
+
+module.exports = function (web3) {
+  class BatchRequest {
+    constructor() {
+      this.promises = []
+      this.batch = new web3.BatchRequest()
+    }
+
+    add(call, ...params) {
+      this.promises.push(
+        new Promise((resolve, reject) => {
+          const request = call.request(...params, (error, data) => {
+            if (error) {
+              reject(error)
+            } else {
+              resolve(data)
+            }
+          })
+          this.batch.add(request)
+        })
+      )
+    }
+
+    execute() {
+      this.batch.execute()
+      return this.promises
+    }
+  }
+
+  return BatchRequest
+}
--- a/scripts/get_old_deployed_brackets.js
+++ b/scripts/get_old_deployed_brackets.js
@@ -1,0 +1,89 @@
+const createCsvWriter = require("csv-writer").createObjectCsvWriter
+const { sleep } = require("./utils/js_helpers")
+const { decodeOrders } = require("@gnosis.pm/dex-contracts")
+const { fetchTokenInfoFromExchange, getExchange, getSafe, isOnlySafeOwner } = require("./utils/trading_strategy_helpers")(
+  web3,
+  artifacts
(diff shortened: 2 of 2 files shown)


scripts/utils/batch_request.js
```diff
@@ -0,0 +1,32 @@
+// https://github.com/ethereum/web3.js/issues/1446
+
+module.exports = function (web3) {
+  class BatchRequest {
+    constructor() {
+      this.promises = []
+      this.batch = new web3.BatchRequest()
```
josojo, line 7: Why did you choose this Batching Method over the discussed method on solidity level? Because it is simpler?


Yes, we knew that it won't fix the issue. The hope was just that it would occur less frequently as we are making less requests. If you have a better idea than batching, let us know.

josojo:
I tried this script and got the "Invlaid Json RPC response" sooo often.

A better way to get the old safes is maybe this dune query:
https://explore.duneanalytics.com/queries/8874/source

bh2smith:
What is the status of this PR? Based on the title, it looks like #451 does the job here.

fedgiac:
This PR is never supposed to be merged, it's here for two reasons:
1. get a trusted, replicable list of all the brackets deployed directly with the proxy factory with a specific owner
   - all old brackets will be withdrawn soon (also with the help of this PR), making this script useless afterwards
   - different from #451 because that PR retrieves deterministically generated brackets, while this retrieves legacy brackets
2. play around with batch requests
   - `scripts/utils/batch_request.js ` could be reused somewhere else in this repo. However, since it doesn't fix the issue with `Invalid JSON RPC response`, it's unclear if adopting batch requests provides any value.
   - since the script is heavy in terms of requests to Infura, other batching mechanisms can be tested for reliability, like the one suggested by Alex.

bh2smith:
Can we close this?

PR #438, opened 2020-08-26
CMM changes for working transactions in the App

andre-meyer:
# Description
This PR removes some unnecessary bundling of transactions, to ensure the transaction can be executed via the App, which does not have access to `DELEGATE_CALL`
It also disables a safety-check for safe-ownership if they were not deployed yet, to accomodate for the precalculated safe brackets used in the CMM App.

### DIFF
--- a/networks.json
+++ b/networks.json
@@ -12,10 +12,15 @@
     }
   },
   "FleetFactoryDeterministic": {
+    "1": {
+      "links": {},
+      "address": "0xb1e1ae60607f24a6647d0339050cb69c312f43ee",
+      "transactionHash": "0xd837747066c20a7162cdd72d950ddba6e4f3e6f22b2ffce9fe97a0404482a98c"
+    },
     "4": {
       "links": {},
-      "address": "0xa8ECf90603AD40a1bB009FB58540d5A9A6db7d5B",
-      "transactionHash": "0xa7dad010d6dff58d0bea66c6e67518b9d95d7d7004686ba898f641cb613f78d2"
+      "address": "0xf02419728f01566b3e9173c6b5c652670105dd96",
+      "transactionHash": "0xeb2e70e947891a519a8a10d1d68ac2d00b99e87be8d714bad9c06ab8fa4a2864"
     }
   },
   "GnosisSafe": {
--- a/networks.json
+++ b/networks.json
@@ -11,6 +11,13 @@
       "transactionHash": "0xf8256a6b6c92a95f9342d0dbb8afee83d1dd6bec8510278893f568ae0114ee3f"
     }
   },
+  "FleetFactoryDeterministic": {
+    "4": {
+      "links": {},
+      "address": "0xa8ECf90603AD40a1bB009FB58540d5A9A6db7d5B",
+      "transactionHash": "0xa7dad010d6dff58d0bea66c6e67518b9d95d7d7004686ba898f641cb613f78d2"
+    }
+  },
   "GnosisSafe": {
     "1": {
       "links": {},
@@ -40,8 +47,8 @@
     },
     "4": {
       "links": {},
-      "address": "0xca39A48b3b0C8284bfF8f2299539e2BbaBE714de",
-      "transactionHash": "0x3120664b4b966dc5fa1e13016856d6ef6d39819ed53706a6abee3665f488017b"
+      "address": "0xAd0B040edF19bB81BDb2f14B26c7ba5F7ec263C2",
+      "transactionHash": "0x34f548082b798f50d41ddd036c7fdf15cafda442c7f55947a1a46db8f5632924"
     }
   },
   "MultiSend": {
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -22,7 +22,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const ERC20 = artifacts.require("ERC20Detailed")
   const BatchExchange = artifacts.require("BatchExchange")
   const GnosisSafe = artifacts.require("GnosisSafe")
-  const FleetFactory = artifacts.require("FleetFactoryDeterministic")
+  const FleetFactory = artifacts.require("FleetFactory")
 
   const exchangePromise = BatchExchange.deployed()
   const gnosisSafeMasterCopyPromise = GnosisSafe.deployed()
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -22,12 +22,12 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const ERC20 = artifacts.require("ERC20Detailed")
   const BatchExchange = artifacts.require("BatchExchange")
   const GnosisSafe = artifacts.require("GnosisSafe")
-  const FleetFactory = artifacts.require("FleetFactory")
+  const FleetFactory = artifacts.require("FleetFactoryDeterministic")
 
   const exchangePromise = BatchExchange.deployed()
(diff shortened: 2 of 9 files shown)


fedgiac:
The app uses an external tool to bundled together transactions, which makes `buildBundledTransaction` redundant. In the scripts of this repo, however, the transactions must be manually bundled before being executed.
To have the list of transactions for the app and leave the scripts in a working state you could introduce a function that creates the transactions to be bundled (which is then used in the app) and then bundle its output in `buildTransferApproveDepositFromList`.

The changes to `isOnlySafeOwner` are not compatible with the expectations of the script, since at any point we use it we want the script to fail if no bracket is present and we definitely can't assume it has been created correctly.
To help the app, I think the ownership check should just be removed from any `build*` functions, since any ownerwhip check should be run in advance before the call and not every time a function creating transactions is called.

Note also that you are merging into the branch `deterministic-fleet-deploy` and not master.
By the way, it's better to split this PR into the two separate changes, it makes the discussion more focused.

andre-meyer:
> The app uses an external tool to bundled together transactions, which makes `buildBundledTransaction` redundant. In the scripts of this repo, however, the transactions must be manually bundled before being executed.
> To have the list of transactions for the app and leave the scripts in a working state you could introduce a function that creates the transactions to be bundled (which is then used in the app) and then bundle its output in `buildTransferApproveDepositFromList`.
> 
> The changes to `isOnlySafeOwner` are not compatible with the expectations of the script, since at any point we use it we want the script to fail if no bracket is present and we definitely can't assume it has been created correctly.
> To help the app, I think the ownership check should just be removed from any `build*` functions, since any ownerwhip check should be run in advance before the call and not every time a function creating transactions is called.
> 
> Note also that you are merging into the branch `deterministic-fleet-deploy` and not master.
> By the way, it's better to split this PR into the two separate changes, it makes the discussion more focused.

Thank you, I will make those changes - but for the demo tomorrow, I will simply use this branch for it. If somebody from the dex-liquidity-provision team would be available to do those changes for me it would be very appreciated.

The reason this PR merges into `deterministic-fleet-deploy` is because it is dependant on the smart contract that was changed, which is also still under review.

networks.json
```diff
@@ -12,10 +12,15 @@
...
+      "links": {},
+      "address": "0xb1e1ae60607f24a6647d0339050cb69c312f43ee",
+      "transactionHash": "0xd837747066c20a7162cdd72d950ddba6e4f3e6f22b2ffce9fe97a0404482a98c"
+    },
     "4": {
       "links": {},
-      "address": "0xa8ECf90603AD40a1bB009FB58540d5A9A6db7d5B",
-      "transactionHash": "0xa7dad010d6dff58d0bea66c6e67518b9d95d7d7004686ba898f641cb613f78d2"
+      "address": "0xf02419728f01566b3e9173c6b5c652670105dd96",
+      "transactionHash": "0xeb2e70e947891a519a8a10d1d68ac2d00b99e87be8d714bad9c06ab8fa4a2864"
```
fedgiac, line 23: I think something went wrong in the deployment.
See the gas used by the contract creation on [mainnet](https://etherscan.io/tx/0xd837747066c20a7162cdd72d950ddba6e4f3e6f22b2ffce9fe97a0404482a98c) and [Rinkeby](https://rinkeby.etherscan.io/tx/0xeb2e70e947891a519a8a10d1d68ac2d00b99e87be8d714bad9c06ab8fa4a2864) and compare it with the gas used by the deployment of the [original FleetFactory](https://etherscan.io/tx/0xa2125e6e30056c79569d204197df05a3f9fe09103cfde5569209e9b267518317).

In general, I'd use Truffle instead of relying on external tools for deploying our contracts, in this way we are in control of the deployment method and check that everything went through correctly.

scripts/utils/trading_strategy_helpers.js
```diff
@@ -56,8 +58,18 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @returns {boolean} whether owned is indeed owned only by master
    */
   const isOnlySafeOwner = async function (masterAddress, owned) {
-    const ownedSafe = typeof owned === "string" ? await getSafe(owned) : owned
-    const ownerAddresses = await ownedSafe.getOwners()
+    let ownedSafeInstance
```
josojo, line 61: @andre-meyer These code changes here were only made for debugging, right? We don't need to adapt to it?

scripts/wrapper/withdraw.js
```diff
@@ -74,15 +74,16 @@ module.exports = function (web3, artifacts) {
         const token = tokenBracketPairs[index][1]
         const bracketAddress = tokenBracketPairs[index][0]
         const usdValue = await amountUSDValue(amount, token, globalPriceStorage)
-        if (usdValue.gte(ONE)) {
+        log(`Request for ${token.symbol} on bracket ${bracketAddress} USD value: ${usdValue}$`)
```
josojo, line 77: @andre-meyer same question here: You actually don't need this code change, right? Or is it because the price feed on rinkeby is unreliable?

bh2smith:
Is this PR slowly becoming deprecated in favour of some other more recent ones? If so, could we label them here and close this?

scripts/utils/trading_strategy_helpers.js
andre-meyer: Federico suggested to split this check into a different function. We do need the ability to fund/order with not-yet-created brackets though, for this to work with precalculated addresses.

scripts/wrapper/withdraw.js
andre-meyer: This change is necessary for testing to see if a withdraw can even work - It can be removed here, but I suggest we move this check out of the withdraw logic when the withdraw logic is used like an API, because the implementation should have control over these checks.

scripts/utils/trading_strategy_helpers.js
fedgiac: To my understanding, we don't need these changes to `isOnlySafeOwner` anymore since PR #441.

scripts/wrapper/withdraw.js
josojo: Sorry, I am not sure I understand. You can use the logic already like an API, just with this additional check. Why. would you want to do it without the check. 

I am not sure how I can "move this check out of the withdraw logic", since over there is also not the withdraw logic, just the withdraw building and checking logic. 

Should I make the ONE in the usdValue.gte(ONE) as a parameter and then you can use it as an API with an arbitrary value?  LIke that:
ttps://github.com/gnosis/dex-liquidity-provision/pull/new/movingBalanceCheck

scripts/utils/trading_strategy_helpers.js
```diff
@@ -47,6 +47,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    * @returns {SmartContract} Gnosis Safe Contract
    */
   const getSafe = function (safeAddress) {
+    console.log("trying to get safe at")
```
bh2smith, line 50: Trying to get safe at what?

scripts/wrapper/withdraw.js
```diff
@@ -207,7 +207,7 @@ module.exports = function (web3, artifacts) {
 
     log("Started building withdraw transaction.")
     const transactionPromise = buildWithdrawAndTransferFundsToMaster(argv.masterSafe, withdrawals)
-
+    console.log(transactionPromise)
```
bh2smith, line 210: Was this meant to be left here? If, so I would prefer if you used the logger defined at the beginning of the function that only logs when `debug == true`

bh2smith (approved):
Looks good, not sure about a couple of these console logs.

bh2smith (changes requested):
I think this PR is expired. Can we please close it?

PR #441, opened 2020-08-27
Remove ownership check from transaction builder functions

fedgiac:
Remove ownership checks from all transaction bundling functions.
The checks are instead performed in the scripts using these functions. 
Note that the check in `complete_liquidity_provision.js` doesn't need to be included since it's already performed in the execution of the script.

See #438 for motivation.

### Test plan

Success: unit tests.
Failure:
```
npx truffle exec scripts/deposit.js --masterSafe=0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636 --depositFile=examples/exampleDepositList.json --network=rinkeby --verify
npx truffle exec scripts/transfer_approve_deposit.js --masterSafe=0xb947de73ADe9aBC6D57eb34B2CC2efd41f646636 --depositFile=examples/exampleDepositList.json --network=rinkeby
```

### DIFF
--- a/scripts/deposit.js
+++ b/scripts/deposit.js
@@ -1,8 +1,7 @@
 const fs = require("fs").promises
-const assert = require("assert")
 
 const { signAndSend, transactionExistsOnSafeServer } = require("./utils/gnosis_safe_server_interactions")(web3, artifacts)
-const { buildDepositFromList, isOnlyFleetOwner } = require("./utils/trading_strategy_helpers")(web3, artifacts)
+const { buildDepositFromList, assertIsOnlyFleetOwner } = require("./utils/trading_strategy_helpers")(web3, artifacts)
 const { promptUser } = require("./utils/user_interface_helpers")
 const { default_yargs } = require("./utils/default_yargs")
 const argv = default_yargs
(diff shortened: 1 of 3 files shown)


scripts/deposit.js
```diff
@@ -31,6 +32,14 @@ module.exports = async (callback) => {
     console.log("Preparing transaction data...")
     const transaction = await buildDepositFromList(masterSafe.address, deposits, true)
 
+    assert(
+      await isOnlyFleetOwner(
+        masterSafe.address,
+        deposits.map(({ bracketAddress }) => bracketAddress)
+      ),
+      "All depositors must be owned only by the master Safe"
+    )
```
bh2smith, line 41: I actually wonder if this verification is even necessary. For example, what if I wanted to use an already existing safe as a bracket trader and simply configure it in a way that the `FundAccount` is its owner. I don't see that this should cause problems.

bh2smith (changes requested):
This is ok, but I think we could easily live without this constraint being enforced at all (since it doesn't make much sense in the first place and wouldn't hinder the functionality in anyway). We also see now that there is some code duplication, that could probably be avoided.

bh2smith:
That is, I would be in favour of removing this verification completely.

josojo:
Personally, I prefer to have the checks in there @bh2smith. While you are right, that there other usecases, it's a nice safety check. :) 

Code changes are looking good. approved from that side ;)

bh2smith (approved):

scripts/deposit.js
fedgiac: Silly example: you, user, copypaste the example command from the readme:
```
npx truffle exec scripts/transfer_approve_deposit.js --masterSafe=$MASTER_SAFE --depositFile="./examples/exampleDepositList.json" --network=$NETWORK_NAME
```
You replaced `MASTER_SAFE` in your env, but forgot to change `depositFile` and run the script. You'd be (irrecoverably) depositing funds from your master safe to the example addresses. It's a user error, but something easily preventable by us, while the use case you describe is very specific. I think it's fair to have a user who has this particular need comment out the check manually.

josojo (approved):

PR #457, opened 2020-09-04
Remove Math.random() from FleetFactoryDeterministic tests

fedgiac:
Apply best practices! :)

Tests for FleetFactoryDeterministic are more reproducible after removing explicit use of randomness.

### Test plan

Run `yarn test test/fleet_factory_deterministic.js` twice and observe that all tests are passing in both runs.


josojo (approved):
First principle: Fast, independent, **repeatable**, self-validating,  thorough :)

PR #466, opened 2020-09-08
hotfix for parameter handover for order bundling

josojo:
There was a tiny error with the parameter handover.

This PR fixes the issue.

Testplan:

See that a new deployment runs smoothly and no longer creates this error:
`Error: invalid number value (arg="validUntil", coderType="uint32", value=true)`


scripts/utils/trading_strategy_helpers.js
```diff
@@ -274,6 +274,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     debug = false,
     expiry = DEFAULT_ORDER_EXPIRY
   ) {
+    console.log(debug)
```
fedgiac:
```suggestion
```
Leftover!

fedgiac (approved):
Strange issue, I tested these lines for a while and everything behaved as expected.

This PR is good. I'll look later into this just in case there was some other issue hiding in here.

PR #470, opened 2020-09-09
[HotFix] inconsistent token_address <-> tokenAddress in airdrop file formats

bh2smith:
File format changes to the airdrop have screwed up the transfer file parsing. This fixes it in the easier way without having to change al the Dune dashboard queries.


fedgiac (approved):
Looks good!

PR #488, opened 2020-09-17
No empty withdrawing

fedgiac:
The withdraw scripts should not propose to the user to send an empty withdraw transaction.

### Test plan

See failure when running one of the withdraw script.
```
npx truffle exec scripts/request_withdraw.js --masterSafe=0x0000000000000000000000000000000000000000 --brackets=0x0000000000000000000000000000000000000000 --network=rinkeby
```


josojo (approved):
thanks

PR #489, opened 2020-09-17
Simpler retrieval of deployed brackets

fedgiac:
The function `getDeployedBrackets` now returns all deployed brackets, including the ones deployed deterministically. This simplifies some calls and makes `scripts/balance_viewer.js` work as expected.

### Test plan

Note that `getDeployedBrackets` is only used in `scripts/balance_viewer.js` and `scripts/get_deployed_brackets.js`.

Compare the output of the following command with this PR and on master (50 vs 10 brackets found).
```
npx truffle exec scripts/balance_viewer.js --network=rinkeby --masterSafe=0x04a82eC4857694a28bdDb9cE8E13DCbc090928e7
```

See that the output of the following command on master is the same with this PR and on master.
```
npx truffle exec scripts/get_deployed_brackets.js --masterSafe=0x04a82eC4857694a28bdDb9cE8E13DCbc090928e7 --network=rinkeby --outputFile=testPr.csv
```

### DIFF
--- a/scripts/get_deployed_brackets.js
+++ b/scripts/get_deployed_brackets.js
@@ -4,7 +4,6 @@ const {
   fetchTokenInfoFromExchange,
   getExchange,
   getDeployedBrackets,
-  getDeployedDeterministicBrackets,
 } = require("./utils/trading_strategy_helpers")(web3, artifacts)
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
💯

PR #493, opened 2020-09-22
Correct one test of deterministic fleet deployment

josojo:
it hurts, but the promises in the test were not awaited.

 Fortunately, at least the function was correct.


fedgiac (approved):
Good catch. Anyway, I did quite a lot of single-transact deployments that relied on `calcSafeAddresses`, so I think we would have noticed if the function had been broken!

PR #505, opened 2020-09-28
Add withdraw helper to exports of wrapper/withdraw.js

andre-meyer:
An export was missing that I'd like to use in an external implementation.

**Edit**: Turns out a few more exports were needed. Now the branch name is wrong. Sorry about that 💀


scripts/wrapper/withdraw.js
```diff
@@ -107,7 +107,7 @@ module.exports = function (web3, artifacts) {
               return
             }
             const usdValue = await amountUSDValue(amount, tokenData, globalPriceStorage)
-
+            log(`USD value for ${amount} ${tokenData.symbol} on bracket ${bracketAddress} is ${usdValue}$`)
```
fedgiac, line 110: It would be very nice if the scripts had different logging levels, so that a variable level of detail could be expressed in the logs.
However, since there is only one logging level I think this is too much information: it would just fill the terminal with unnecessary information when doing large withdrawals.

fedgiac (commented):
I see the issue, exports are needed.

Would it help to have a function that returns the transactions created by the three functions `prepare*` instead of having to build them manually from `getWithdrawalsAndTokenInfo`?

bh2smith (approved):
What is the status of this PR? Can we get it merged?

.gitignore
```diff
@@ -1,7 +1,7 @@
 .DS_Store
 .idea
 
-build
+/build
 node_modules
```
bh2smith, line 5: Why on build and not node_modules? is this necessary?

networks.json
```diff
@@ -1,4 +1,42 @@
 {
+  "BatchExchange": {
```
bh2smith, line 2: All these network changes were probably unintended since we have intentionally removed them quite recently... I would expect this file to remain unchanged.

scripts/utils/dependencies.js
```diff
@@ -11,10 +11,10 @@ module.exports = function (web3, artifacts) {
...
   }
 
-  const BatchExchange = makeContract(require("@gnosis.pm/dex-contracts/build/contracts/BatchExchange.json"))
-  const GnosisSafe = makeContract(require("@gnosis.pm/safe-contracts/build/contracts/GnosisSafe.json"))
-  const GnosisSafeProxyFactory = makeContract(require("@gnosis.pm/safe-contracts/build/contracts/GnosisSafeProxyFactory.json"))
-  const MultiSend = makeContract(require("@gnosis.pm/safe-contracts/build/contracts/MultiSend.json"))
+  const BatchExchange = makeContract(require("../../temp/build/contracts/BatchExchange.json"))
+  const GnosisSafe = makeContract(require("../../temp/build/contracts/GnosisSafe.json"))
+  const GnosisSafeProxyFactory = makeContract(require("../../temp/build/contracts/GnosisSafeProxyFactory.json"))
+  const MultiSend = makeContract(require("../../temp/build/contracts/MultiSend.json"))
```
bh2smith, line 17: Please see above comment on the networks file, we shouldn't be importing these artifacts in this way.

test/scripts/withdraw.js
```diff
@@ -376,9 +376,10 @@ contract("Withdraw script", function (accounts) {
         assert.equal(requestedWithdrawal, "0", "A withdrawal request is still pending")
       }
     })
-    it("claims only amounts worth > 1 USD", async () => {
+    /*
+    // TODO: Write a test to somehow check it doesn't try to withdraw from all brackets if only one side is funded?
```
bh2smith, line 380: Not sure that this is desired, because even if only one side is funded (initially), there could be trades and the brackets could hold either token.

bh2smith (changes requested):
What is the reason for all these temp JSON files to be included? I suspect most of the changes included here are undesired.

bh2smith:
Stale PR - closing because we are archiving the repo.

PR #516, opened 2020-10-08
Support Parity Nodes for GasEstimation

fleupold:
Our current gas estimation expects eth.call to return a revert message, which contains the required amount of gas. This works on geth nodes, on parity however the reverted call causes an exception to be thrown which is not handled in the current code path. This PR adds handling and gas estimate extraction form the exception's error message.

xDAI only runs with parity nodes, so this is needed in order to get the scripts running on xDAI.

### Test Plan

running complete liquidity provision on xDAI (next PR).


bh2smith (approved):
Nice and easy!

fedgiac (approved):
I don't know how the st

Issue #517, opened 2020-10-08
Write Safe Artefacts with deployed addresses on Migration


fleupold:
Our Safe Contracts are currently not written to the artefacts folder with their originally deployed addresses. On migration, we instead add the addresses manually to `networks.json` and inject them into the freshly compiled artefacts.

This is dangerous as we would not notice address changes, e.g. when the Safe upgrades.

@rmeissner does the Safe expose a method to get the currently deployed artefact from within it's own package that we could use in our JS rather than `const GnosisSafe = artifacts.require("GnosisSafe")`? I believe the reason we use the artefact from our local build folder is so that we can access the ganache deployment during tests.

For GP we "copy" the artefact whenever we do a migration cf.

https://github.com/gnosis/dex-liquidity-provision/blob/5504808e71ba9b3f3950cac806b2b8bb46f2db7f/migrations/2_dev_dependencies.js#L8-L18

However, this would also not detect upgrades to the underlying contract unless we re-migrate.

rmeissner:
you mean something like https://github.com/gnosis/safe-modules/blob/master/allowances/test/allowanceSingle.js#L5 ?

fleupold:
Yes, I think the issue we had with this approach is that we still need to deploy the dependency e.g. when running everything locally on ganache. In this case e.g. our scripts would have to check whether to use the local artefact or the artefact from the npm package.

PR #518, opened 2020-10-08
Xdai deployment

fleupold:
This PR upates the dependencies to contain xDAI ready contracts and performs an xDAI migration.

```
yarn truffle migrate --network xdai
```

Note that the addresses for the Safe contracts (GnosisSafeProxyFactory, GnosisSafe, MultiSend) have been added to networks.json by hand, partially because of https://github.com/gnosis/safe-contracts/issues/231 and partially due to https://github.com/gnosis/dex-liquidity-provision/issues/517

### Test Plan
```
yarn truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=8 --lowestLimit=300 --highestLimit=400 --currentPrice=340 --masterSafe=$MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=0 --numBrackets=10 --network=$NETWORK_NAME
```

### DIFF
--- a/networks.json
+++ b/networks.json
@@ -67,6 +67,10 @@
     "4": {
       "links": {},
       "address": "0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F"
+    },
+    "100": {
+      "links": {},
+      "address": "0x6851D6fDFAfD08c0295C392436245E5bc78B0185"
     }
   },
   "GnosisSafeProxyFactory": {
@@ -111,6 +115,10 @@
       "links": {},
       "address": "0x8D29bE29923b68abfDD21e541b9374737B49cdAD",
       "transactionHash": "0x6fabaeeba1dba60983da94818965ecbac7b2ed81fc71c08d6582e29dc05745c7"
+    },
+    "100": {
+      "links": {},
+      "address": "0x8D29bE29923b68abfDD21e541b9374737B49cdAD"
     }
   }
 }
--- a/networks.json
+++ b/networks.json
@@ -44,7 +44,7 @@
     },
     "100": {
       "links": {},
-      "address": "0x6851D6fDFAfD08c0295C392436245E5bc78B0185"
+      "address": "0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F"
     }
   },
   "GnosisSafeProxyFactory": {
--- a/networks.json
+++ b/networks.json
@@ -1,30 +1,4 @@
 {
-  "BatchExchange": {
-    "1": {
-      "links": {
-        "IdToAddressBiMap": "0xED4d05496C71e71cC2A8726af1242C22108d1761",
-        "IterableAppendOnlySet": "0xCDDB32b6Bb2808D5B5115dAab207479cE98d2636"
-      },
-      "address": "0x6F400810b62df8E13fded51bE75fF5393eaa841F",
-      "transactionHash": "0xe74e8e965afc67f96a38caee794f008be33a7fe7ea96b9baad7b2963250ac36a"
-    },
-    "4": {
-      "links": {
-        "IdToAddressBiMap": "0x5c4C6bf91240A5fdBfB9a1BEd8d43227046e2feA",
-        "IterableAppendOnlySet": "0x0D47D0548FDAD66B06E81a826EED8c687aCddBCB"
-      },
-      "address": "0xC576eA7bd102F7E476368a5E98FA455d1Ea34dE2",
-      "transactionHash": "0xc5bf4f48747093a79b623c2a2392ccbdb73a3d788e7d1f0f53f640ea18496d90"
-    },
-    "100": {
-      "links": {
-        "IdToAddressBiMap": "0x048E53A455a058462eA58442E1D94Fbc955495cB",
-        "IterableAppendOnlySet": "0x57E6B987c2ccd421859A244dd22a0D5a62D88f91"
-      },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -43,11 +43,11 @@
     "truffle-plugin-verify": "^0.4.0"
   },
   "dependencies": {
-    "@gnosis.pm/dex-contracts": "^0.4.2",
+    "@gnosis.pm/dex-contracts": "^0.4.3",
     "@gnosis.pm/owl-token": "^3.1.0",
-    "@gnosis.pm/safe-contracts": "v1.1.1-dev.2",
+    "@gnosis.pm/safe-contracts": "v1.2.0",
```
fedgiac, line 48: I already did the research and I believe that this version switch is safe in this case.
On the other end, I think the change deserves its own PR, and without any check it's very risky because the code in this repo would use the artifacts of a Safe v1.2.0 to manage the currently existing Safe v1.1.1, and there are no unit tests covering this scenario.

scripts/utils/gnosis_safe_server_interactions.js
```diff
@@ -13,6 +13,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
   const linkPrefix = {
     rinkeby: "rinkeby.",
     mainnet: "",
+    xdai: "xdai",
```
fedgiac, line 16:
```
$ curl  https://xdai.gnosis-safe.io/app/#/welcome
curl: (6) Could not resolve host: xdai.gnosis-safe.io
```
There is currently no frontend at this address, is this expected to change?

fedgiac (commented):
Thanks for putting in the work.

I can't see the [contract code](https://blockscout.com/poa/xdai/address/0x473E1D34458F26BFeB40f304a13c80FA291d0Abc/contracts) for the fleet factories on xDAI, can you include the xDAI contract verification procedure?

The other contracts look good (but see comments).

scripts/utils/gnosis_safe_server_interactions.js
fedgiac: It's good however that the [transaction service exists](https://safe-transaction.xdai.gnosis.io/api/v1/about/) .

networks.json
```diff
@@ -1,4 +1,30 @@
...
+    },
+    "100": {
+      "links": {
+        "IdToAddressBiMap": "0x048E53A455a058462eA58442E1D94Fbc955495cB",
+        "IterableAppendOnlySet": "0x57E6B987c2ccd421859A244dd22a0D5a62D88f91"
+      },
+      "address": "0x25B06305CC4ec6AfCF3E7c0b673da1EF8ae26313",
+      "transactionHash": "0xd5e7da686ace1ae7f1917ad54b8610ee936bdc900e56b495739795fba3d5e7c1"
+    }
+  },
```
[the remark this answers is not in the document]
fleupold: They got included when I ran `yarn networks-extract`to get the new xDAI addresses. I will remove them manually but it seems a bit weird that we have to manually adjust the networks.json in so many places.

scripts/utils/gnosis_safe_server_interactions.js
fleupold: Yes, the frontend will be deployed ETA next week

networks.json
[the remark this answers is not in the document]
fedgiac: Yeah, I agree. 
A good solution might be to change `@gnosis.pm/util-contracts/src/inject_network_info.js` so that if you are trying to inject a network into an artifact that already has a (different) address then the execution fails. In this way, we can keep all addresses we use in this file, and if some artifacts get a different address with an update we get a failing CI. (Which is good, since changes to these addresses would most likely command some manual check.)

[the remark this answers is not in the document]
fleupold: Usually, you run `yarn networks-inject` before migrating though, so the address would be set as there is non already. I think we should try to avoid keeping our own copy of the SafeArtefact in the repo and instead of using 

```
const GnosisSafe = artifacts.require("./GnosisSafe.sol")
```

use as Richard suggested

```
const GnosisSafeBuildInfo = require("@gnosis.pm/safe-contracts/build/contracts/GnosisSafe.json")
const GnosisSafe = truffleContract(GnosisSafeBuildInfo)
```

and somehow make sure development deployments are recorded in the dependent npm artefact.

fleupold:
> can you include the xDAI contract verification procedure?

Same as on etherscan. Click verify/publish on the Contract address and add the code (can be taken from Etherscan as well as the compiler version). Verified contracts are here:

https://blockscout.com/poa/xdai/address/0xE91c3d95629465e597CC35ee407d0f11aE19646E/contracts
https://blockscout.com/poa/xdai/address/0x473E1D34458F26BFeB40f304a13c80FA291d0Abc/contracts

scripts/utils/gnosis_safe_server_interactions.js
```diff
@@ -132,7 +133,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
    *
    * @param {Address} multisigAddress Address of the multisig Safe for
    * which to retrieve the nonce.
-   * @param {string} network Either rinkeby or mainnet.
+   * @param {string} network Either rinkeby, dai or mainnet.
```
bh2smith, line 136:
```suggestion
   * @param {string} network Either rinkeby, xdai or mainnet.
```

bh2smith (approved):
This looks fantastic! Nice work and no invasive changes! 🥇

fedgiac (approved):

PR #519, opened 2020-10-08
Allow liquidity provision without Web

fleupold:
This PR makes it so that we can do liquidity provision without requiring the Safe Webinterface (by invoking executeTransaction directly, assuming a single owner safe).

This will be particularly useful on xDAI where the Safe ecosystem won't be ready before earliest the end of next week. The default behaviour didn't change.

### Test Plan

Old way still works:

```
yarn truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=8 --lowestLimit=300 --highestLimit=400 --currentPrice=340 --masterSafe=$MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=0 --numBrackets=10 --network=$NETWORK_NAME
```

But also new way works and leads to executed txs (e.g. [this one](https://blockscout.com/poa/xdai/tx/0x93e5608647859c8f0486f92b9fdee2e52689fc605ff1e4d3e7f5208208b49a17/token-transfers))
```
 yarn truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=8 --lowestLimit=300 --highestLimit=400 --currentPrice=340 --masterSafe=$MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=0 --numBrackets=10 --network=$NETWORK_NAME --executeOnchain --nonce 4
```

### DIFF
--- a/scripts/utils/gnosis_safe_server_interactions.js
+++ b/scripts/utils/gnosis_safe_server_interactions.js
@@ -6,9 +6,8 @@
 
 module.exports = function (web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
-  const { getSafeCompatibleSignature, estimateGas } = require("./internals")(web3, artifacts)
+  const { signTransaction, estimateGas } = require("./internals")(web3, artifacts)
   const { ZERO_ADDRESS } = require("./constants")
-  const CommonBaseGasForGnosisSafeTransaction = 0
 
   const linkPrefix = {
     rinkeby: "rinkeby.",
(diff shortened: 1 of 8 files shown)


fedgiac (approved):
Works nicely with my 1 out of 2 Safe!
Can you add a check to make the script fail if the Safe's `threshold` is larger than one? Otherwise the script makes the user pay to create new Safes and then fails afterwards.

scripts/utils/gnosis_safe_server_interactions.js
```diff
@@ -6,9 +6,8 @@
 
 module.exports = function (web3 = web3, artifacts = artifacts) {
   const axios = require("axios")
-  const { getSafeCompatibleSignature, estimateGas } = require("./internals")(web3, artifacts)
+  const { signTransaction, estimateGas } = require("./internals")(web3, artifacts)
```
bh2smith, line 9: I like this rename and functionality change!

bh2smith (approved):
Very nice looking changes here!

PR #526, opened 2020-10-09
[Easy] Run lint/prettier on pre-commit hook

fleupold:
I have run into a few failing travis builds because I didn't check fast/easy things like lint/prettier locally. This PR makes them a pre-commit hook using husky so that we won't commit code that will for trivial reasons not pass CI.

### Test Plan

See it being run on this PR.


bh2smith (approved):
Yes please!

PR #532, opened 2020-10-13
Allow withdrawl of balances < 1$

fleupold:
We have a feature preventing small amounts (<$1) from being withdrawn, since on mainnet this is likely not economically viable. However, this feature heavily relies on a working 1Inch price estimator. This may not work for some tokens such as wxDAI. Also such a limit does not necessarily make sense on other networks where gas fees are low ($1 seems a bit arbitrary and should probably be configurable to begin with). 

We do have an option for the `request_withdraw` script that will ignore this "protection" and withdraw amounts regardless of value. However, the check for small balance not only happens in the amount estimation function ([here](https://github.com/gnosis/dex-liquidity-provision/blob/f80edbdeb3f11c97f400eb0e974815417c37693e/scripts/wrapper/withdraw.js#L141-L149)) but also in `getWithdrawalsAndTokenInfo` where the `noBalanceCheck` flag is not considered. Thus this check will fail even if we specify `--noBalanceCheck`.

Thus it is currently not possible to force withdrawal of small amounts.

This PR removes the second balance check in `getWithdrawalsAndTokenInfo` as it doesn't seem necessary. If the balance is not sufficient to begin with we will have not requested a small amount to be withdrawn and thus won't even attempt to claim the 0 withdraw. While at it I changed the string comparison to a BigNumber comparison as some `getAmount` functions returned a BN (in `prepareWithdraw` & `prepareWithdrawAndTransferFundsToMaster`). Thus we were comparing Strings with BN via `===` (typescript would have caught this).

### Test Plan

Withdrawing my test liquidity on xDAI works without a 1Inch price estimation failure

```
yarn truffle exec scripts/request_withdraw.js --masterSafe=$MASTER_SAFE --brackets=0x5D6b329918c2BF4d9B722c5c8F8d981E61974d03,0x52033e1F448C69eD6AdDCebF47f9a5836df6395D,0x9A5aB9da6Ee6D9907B19738824C25dfAf0482CD6,0x876EF54A5833DaF55F2c27b4fD098cfdbdb92b70,0x0BdB46E41c80749F5506f51037a981276Dd982E4,0x463Ba08241022F65416a57b428EE32cB25402918,0xf1E115BB04789f85A85f033be932953Cb09DFc72,0xAB795A52D809e4Aa56da23b215Db7663d57CDE3A,0x45d8a3d10cd7Da86Dac21a51B305312Ebf80072d,0x2952bC69739c9534d09525ccCa33950F311CbE88 --tokenIds=1,4 --network=$NETWORK_NAME --executeOnchain --noBalanceCheck
```

Also unit tests still pass.

### DIFF
--- a/scripts/wrapper/withdraw.js
+++ b/scripts/wrapper/withdraw.js
@@ -14,7 +14,7 @@ module.exports = function (web3, artifacts) {
   } = require("../utils/trading_strategy_helpers")(web3, artifacts)
   const { default_yargs, checkBracketsForDuplicate } = require("../utils/default_yargs")
   const { fromErc20Units, shortenedAddress } = require("../utils/printing_tools")
-  const { MAXUINT256, ONE } = require("../utils/constants")
+  const { MAXUINT256, ONE, ZERO } = require("../utils/constants")
   const { uniqueItems } = require("../utils/js_helpers")
 
   const assertGoodArguments = function (argv) {
@@ -43,8 +43,7 @@ module.exports = function (web3, artifacts) {
     brackets,
     tokens,
     tokenIds,
-    printOutput = false,
-    globalPriceStorage = {}
+    printOutput = false
   ) {
     const log = printOutput ? (...a) => console.log(...a) : () => {}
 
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
I am very much in favour of these changes! Especially to remove that `globalPriceStorage` memo (although it seems like a somewhat unrelated change).

fedgiac (approved):
Looks good!

fleupold:
> Especially to remove that globalPriceStorage memo (although it seems like a somewhat unrelated change).

It's not unrelated as the linter complains if the variable is passed but not read (which it no longer is).

PR #534, opened 2020-10-13
Inverted price log

### LINKED ISSUE #515: 1Inch PriceEstimate and amount is inverse.
The console log says:

```
1inch price: 353.5065894727491 USDC bought for 1 WETH
```

But what we estimate is https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=USDC&toTokenSymbol=WETH&amount=1000000

which gives us the amount WETH bought for 1 USDC.

At the time of writing this price ($353) is significantly different than the inverse ($335)

bh2smith:
Inverted meaning in log statements lead to confusion of interpretation. 

See [here](https://www.axiory.com/trading-resources/trading-terms/currency-quotes) for definitions of base and quote aligned with the correspondence between 1Inch terminology that `from = quote` and `to = base`

Closes #515

### DIFF
--- a/scripts/utils/price_utils.js
+++ b/scripts/utils/price_utils.js
@@ -181,9 +181,9 @@ const isPriceReasonable = async (
     console.log(
       `Warning: the chosen price differs by more than ${acceptedPriceDeviationInPercentage} percent from the price found on ${onlinePriceSlice.source}.`
     )
-    console.log(`    chosen price: ${price} ${quoteTokenData.symbol} bought for 1 ${baseTokenData.symbol}`)
+    console.log(`    chosen price: ${price} ${quoteTokenData.symbol} needed to buy 1 ${baseTokenData.symbol}`)
     console.log(
-      `    ${onlinePriceSlice.source} price: ${onlinePrice} ${quoteTokenData.symbol} bought for 1 ${baseTokenData.symbol}`
+      `    ${onlinePriceSlice.source} price: ${onlinePrice} ${quoteTokenData.symbol} needed to buy 1 ${baseTokenData.symbol}`
     )
     return false
   }


fedgiac (approved):

nlordell:
> See here for definitions of base and quote aligned with the correspondence between 1Inch terminology that `from = quote` and `to = base`

I don't think this is true. I think if you have `from = quote` and `to = base` then you getting the bid price for a given amount. If you were to invert it and say `from = base` and `to = quote` then you would be getting an ask price. Both are correct, they just price different things.

What is import is that the price is always computed as `quoteAmount / baseAmount`.

nlordell (approved):
Log statements are changed to be more accurate given how price estimates are currently computed with 1Inch.

bh2smith:
@nlordell 

> I don't think this is true. 

While this may not be correct, my statement is merely referring to the fact that this is the mapping we are currently using in our 1Inch price query URL here:

https://github.com/gnosis/dex-liquidity-provision/blob/aab35854c1c9e6814c6d81e6bd7063ac5570545c/scripts/utils/price_utils.js#L134-L140

nlordell:
> my statement is merely referring to the fact that this is the mapping we are currently using in our 1Inch price query URL here

Ah, sorry I misunderstood! I thought you meant that as a general statement about what `to` and `from` correspond to in 1Inch.

> We could also change this around to be more accurate if necessary, but I would think this should be a separate PR.

I don't think that is necessary. Its just important to keep in mind that we are estimating a bid order, i.e. finding the best asks. In the same market (base-quote) there are always two directions bids and asks.

One cleanup thing that we should do is to maybe rename the function to be more explicit about *what* 1Inch price is being estimated.

PR #536, opened 2020-10-13
use packaged asset for GnosisSafeProxyFactory

fleupold:
As #528 but this time for GnosisSafeProxyFactory

### Test Plan

CI

### DIFF
--- a/migrations/2_dev_dependencies.js
+++ b/migrations/2_dev_dependencies.js
@@ -1,7 +1,7 @@
 const Artifactor = require("@truffle/artifactor")
 const migrateBatchExchange = require("@gnosis.pm/dex-contracts/src/migration/PoC_dfusion")
 const { GnosisSafe } = require("../scripts/utils/dependencies")(web3, artifacts)
-const ProxyFactory = artifacts.require("./GnosisSafeProxyFactory.sol")
+const { GnosisSafeProxyFactory } = require("../scripts/utils/dependencies")(web3, artifacts)
 const MultiSend = artifacts.require("./MultiSend.sol")
 
 module.exports = async function (deployer, network, accounts) {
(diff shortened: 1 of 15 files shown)


bh2smith (approved):
Sweet sauce! Are there any others like this coming soon?

fleupold:
> Are there any others like this coming soon?

Yes, MultiSend and BatchExchange dependency are still missing.

bh2smith:
> Yes, MultiSend and BatchExchange dependency are still missing.

I would be happy to take on one of these today if you don't mind.

migrations/2_dev_dependencies.js
```diff
@@ -1,7 +1,7 @@
 const Artifactor = require("@truffle/artifactor")
 const migrateBatchExchange = require("@gnosis.pm/dex-contracts/src/migration/PoC_dfusion")
 const { GnosisSafe } = require("../scripts/utils/dependencies")(web3, artifacts)
-const ProxyFactory = artifacts.require("./GnosisSafeProxyFactory.sol")
+const { GnosisSafeProxyFactory } = require("../scripts/utils/dependencies")(web3, artifacts)
```
bh2smith, line 4: Can these both be imported in the same line?

fleupold: Yes they can...

migrations/2_dev_dependencies.js
```diff
@@ -28,7 +27,9 @@ module.exports = async function (deployer, network, accounts) {
 }
 
 async function deploySafe(deployer) {
```
bh2smith, line 29: If we plan to move other items into here like Multisend, perhaps this function could use a rename `deploySafeItems`. Or is Multisend something that would continue to live outside?

fleupold:
> I would be happy to take on one of these today if you don't mind.

Sorry I just saw your comment after already having converted the MultiSend one. 
But please go ahead with the BatchExchange one. It should be very similar (just need to use another Artefactor with a path to the dex-contracts repo as the build info is not in the safe package).

migrations/2_dev_dependencies.js
fleupold: Done in #537

fedgiac (approved):
Looks good!

PR #537, opened 2020-10-14
[Easy] Use packaged MultiSend contract instead of local deployment

fleupold:
Like #536 but smaller

### Test Plan

CI


bh2smith (approved):
Looks good although I was expecting to that see some import statements would need changing in the test suite. If this is not the case then all good.

fedgiac (approved):

PR #543, opened 2020-10-19
Packaged BatchExchange

### LINKED ISSUE #517: Write Safe Artefacts with deployed addresses on Migration

bh2smith:
This is the last PR in a series of updated migrations (#528, #536, #537) that use packaged versions of contract artifacts.

Will require updated npm package for dex-contracts after https://github.com/gnosis/dex-contracts/pull/1022 has been merged.

Closes #517

### DIFF
--- a/contracts/Dependencies.sol
+++ b/contracts/Dependencies.sol
@@ -8,8 +8,8 @@ pragma solidity ^0.5.0;
 // Token Dependencies
 import "@openzeppelin/contracts/token/ERC20/ERC20Detailed.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20Mintable.sol";
-import "@gnosis.pm/owl-token/contracts/TokenOWLProxy.sol";
-import "@gnosis.pm/owl-token/contracts/TokenOWL.sol";
+import "@gnosis.pm/owl-token/contracts/5/TokenOWLProxy.sol";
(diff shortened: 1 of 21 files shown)


fleupold (approved):
Looks great! Can you remove the addresses from the networks.json please?

bh2smith:
> Can you remove the addresses from the networks.json please?

I already removed the `BatchExchange` address... is there something else, like OWL that I missed?

PR #547, opened 2020-10-21
[Easy] update owl-token dependency

bh2smith:
This should be the only project depending on the one specific version of owl token that we are hoping to unpublish!


nlordell (approved):

bh2smith:
Closed in favour of #548

PR #549, opened 2020-10-23
Add withdrawal verification

fedgiac:
Add the option to verify withdraw transactions.

This required making the order of the computed withdrawals deterministic by sorting them.

### Test plan

Full cycle of deploying liquidity -> withdrawing request -> withdraw.

You can also check that verification works by running the following line:
```
npx --no-install truffle exec scripts/request_withdraw.js --masterSafe 0x04a82eC4857694a28bdDb9cE8E13DCbc090928e7 --brackets 0x5c4c56061d6f60a7bef0670cb1381bf4a5689c93,0x84f697cfea104668f708724087f1927b48a7a163,0xfc4cf9b9d87c2c89781f68ea2890faed17cad433,0x6275cbc5b1a3982a61b26434d162d03c24469124,0x3edce898e7c721c809fd3b697406d04ee44f34ff,0x61d65f773bfe13d99c5585c2ba7f29b53e1ba8c3,0x46f0ce88f38cad5dc612ac1e7f7b79853861402b,0x9ef6e107492bff82f1a826ab56663b9049692708,0x91c96989dcda2549e8f915fb936c9117a2671f82,0x318e9d4818d3b44cac28f8dc5ae384f4357e983d --network=rinkeby --nonce=46 --verify
```
https://rinkeby.gnosis-safe.io/app/#/safes/0x04a82eC4857694a28bdDb9cE8E13DCbc090928e7/transactions


scripts/utils/gnosis_safe_server_interactions.js
```diff
@@ -142,9 +143,52 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
...
+   * @param {boolean} verifyOnly Whether to verify if the transaction exists
+   * or creating a new one.
+   * @param {SmartContract} masterSafe The Safe used to execute the
+   * transaction.
+   * @param {number} nonce The nonce used by the transaction.
+   * @param {Transaction} transaction The transaction to be sent.
+   * @param {string} network Either rinkeby, xdai or mainnet.
+   * @param {boolean} executeOnchain Whether the transaction should be
+   * executed immediately onchain or just created in the web interface.
+   * @param {boolean} mustPromptUser Whether to prompt the user with a
```
bh2smith, line 159: I would be a bit happier if this were simply called `promptUser`, but I guess this would conflict with the function name.

bh2smith (changes requested):
This PR includes a very nice change, but it looks like there are two independent things happening here. That is

1. Cleanup the scripts by introducing a nice new utility method `processTransaction`
2. Sorting withdraws

Sorry to say, I would think that these should really be split into two PRs.

scripts/utils/gnosis_safe_server_interactions.js
fedgiac: I agree with you, so much so that this was the original name I gave it. Then I realized that this was the name of the function prompting the user...

bh2smith (approved):
Looks good to me! Nice work.

PR #555, opened 2020-10-26
Add single function to manage transaction submission

fedgiac:
Prune repeating logic when sending transactions. See #549 for an example use case.

As a proof of concept, I used this function to implement `--executeOnchain` for `single_transaction_liquidity_provision`.

### Test plan

Try that everything works as expected, for example:
```
npx truffle exec scripts/single_transaction_liquidity_provision.js --baseTokenId=0 --quoteTokenId=3 --lowestLimit=1 --highestLimit=100 --currentPrice=1.1 --masterSafe $MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=10000 --numBrackets=10 --network=rinkeby
```

Then follow the instructions on screen to test that verification is working and execute the transaction.

Finally test `--executeOnchain`:
```

npx truffle exec scripts/single_transaction_liquidity_provision.js --baseTokenId=0 --quoteTokenId=3 --lowestLimit=1 --highestLimit=100 --currentPrice=1.1 --masterSafe $MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=10000 --numBrackets=10 --network=rinkeby --executeOnchain
```

### DIFF
--- a/scripts/utils/trading_strategy_helpers.js
+++ b/scripts/utils/trading_strategy_helpers.js
@@ -1002,6 +1002,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     const tokenInfoPromises = fetchTokenInfoFromExchange(await exchangePromise, [baseTokenId, quoteTokenId])
     const createFleetTransaction = buildDeterministicFleetOfSafes(masterAddress, fleetSize, await noncePromise)
     const bracketAddresses = await calcSafeAddresses(fleetSize, await noncePromise)
+    console.log(bracketAddresses)
 
     const createOrderTransactions = transactionsForOrders(
       masterAddress,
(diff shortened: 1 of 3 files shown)


scripts/utils/trading_strategy_helpers.js
```diff
@@ -1002,6 +1002,7 @@ module.exports = function (web3 = web3, artifacts = artifacts) {
     const tokenInfoPromises = fetchTokenInfoFromExchange(await exchangePromise, [baseTokenId, quoteTokenId])
     const createFleetTransaction = buildDeterministicFleetOfSafes(masterAddress, fleetSize, await noncePromise)
     const bracketAddresses = await calcSafeAddresses(fleetSize, await noncePromise)
+    console.log(bracketAddresses)
```
bh2smith, line 1005: Was this intended? Do we not have a dedicated logger in this method?

bh2smith (approved):
Really nice improvement to the repo. Just one console log that might be unintentionally included!

bh2smith (commented):
Also, I thought there were a lot more places where this `processTransaction` functionality could be adopted. For example `requestWithdraw`, `claimWithdraw`, etc or anywhere we are using the `promptUser` function.

scripts/utils/trading_strategy_helpers.js
fedgiac: Whoops!

fedgiac:
Using the new function on the withdraw scripts would lead to buggy behavior, since the transactions are not sorted, so I'll introduce it there in #549. Since there could be the same problem for the other scripts and testing them takes time, I'd first merge this, then the withdraw one, and eventually the remaining functions if everything looks good.

PR #556, opened 2020-10-27
Fix incorrect use of Yarn conflicts

fedgiac:
This bug fixes a recent issue in the script because of the newly introduced flags `--nonce` and `--executeOnchain`. Currently, the complete_liquidity_provision script is not working on master because of a yargs mismatch.

Currently, we use Yargs's `conflicts` to enforce that the two flags are not used together. However, since we give a default value to both (different from `undefined`) this counts as if both flags were set.

This PR gets rid of the `conflicts` flag and replaces it with a custom function.

### Test plan.

Run the following command both in master and on the curent PR to see the improvement. (No need to execute or send transactions, it's enough to see that the options are properly parsed.)
```
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=3 --lowestLimit=200 --highestLimit=250 --currentPrice=245 --masterSafe $MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=0 --numBrackets=5 --network=rinkeby
```

Test for (expected) failure with both flags:
```
npx truffle exec scripts/complete_liquidity_provision.js --baseTokenId=1 --quoteTokenId=3 --lowestLimit=200 --highestLimit=250 --currentPrice=245 --masterSafe $MASTER_SAFE --depositBaseToken=0 --depositQuoteToken=0 --numBrackets=5 --network=rinkeby --nonce=10 --executeOnchain
```

### DIFF
--- a/scripts/complete_liquidity_provision.js
+++ b/scripts/complete_liquidity_provision.js
@@ -15,7 +15,7 @@ const { signAndExecute } = require("./utils/internals")(web3, artifacts)
 const { proceedAnyways } = require("./utils/user_interface_helpers")
 const { sleep } = require("./utils/js_helpers")
 const { DEFAULT_NUM_SAFES } = require("./utils/constants")
-const { default_yargs, checkBracketsForDuplicate } = require("./utils/default_yargs")
+const { default_yargs, checkBracketsForDuplicate, noNonceWithExecuteOnchain } = require("./utils/default_yargs")
 
 const argv = default_yargs
   .option("masterSafe", {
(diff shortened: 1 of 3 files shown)


bh2smith (approved):
Seems legit. Nice find, although I am generally not satisfied with having conflicting arguments, I suppose it is inevitable with such an elaborate script.

fedgiac:
Yeah, it gets hard to keep track of everything. Here the conflict comes from the fact that if you execute onchain then you must use the first nonce available onchain, otherwise the transaction fails.

PR #557, opened 2020-10-27
Patch changes to 1inch API WETH handling

fedgiac:
It seems like 1inch changed their API in a way that makes WETH an invalid token (price returned is zero). Compare a [WETH](https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=USDC&toTokenSymbol=WETH&amount=1000000) trade with an [ETH](https://api.1inch.exchange/v1.1/quote?fromTokenSymbol=USDC&toTokenSymbol=ETH&amount=1000000) trade. (Zero `toTokenAmount` in one case, reasonable value in the other case.)

This PR patches this change by replacing `WETH` with `ETH` in the function that retrieves 1inch prices.

### Test plan

Try to request withdraw WETH without ignoring the price check. Unit tests.


bh2smith (approved):
Sweet deal. Nice fix.

PR #561, opened 2020-10-30
Reorganize balance checks on request withdraw

fedgiac:
This PR introduces a new flag `--onlySkipNonzero` for `request_withdraw.js`. If a withdraw is requested with this flag, all nonzero user balances are requested to be withdrawn from the exchange. This is helpful to more efficiently withdraw tokens that aren't priced by our price oracle (e.g. `xDAI`), which currently require skipping the balance check altogether, and was explicitly requested for the development of the CMM app (please confirm this works for you @andre-meyer).

The default behavior of the scripts is still to check the withdrawn balance, since trading on GP always leaves some dust amounts unmatched.

Two more minor changes are part of this reorganization:
1. the flag `--noBalanceCheck` is replaced by `--stopTrading`, to reflect the main purpose of this flag after introducing the new flag.
2. `amountUSDValue` accepts the promise of an amount, in this way the price request to 1inch can be executed in parallel to the retrieving of the balance onchain.

### Test plan

Deploy a strategy with some funds, e.g.:
```
npx truffle exec scripts/single_transaction_liquidity_provision.js --baseTokenId=4 --quoteTokenId=3 --lowestLimit=0.9 --highestLimit=1 --currentPrice=0.92 --masterSafe $MASTER_SAFE --depositBaseToken=1 --depositQuoteToken=10000 --numBrackets=10 --network=rinkeby --executeOnchain
```

Try all new combinations of withdraw flags and see in the resulting console output that everything behaves as expected (no need to send the transaction):
1. skips value <1$
```
npx truffle exec scripts/request_withdraw.js --masterSafe $MASTER_SAFE --brackets $BRACKETS --network=rinkeby
```
2. skips only empty balances
```
npx truffle exec scripts/request_withdraw.js --masterSafe $MASTER_SAFE --brackets $BRACKETS --network=rinkeby --onlySkipNonzero
```
3. withdraws everything
```
npx truffle exec scripts/request_withdraw.js --masterSafe $MASTER_SAFE --brackets $BRACKETS --network=rinkeby --stopTrading
```
4. failure, invalid arguments
```
npx truffle exec scripts/request_withdraw.js --masterSafe $MASTER_SAFE --brackets $BRACKETS --network=rinkeby --stopTrading --onlySkipNonzero
```

### DIFF
--- a/scripts/request_withdraw.js
+++ b/scripts/request_withdraw.js
@@ -14,7 +14,7 @@ const argv = defaultWithdrawYargs
   .option("onlySkipNonzero", {
     type: "boolean",
     default: false,
-    describe: "Withdraw balance of all nonzero balances, even if the amounts have very small USD value",
+    describe: "Withdraw balance of all non-zero balances, even if the amounts have very small USD value",
   }).argv
 
 module.exports = async (callback) => {
--- a/scripts/request_withdraw.js
+++ b/scripts/request_withdraw.js
@@ -11,7 +11,7 @@ const argv = defaultWithdrawYargs
     describe:
       "Request withdraw of all funds on each bracket for each traded tokens, independently of their actual balance. Selected brackets will stop any trading from the following batch",
   })
-  .option("onlySkipNonzero", {
+  .option("skipZero", {
     type: "boolean",
     default: false,
     describe: "Withdraw balance of all non-zero balances, even if the amounts have very small USD value",
--- a/scripts/request_withdraw.js
+++ b/scripts/request_withdraw.js
@@ -4,12 +4,18 @@ const { defaultWithdrawYargs, prepareWithdrawRequest } = require("./wrapper/with
 const { signAndExecute } = require("./utils/internals")(web3, artifacts)
 const { promptUser } = require("./utils/user_interface_helpers")
 
-const argv = defaultWithdrawYargs.option("noBalanceCheck", {
-  type: "boolean",
-  default: false,
-  describe:
-    "Request withdraw for tokens without checking if the token has no balance. Useful to account for trading that might be happening during the withdraw request",
-}).argv
+const argv = defaultWithdrawYargs
+  .option("stopTrading", {
+    type: "boolean",
+    default: false,
+    describe:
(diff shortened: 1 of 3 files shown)


scripts/request_withdraw.js
```diff
@@ -4,12 +4,18 @@ const { defaultWithdrawYargs, prepareWithdrawRequest } = require("./wrapper/with
...
-    "Request withdraw for tokens without checking if the token has no balance. Useful to account for trading that might be happening during the withdraw request",
-}).argv
+const argv = defaultWithdrawYargs
+  .option("stopTrading", {
+    type: "boolean",
+    default: false,
+    describe:
+      "Request withdraw of all funds on each bracket for each traded tokens, independently of their actual balance. Selected brackets will stop any trading from the following batch",
+  })
+  .option("onlySkipNonzero", {
```
bh2smith, line 14: Are we sure this name shouldn't be `onlySkipZero`. Why would I want to skip non-zero balances? Either way, if we keep the name here I think the Z should be capitalized. Furthermore, in either case, `onlySkipZero`, I think this name could be reduced to `skipZero` (or `skipNonZero`) - i.e. without the only.

scripts/request_withdraw.js
```diff
@@ -4,12 +4,18 @@ const { defaultWithdrawYargs, prepareWithdrawRequest } = require("./wrapper/with
...
+    type: "boolean",
+    default: false,
+    describe: "Withdraw balance of all nonzero balances, even if the amounts have very small USD value",
```
bh2smith, line 17:
```suggestion
    describe: "Withdraw balance of all non-zero balances, even if the amounts have very small USD value",
```

bh2smith (approved):
This looks fine. Although, as I have said a few times before, I am still not terribly keen on having all these clashing script argument constraints. I think we should consider renaming the `onlySkipNonzero` for a few different reasons explained in the inline comments.

scripts/request_withdraw.js
fedgiac: I agree this is confusing, I changed it to `--skipZero`.
The "only" was there to point out that if you withdraw without `--skipZero` you are going to skip zero balances anyway, along with all balances of value <1$.

bh2smith:
This was never merged and is essentially stale now. Closing because we are archiving this repo.

